PR #3, opened 2025-03-21
Initial MCP client/server implementation

jakemac53:
This implements just enough to call tools registered from a server from a client.

This is about the minimal thing I can send out and actually test, I will try and keep future PRs smaller.

We also will likely want to tweak the APIs here in the future.


pkgs/dart_mcp/lib/src/api.dart
```diff
@@ -0,0 +1,1300 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
```
jakemac53, line 1: Note that there are many commented out typescript definitions here - these are the ones I haven't yet ported.

We likely want a code generator, I ported just enough to get the steel thread sort of through here, but its easiest to understand the current state to just leave in these un-ported types as comments.

jwren:
LGTM great progress. I assume that all the `print` statements will be replaced by logs with time?

jwren (approved):

devoncarew (commented):
As a drive-by-comment, I'd also contribute a workflow for this package; perhaps something simple, like the one for https://github.com/dart-lang/tools/blob/main/.github/workflows/graphs.yaml?

jakemac53:
> As a drive-by-comment, I'd also contribute a workflow for this package; perhaps something simple, like the one for https://github.com/dart-lang/tools/blob/main/.github/workflows/graphs.yaml?

Yeah I was planning to do that next just didn't get to it quite yet

jakemac53:
> LGTM great progress. I assume that all the `print` statements will be replaced by logs with time?

We could certainly integrate some sort of logging framework here yeah, for now these prints are only in the example client itself to help you follow along.

But, I think it would be useful for debugging to be able to output transcripts of all the messages sent to/from each server to files as well so we probably do want to add something like that.

PR #9, opened 2025-03-24
Add the beginnings of a Dart tooling MCP server

jakemac53:
This service demonstrates our ability to connect to the dart tooling daemon to perform tasks, as well as the vm service connections we can access through that.

Currently screenshots work for flutter apps, but we can easily plug in more.

In the future we likely are going to want to make the DTD connection more lazy - so that we don't require it on startup, but instead ask for it as a part of the tool methods that require it. This should result in a more friendly user experience for setting up the tool (assuming the clients prompt them for it).

TODO: tests, which will likely require some significant setup (flutter SDK + active DTD connection), hopefully we can leverage some existing devtools testing functionality for this.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/README.md
+++ b/pkgs/dart_tooling_mcp_server/README.md
@@ -36,3 +36,13 @@ If you are directly editing your mcp.json file, it should look like this:
   }
 }
 ```
+
+## Debugging MCP Servers
+
+See https://modelcontextprotocol.io/docs/tools/inspector.
+
+The setup here is basically the same as the cursor config, launch the inspector
+with no options, and then plug in the executable path and DTD Uri into the
+connection window.
+
+You should then see all available functionality from the server.
--- a/pkgs/dart_tooling_mcp_server/README.md
+++ b/pkgs/dart_tooling_mcp_server/README.md
@@ -1,0 +1,5 @@
+Experimental MCP server which exposes Dart development tool actions to clients.
+
+## Status
+
+WIP
--- a/pkgs/dart_tooling_mcp_server/README.md
+++ b/pkgs/dart_tooling_mcp_server/README.md
@@ -3,3 +3,36 @@ Experimental MCP server which exposes Dart development tool actions to clients.
 ## Status
 
 WIP
+
+## Using this package
+
+To use this package, will need to compile the `bin/main.dart` script to exe
+(`dart compile bin/main.dart`) and use the compiled path as the command in your
+MCP server config.
+
+The command also requires a DTD Uri to connect to your current debug session,
+so the first argument will have to be that URI. In VsCode, you can get this
+by using the "Copy DTD URI to Clipboard" command from the Dart extension.
+
+### With Cursor
+
+Go to Cursor -> Settings -> Cursor Settings and select "MCP".
+
+Then, click "Add new global MCP server". Put in the full path to the executable
+you created in the first step, and paste in the DTD uri into the arguments
+section.
+
+If you are directly editing your mcp.json file, it should look like this:
+
+```yaml
+{
+  "mcpServers": {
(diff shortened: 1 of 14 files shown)


pkgs/dart_tooling_mcp_server/README.md
```diff
@@ -0,0 +1,48 @@
+Experimental MCP server which exposes Dart development tool actions to clients.
+
+## Status
+
+WIP
+
+## Using this package
+
+To use this package, will need to compile the `bin/main.dart` script to exe
```
kenzieschmoll, line 9: "...package, <you> will need..."

pkgs/dart_tooling_mcp_server/lib/src/dtd.dart
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:dtd/dtd.dart';
+import 'package:json_rpc_2/json_rpc_2.dart';
+
```
kenzieschmoll, line 9: maybe file an issue and link a TODO: it might be work seing how we can share some of this code in a Dart package. DevTools implements some of the same logic to support the Flutter sidebar in VS code, which calls all these methods over DTD. Maybe at some point it would be worth creating a package dtd_editor_service or similar to share the logic. Out of scope for this PR though and something we'd only want to consider once we have more than one stable use case suggesting a reason to try to de-duplicate.

pkgs/dart_tooling_mcp_server/lib/src/server.dart
```diff
@@ -0,0 +1,128 @@
...
+  @override
+  final implementation = ServerImplementation(
+    name: 'dart tooling daemon',
+    version: '0.1.0-wip',
+  );
+
+  @override
+  final instructions =
+      'This server helps to connect Dart and Flutter developers to their '
+      'running apps.';
```
kenzieschmoll, line 26: not just running apps, but really to dart & flutter developer tools, including access to their running app.

kenzieschmoll (approved):

pkgs/dart_tooling_mcp_server/lib/src/dtd.dart
jakemac53: We could put this stuff directly in the dart tooling daemon package - maybe in a different import (`vscode.dart` or similar?). I am not sure if the Intellij extension registers the exact same extensions (hopefully?).

pkgs/dart_tooling_mcp_server/lib/src/server.dart
```diff
@@ -0,0 +1,128 @@
...
+                    'current state',
+                inputSchema: InputSchema(),
+              ),
+              takeScreenshot,
+            );
+          }
+        case 'ServiceUnregistered':
+          if (e.data['service'] == 'Editor' &&
+              e.data['method'] == 'getDebugSessions') {
+            unregisterTool('take_screenshot');
```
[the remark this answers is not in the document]
jakemac53: I just extracted the whole Tool definition, it can't be const but it is final, and in a top level variable.

pkgs/dart_tooling_mcp_server/lib/src/dtd.dart
kenzieschmoll: Both editors should implement the same extensions (although right now IntelliJ may register a subset of the ones VS Code registers, only including the ones for the property editor). Would be great to have some code generation eventually and some testing that ensures Flutter supported IDEs register the same set of functionality, especially the services we rely on for other features.

PR #11, opened 2025-03-25
Extract the DTD support out to its own mixin, add connect method

jakemac53:
Removes the up front requirement for a DTD uri, instead moving it to a tool method so the user is prompted for a uri.


pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -0,0 +1,292 @@
...
+    final debugSessions = response.debugSessions;
+    if (debugSessions.isEmpty) return _noActiveDebugSession;
+
+    // TODO: Consider holding on to this connection.
+    final vmService = await vmServiceConnectUri(
+      debugSessions.first.vmServiceUri,
+    );
+
+    try {
+      final vm = await vmService.getVM();
```
kenzieschmoll, line 118: out of scope for this PR, but we may want to break up which tools are available for a Flutter app vs a pure Dart app.

jakemac53: ack

PR #12, opened 2025-03-25
Add basic Resources support, refactor some things

jakemac53:
- Adds support for [resources](https://modelcontextprotocol.io/docs/concepts/resources).
- Refactors some things to separate the mixin implementations out into separate files.
- Introduces a TestEnvironment to encapsulate the basic test setup.

PR #13, opened 2025-03-26
Add client support and tests for Resource and Tool streams

jakemac53:
Building on https://github.com/dart-lang/ai/pull/12 (actual new commits are 3c6dfdf4a393c36884e51c53edbda8c2b788fa54 and onwards)

Also makes the base classes actually be marked `base` - these should only be extended as we rely heavily on `initialize` being called all the way up the super chain.


pkgs/dart_mcp/test/resources_support_test.dart
```diff
@@ -45,9 +46,99 @@ void main() {
...
+      ),
+    );
+
+    serverConnection.unsubscribeResource(
+      UnsubscribeRequest(uri: fooResource.uri),
+    );
+    // Let the server process the request
+    await pumpEventQueue();
+
+    fooContents = 'baz';
```
kenzieschmoll, line 120: I assume we are setting to 'baz' again to show that the resource is not updated if nothing changes? If so, can you add a comment?

kenzieschmoll (approved):

pkgs/dart_mcp/test/resources_support_test.dart
jakemac53: Oh, this was actually a mistake, although it doesn't actually need to change at all.

There is no change notification due to the unsubscribe a few lines above.

jakemac53, line 120:
```suggestion
    fooContents = 'zap';
```

PR #17, opened 2025-03-28
Add Logging support

jakemac53:
See https://spec.modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging/

I also fixed a bug in the resource support - I realized that subscribe/unsubscribe are supposed to return an empty response instead of no response.

Also updated a few of the extension types to take a `MetaWithProgressToken`.

### DIFF
--- a/pkgs/dart_mcp/lib/src/api.dart
+++ b/pkgs/dart_mcp/lib/src/api.dart
@@ -224,7 +224,7 @@ extension type RootsCapabilities.fromMap(Map<String, Object?> _value) {
 extension type ServerCapabilities.fromMap(Map<String, Object?> _value) {
   factory ServerCapabilities({
     Map<String, Object?>? experimental,
-    Map<String, Object?>? logging,
+    Logging? logging,
     Prompts? prompts,
     Resources? resources,
     Tools? tools,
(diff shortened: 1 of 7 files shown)


pkgs/dart_mcp/lib/src/api.dart
```diff
@@ -1121,59 +1133,83 @@ extension type InputSchema.fromMap(Map<String, Object?> _value) {
...
-//     /**
-//      * The level of logging that the client wants to receive from the server.
-//      * The server should send all logs at this level and higher (i.e., more
-//      * severe) to the client as notifications/message.
-//      */
-//     level: LoggingLevel;
-//   };
-// }
+/// Extension type for the `logging` capability.
+extension type Logging.fromMap(Map<String, Object?> _value) {
```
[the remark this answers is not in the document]
jakemac53: Sure yeah that makes sense - ok if I do that in a follow-up?

pkgs/dart_mcp/lib/src/server/logging_support.dart
```diff
@@ -0,0 +1,71 @@
...
+  @override
+  FutureOr<InitializeResult> initialize(InitializeRequest request) async {
+    _peer.registerMethod(
+      SetLevelRequest.methodName,
+      convertParameters(handleSetLevel),
+    );
+
+    final result = await super.initialize(request);
+    result.capabilities.logging ??= Logging();
+    return result;
```
kenzieschmoll, line 23: nit: this can be simplified, although maybe less readable, so I will leave up to you.
```dart
return (await super.initialize(request))
  ..capabilities.logging ?? = Logging();
```

pkgs/dart_mcp/test/logging_test.dart
```diff
@@ -0,0 +1,125 @@
...
+      data: 'message',
+    );
+
+    expect(serverConnection.onLog, emits(lazyNotification));
+
+    server.log(lazyNotification.level, () => lazyNotification.data);
+
+    // Below logging level, never gets evaluated.
+    server.log(LoggingLevel.info, () => throw StateError('Unreachable'));
+  });
```
kenzieschmoll, line 120: can we add a test case for throwing an argument error when the function is the wrong type?

pkgs/dart_mcp/lib/src/api.dart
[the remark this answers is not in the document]
kenzieschmoll: Yep follow up refactor sgtm

kenzieschmoll (approved):

pkgs/dart_mcp/test/logging_test.dart
jakemac53: added tests for a few different function types

PR #25, opened 2025-04-01
Port Annotation apis and add test exercising them.

jakemac53:
Not sure why I didn't just do this earlier, but here it is :)

### DIFF
--- a/pkgs/dart_mcp/lib/src/api/api.dart
+++ b/pkgs/dart_mcp/lib/src/api/api.dart
@@ -311,12 +311,13 @@ extension type Annotated._fromMap(Map<String, Object?> _value) {
 
 /// The annotations for an [Annotated] object.
 extension type Annotations.fromMap(Map<String, Object?> _value) {
-  factory Annotations({List<Role>? audience, double? priority}) =>
-      Annotations.fromMap({
-        if (audience != null)
-          'audience': [for (var role in audience) role.name],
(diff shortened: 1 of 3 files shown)


pkgs/dart_mcp/lib/src/api/api.dart
```diff
@@ -293,31 +301,41 @@ extension type EmbeddedResource.fromMap(Map<String, Object?> _value)
...
+/// The client can use annotations to inform how objects are used or displayed.
+extension type Annotated._fromMap(Map<String, Object?> _value) {
+  /// Annotations for this object.
+  Annotations? get annotations => _value['annotations'] as Annotations?;
+}
+
+/// The annotations for an [Annotated] object.
+extension type Annotations.fromMap(Map<String, Object?> _value) {
+  factory Annotations({List<Role>? audience, double? priority}) =>
+      Annotations.fromMap({
```
kenzieschmoll, line 315: should we assert that priority is between 0 and 1?

kenzieschmoll (approved):
one comment, then lgtm

pkgs/dart_mcp/lib/src/api/api.dart
jakemac53: done, this won't happen when casting from a Map, but the main use case is when going through this "constructor" anyways.

PR #38, opened 2025-04-07
add health and publish workflows

jakemac53:
Uses firehose https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose

Also adds a changelog and removes `publish_to: none` from dart_mcp.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -7,7 +7,7 @@ on:
     branches: [ main ]
     types: [opened, synchronize, reopened, labeled, unlabeled]
   push:
-    tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+*' ]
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,22 @@
...
+    tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+*' ]
+
+jobs:
+  publish:
+    if: ${{ github.repository_owner == 'dart-lang' }}
+    uses: dart-lang/ecosystem/.github/workflows/publish.yaml@main
+    with:
+      write-comments: false
+      sdk: dev
+      environment: pub.dev # Can be any name, this is the convention though.
```
devoncarew, line 19: I don't think we want to use an environment when publishing (another button to push w/o a clear benefit).

jakemac53: Oh is this the thing that requires the extra approval step?

PR #39, opened 2025-04-07
move ServerImplementation and instructions to a final field

jakemac53:
There was a bit of a discrepancy between the MCPClient and MCPServer interfaces here, the client had a final field you had to initialize in the constructor, and the server made you override abstract getters.

I opted for the MCPClient pattern here, but I don't think there is any strong reason to choose one over the other, it just needs to be consistent.


kenzieschmoll (approved):

natebosch (approved):
I think this is a good choice - I almost always prefer passing information explicitly through arguments over implicitly through overrides.

Issue #29, opened 2025-04-02
Add extension types for input schemas


jakemac53:
Today the InputSchema just takes a map, but we could make extension types for it so its easier to create schemas.

PR #40, opened 2025-04-08
Add Schema type for creating input schemas using typed interfaces

jakemac53:
Closes https://github.com/dart-lang/ai/issues/29

This isn't a full implementation of the spec but should cover all the common use cases. You can always just bring your own map here and cast it to the `ObjectSchema` type, if needed.

### DIFF
--- a/pkgs/dart_mcp/lib/src/api/tools.dart
+++ b/pkgs/dart_mcp/lib/src/api/tools.dart
@@ -218,13 +218,13 @@ extension type ObjectSchema.fromMap(Map<String, Object?> _value)
 
   /// A map of the properties of the object to the nested [Schema]s for those
   /// properties.
-  Map<String, Schema?>? get properties =>
-      (_value['properties'] as Map?)?.cast<String, Schema?>();
+  Map<String, Schema>? get properties =>
+      (_value['properties'] as Map?)?.cast<String, Schema>();
 
   /// A map of the property patterns of the object to the nested [Schema]s for
   /// those properties.
-  Map<String, Schema?>? get patternProperties =>
-      (_value['patternProperties'] as Map?)?.cast<String, Schema?>();
+  Map<String, Schema>? get patternProperties =>
+      (_value['patternProperties'] as Map?)?.cast<String, Schema>();
 
   /// A list of the required properties by name.
   List<String>? get required => (_value['required'] as List?)?.cast<String>();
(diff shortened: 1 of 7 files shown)


pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -122,26 +122,344 @@ extension type Tool.fromMap(Map<String, Object?> _value) {
...
+/// the sub-schemas directly to the specified type from the parent schema.
+///
+/// See https://json-schema.org/understanding-json-schema/reference for the full
+/// specification.
+///
+/// **Note:** Only a subset of the json schema spec is supported by these types,
+/// if you need something more complex you can create your own
+/// `Map<String, Object?>` and cast it to [Schema] (or [ObjectSchema]) directly.
+extension type Schema.fromMap(Map<String, Object?> _value) {
+  factory Schema.withCombinators({
```
natebosch, line 164: [nit] the `withCombinators` name feels a little clunky. Can we use the unnamed constructor for this one?

pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -122,26 +122,344 @@ extension type Tool.fromMap(Map<String, Object?> _value) {
...
+  int? get minProperties => _value['minProperties'] as int?;
+
+  /// The maximum number of properties in this object.
+  int? get maxProperties => _value['maxProperties'] as int?;
+}
+
+/// A JSON Schema definition for a String.
+extension type const StringSchema.fromMap(Map<String, Object?> _value)
+    implements Schema {
+  factory StringSchema({
```
natebosch, line 288: Can extension types have forwarding constructors? I think `Schema.string(...)` is a friendly API than `StringSchema(...)` because it has nice autocomplete after `Schema.`

natebosch (approved):

pkgs/dart_mcp/lib/src/api/tools.dart
jakemac53: Well, it only  makes sense to use this if you are using combinators, otherwise you should construct one of the subtypes.

pkgs/dart_mcp/lib/src/api/tools.dart
jakemac53: The main thing is you want to get the proper subtype of out it.

I could use static methods to achieve that API though.

jakemac53: I think if I did that approach I would also want to remove the regular constructors from the subtypes, so I don't have to duplicate them all?

pkgs/dart_mcp/lib/src/api/tools.dart
jakemac53: I renamed this to `combined` and documented it.

PR #42, opened 2025-04-08
[dart_tooling_mcp_server] Add a hot reload tool

kenzieschmoll:
Adds a server tool to perform a hot reload against a Flutter app.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
@@ -244,7 +244,7 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
 
   @visibleForTesting
   static final screenshotTool = Tool(
-    name: 'take_screenshot',
+    name: 'takeScreenshot',
     description: 'Takes a screenshot of the active flutter application in its '
         'current state. Requires "${connectTool.name}" to be successfully '
         'called first.',
@@ -253,7 +253,7 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
 
   @visibleForTesting
   static final hotReloadTool = Tool(
-    name: 'Hot Reload',
+    name: 'hotReload',
     description: 'Performs a hot reload of the active Flutter application. '
         'This is to apply the latest code changes to the running application. '
         'Requires "${connectTool.name}" to be successfully called first.',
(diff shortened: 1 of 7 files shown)


pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -97,6 +99,95 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
...
+            if (serviceName == 'reloadSources') {
+              // This may look something like 's0.reloadSources'.
+              hotReloadMethodName = e.method!;
+            }
+          }
+        });
+        await vmService.streamListen(EventStreams.kService);
+
+        // Short delay to allow the service registration event to be sent.
+        await Future<void>.delayed(const Duration(seconds: 1));
```
jakemac53, line 158: nit: make the hotReloadMethodName a completer and complete it in the onEvent listener, then await that here with a timeout

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -168,6 +232,25 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
     inputSchema: InputSchema(),
   );
 
+  static final _hotReloadTool = Tool(
+    name: 'hot_reload',
```
jakemac53, line 236: Note that this name ends up being user visible, so we actually probably want to make these more human readable (and spaces are fine)

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -164,13 +244,22 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
...
+    name: 'takeScreenshot',
     description: 'Takes a screenshot of the active flutter application in its '
         'current state. Requires "${connectTool.name}" to be successfully '
         'called first.',
     inputSchema: ObjectSchema(),
   );
 
+  @visibleForTesting
+  static final hotReloadTool = Tool(
+    name: 'hotReload',
```
kenzieschmoll, line 256: Having spaces like "Hot Reload" showed up in the chat window as "Hot_Reload" (at least in cursor), so I changed these to use camel case for now as that looks better in the chat window

jakemac53 (commented):
Can we reasonably test this? Maybe we should log a followup to test it once we can interact a bit more deeply with the app itself so we can tell if a change was applied?

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
jakemac53: Oh interesting ok

jakemac53 (approved):

PR #43, opened 2025-04-08
Add basic analysis support

jakemac53:
Requires clients to implement "Roots", and creates one analysis context per root.

Then, requests to analyze groups of files can be issued via the "analyze files" command, which requires a root for each group of files.

We also run a file watcher, which is a bit right now because we could be a second or so behind if using a polling watcher. We might want to try and do some sort of "file overlay" situation for other scenarios, passing the file contents to analyze directly.

This is currently based on the input schema branch - I will wait to merge until that is merged and then change the target branch here.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
@@ -37,6 +37,16 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
     return super.initialize(request);
   }
 
+  @override
+  Future<void> shutdown() async {
+    await super.shutdown();
+    await _analysisContexts?.dispose();
+    for (var subscription in _watchSubscriptions) {
+      unawaited(subscription.cancel());
+    }
+    _watchSubscriptions.clear();
+  }
+
   /// Lists the roots, and listens for changes to them.
   ///
   /// Whenever new roots are found, creates a new [AnalysisContextCollection].
@@ -65,7 +75,7 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
         throw ArgumentError.value(
             root.uri, 'uri', 'Only file scheme uris are allowed for roots');
       }
-      paths.add(p.normalize(uri.toFilePath()));
+      paths.add(p.normalize(uri.path));
     }
 
     for (var subscription in _watchSubscriptions) {
@@ -75,11 +85,11 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
 
     for (var rootPath in paths) {
       var watcher = DirectoryWatcher(rootPath);
-      watcher.events.listen((event) {
+      _watchSubscriptions.add(watcher.events.listen((event) {
         final context = _analysisContexts?.contextFor(event.path);
         if (context == null) return;
         context.changeFile(p.normalize(event.path));
-      });
+      }));
     }
 
     _analysisContexts =
@@ -120,12 +130,12 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
         ], isError: true);
       }
 
-      var context = contexts.contextFor(rootUri.toFilePath());
+      var context = contexts.contextFor(p.normalize(rootUri.path));
       await context.applyPendingFileChanges();
 
       for (var path in paths) {
-        var normalized = p.normalize(
-            p.isAbsolute(path) ? path : p.join(rootUri.toFilePath(), path));
+        var normalized =
+            p.normalize(p.isAbsolute(path) ? path : p.join(rootUri.path, path));
         var errorsResult = await context.currentSession.getErrors(normalized);
         if (errorsResult is! ErrorsResult) {
           return CallToolResult(content: [
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
@@ -8,6 +8,7 @@ import 'package:analyzer/dart/analysis/analysis_context_collection.dart';
 import 'package:analyzer/dart/analysis/results.dart';
 import 'package:dart_mcp/server.dart';
 import 'package:meta/meta.dart';
+import 'package:path/path.dart' as p;
 
 /// Mix this in to any MCPServer to add support for analyzing Dart projects.
 ///
@@ -55,7 +56,7 @@ base mixin DartAnalyzerSupport on ToolsSupport {
     _analysisContexts = AnalysisContextCollection(includedPaths: paths);
   }
 
-  /// Implementation of the [_analyzeFilesTool], analyzes the requested files
+  /// Implementation of the [analyzeFilesTool], analyzes the requested files
   /// under the requested project roots.
   Future<CallToolResult> _analyzeFiles(CallToolRequest request) async {
     var contexts = _analysisContexts;
@@ -91,8 +92,9 @@ base mixin DartAnalyzerSupport on ToolsSupport {
       }
 
       for (var path in paths) {
-        var errorsResult = await context.currentSession
-            .getErrors(rootUri.resolve(path).toFilePath());
+        var normalized = p.normalize(
+            p.isAbsolute(path) ? path : p.join(rootUri.toFilePath(), path));
+        var errorsResult = await context.currentSession.getErrors(normalized);
         if (errorsResult is! ErrorsResult) {
           return CallToolResult(content: [
             TextContent(text: 'Error computing analyzer errors $errorsResult'),
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:io';
 
 import 'package:analyzer/dart/analysis/analysis_context_collection.dart';
 import 'package:analyzer/dart/analysis/results.dart';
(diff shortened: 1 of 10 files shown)


bwilkerson:
Is there a reason why this is using the analyzer package directly rather than going through the analysis server?

The server already has the ability to watch files (so some of this code is duplication of existing code), and by not going through the server we prevent analyzer plugins from contributing to the results. The latter seems like an especially unfortunate outcome.

jakemac53:
So, I do think we might want to explore that in the future - especially if we can re-use the analysis server  that is already started by the IDE - but the main reason is that the language server protocol from what I could tell doesn't appear to be well suited to the task.

It appeared to me (happy to be corrected!) that the protocol is extremely tied to the IDE experience - errors etc are pushed to the IDE instead of pulled from the IDE. In this case we really want to just ask the analyzer for errors and get a result, not have them pushed to us asynchronously.

bwilkerson:
That's more true of LSP than it is of the legacy protocol.

We have the same need in the command-line analyzer. We solved that problem by having a notification from the server to indicate when analysis was complete and all of the diagnostics that are going to be generated have already been sent. That allows the command-line analyzer to know when to exit, and would allow this class to know when to return results to the mcp client.

If we want to use LSP then we'd probably need to add a custom protocol, either to ask for all of the diagnostics in a single operation or to notify clients when all the diagnostics have been computed. (We'll need to do that if we ever move the command-line analyzer to use LSP as well.)

jakemac53:
As discussed in the meeting today, we will run with the current approach for now, and once we have a better idea of the real functionality that we end up implementing here we will re-asses whether interacting over LSP is a better approach.

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -0,0 +1,182 @@
...
+          'This server requires the "roots" capability to be implemented.');
+    }
+    registerTool(analyzeFilesTool, _analyzeFiles);
+    initialized.then(_listenForRoots);
+    return super.initialize(request);
+  }
+
+  @override
+  Future<void> shutdown() async {
+    await super.shutdown();
```
kenzieschmoll, line 42: should super.shutdown get called last like we do for the dispose lifecycle method in flutter? Or should this be the first line in the method?

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -0,0 +1,182 @@
...
+      _createAnalysisContext(await listRoots(ListRootsRequest()));
+    });
+    _createAnalysisContext(await listRoots(ListRootsRequest()));
+  }
+
+  /// Creates an analysis context from a list of roots.
+  //
+  // TODO: Better configuration for the DART_SDK location.
+  void _createAnalysisContext(ListRootsResult result) async {
+    final sdkPath = Platform.environment['DART_SDK'];
```
kenzieschmoll, line 66: can we use Platform.resolvedExecutable to find this?

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -0,0 +1,182 @@
...
+        throw ArgumentError.value(
+            root.uri, 'uri', 'Only file scheme uris are allowed for roots');
+      }
+      paths.add(p.normalize(uri.path));
+    }
+
+    for (var subscription in _watchSubscriptions) {
+      unawaited(subscription.cancel());
+    }
+    _watchSubscriptions.clear();
```
kenzieschmoll, line 84: this is duplicated above, please create a helper

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -0,0 +1,182 @@
...
+
+    for (var rootPath in paths) {
+      var watcher = DirectoryWatcher(rootPath);
+      _watchSubscriptions.add(watcher.events.listen((event) {
+        final context = _analysisContexts?.contextFor(event.path);
+        if (context == null) return;
+        context.changeFile(p.normalize(event.path));
```
kenzieschmoll, line 91: nit: this can be reduced:
```
_analysisContexts?.contextFor(event.path)?.changeFile(p.normalize(event.path))
```

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: it doesn't matter, it just needs to get called

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: We can't because that doesn't work when AOT compiled (you just get the AOT script).

jakemac53: (and we have to AOT compile because pub breaks JSON servers communicating over stdin/stdout with compiling messages) cc @sigurdm @jonasfj this has been discussed in the past right?

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -0,0 +1,182 @@
...
+  /// Implementation of the [analyzeFilesTool], analyzes the requested files
+  /// under the requested project roots.
+  Future<CallToolResult> _analyzeFiles(CallToolRequest request) async {
+    var contexts = _analysisContexts;
+    if (contexts == null) {
+      return CallToolResult(content: [
+        TextContent(
+            text: 'Analysis not yet ready, please wait a few seconds and try '
+                'again.')
+      ], isError: true);
```
kenzieschmoll, line 108: Can we use and enforce the new Dart formatter on this codebase?

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -0,0 +1,182 @@
...
+        ], isError: true);
+      }
+
+      var context = contexts.contextFor(p.normalize(rootUri.path));
+      await context.applyPendingFileChanges();
+
+      for (var path in paths) {
+        var normalized =
+            p.normalize(p.isAbsolute(path) ? path : p.join(rootUri.path, path));
+        var errorsResult = await context.currentSession.getErrors(normalized);
```
kenzieschmoll, line 139: nit: we can consider enabling the prefer finals lint

pkgs/dart_tooling_mcp_server/test/dart_tooling_mcp_server_test.dart
```diff
@@ -34,4 +37,73 @@ void main() {
...
+      final result = await testHarness.callToolWithRetry(request);
+      expect(result.isError, isNot(true));
+      expect(result.content, isEmpty);
+    });
+
+    test('can handle project changes', () async {
+      final example =
+          d.dir('example', [d.file('main.dart', 'void main() => 1 + "2";')]);
+      await example.create();
+      final exampeRoot = Root(uri: example.io.absolute.uri.toString());
```
kenzieschmoll, line 71: typo "exampleRoot"

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: Previously the flutter stable we were getting didn't support it but it looks like it should now so I can try upping our min SDK again to do that. Will do it as a separate CL to reduce churn here.

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: filed https://github.com/dart-lang/ai/issues/46

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: (filed https://github.com/dart-lang/ai/issues/46)

kenzieschmoll (approved):

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
sigurdm: It should not print any compilation messages if no terminal is attached.
If that is not the case, please file an issue.

sigurdm: I don't know how this is gonna be deployed.
You could also consider compiling to a jit-snapshot via. `dart --snapshot=mcp.snapshot` and running with `dart mcp.snapshot`. No pub involved, and still knows the dart executable...

jakemac53: I think we want to deploy it as a single executable anyways, that makes the most sense, it is mostly annoying for development.

I definitely was running into this but IIRC it was using the MCP inspector and I don't know exactly how it was spawning things? For instance if you were to pass `runInShell: true` when spawning from a dart program, would it still print to stdout?

sigurdm: `runInShell: true` should not make it print to stdout.
You need to attach a pty to the process, something like https://github.com/TerminalStudio/pty for it to print.

(there might be bugs, but they should be fixable). The logic is here: https://github.com/dart-lang/pub/blob/639536aa78f7382b4a948669bfdc283f830b2ae9/lib/src/io.dart#L742

sigurdm: There is an env-var for overriding whatever behavior you are otherwise getting...

Issue #44, opened 2025-04-09
[productivity] Update test harness to run faster


jakemac53:
- Share the compiled MCP server across tests
- Don't always require an attached flutter app (for the analyze tests for instance)

PR #45, opened 2025-04-09
refactor the test harness to separate out debug sessions

jakemac53:
Part of https://github.com/dart-lang/ai/issues/44

- No longer always start up the counter app when starting a debug session, instead we allow you to start a debug session explicitly as needed. Allows for multiple debug sessions to be registered so we can test those workflows in the future.
- Starts an explicit DTD service of our own instead of getting one from `flutter run` or `dart run`. This mimics more the IDE environment, and allows for one connection with multiple apps.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/test/test_harness.dart
+++ b/pkgs/dart_tooling_mcp_server/test/test_harness.dart
@@ -37,7 +37,8 @@ class TestHarness {
     this.dtdUri,
   );
 
-  /// Starts a flutter process as well as an MCP client and server.
+  /// Starts a Dart Tooling Daemon as well as an MCP client and server, and
+  /// a [FakeEditorExtension] to manage registering debug sessions.
   ///
   /// Handles the initialization handshake between the MCP client and server as
   /// well.
@@ -47,6 +48,8 @@ class TestHarness {
   /// breakpoints in the server don't work however, so you can set [debugMode]
   /// to `true` and we will run it in process instead, which allows breakpoints
   /// since everything is running in the same isolate.
+  ///
+  /// Use [startDebugSession] to start up apps and connect to them.
   static Future<TestHarness> start({
     @Deprecated('For debugging only, do not submit') bool debugMode = false,
   }) async {
--- a/pkgs/dart_tooling_mcp_server/test/test_harness.dart
+++ b/pkgs/dart_tooling_mcp_server/test/test_harness.dart
@@ -122,8 +122,7 @@ class TestHarness {
   }
 }
 
-/// The debug session for a single app - has its own dtd instance and
-/// vm service URI.
+/// The debug session for a single app.
 ///
 /// Should be started using [TestHarness.startDebugSession].
 final class AppDebugSession {
--- a/pkgs/dart_tooling_mcp_server/test/test_harness.dart
+++ b/pkgs/dart_tooling_mcp_server/test/test_harness.dart
@@ -25,15 +25,15 @@ import 'package:test_process/test_process.dart';
 ///   allow for breakpoints, but the default mode is to run the server in a
 ///   separate process.
 class TestHarness {
-  final TestProcess flutterProcess;
+  final FakeEditorExtension fakeEditorExtension;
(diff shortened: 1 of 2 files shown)


pkgs/dart_tooling_mcp_server/test/test_harness.dart
```diff
@@ -47,40 +48,34 @@ class TestHarness {
...
     final connection = await _initializeMCPServer(mcpClient, debugMode);
 
-    final fakeEditorExtension = await FakeEditorExtension.connect(
-      flutterProcess,
-      dtdUri,
-    );
+    final dtdProcess = await TestProcess.start('dart', ['tooling-daemon']);
+    final dtdUri = await _getDTDUri(dtdProcess);
+
+    final fakeEditorExtension = await FakeEditorExtension.connect(dtdUri);
```
kenzieschmoll, line 63: technically to mimic the IDE environment, the fake editor should start DTD and pass the workspace roots (this is a protected operation that only the IDE, or the thing that starts DTD, can do). Idk if this difference matters right now or not, but if workspace roots need to be set on DTD for anything, it may.

jakemac53: It probably doesn't ultimately matter right now - is that protected by the extra "client secret" that I saw it log?

jakemac53: Fwiw MCP also has a notion of project roots, so right now the MCP server is asking the client for that not DTD

pkgs/dart_tooling_mcp_server/test/test_harness.dart
```diff
@@ -119,6 +114,81 @@ class TestHarness {
   }
 }
 
+/// The debug session for a single app - has its own dtd instance and
```
kenzieschmoll, line 117: do we want this to continue having its own DTD instance? If we want to support a multi-app use case, then we'd want 1 dtd instance to 1 or many vm service connections

jakemac53: Sorry this comment is out of date, will remove :). I think we want always one primary DTD instance

pkgs/dart_tooling_mcp_server/test/test_harness.dart
kenzieschmoll: > is that protected by the extra "client secret"

Yeah. So only the IDE who has the key can set these.

jakemac53: Yeah ok, so we do have access to it and could plumb it through if we need to

pkgs/dart_tooling_mcp_server/test/test_harness.dart
```diff
@@ -122,18 +122,88 @@ class TestHarness {
...
+        ? 'linux'
+        : Platform.isMacOS
+            ? 'macos'
+            : throw StateError(
+                'unsupported platform, only mac and linux are supported',
+              );
+    final process = await TestProcess.start(
+      isFlutter ? 'flutter' : 'dart',
+      [
+        'run',
```
kenzieschmoll, line 154: this will still start a second DTD instance in the case of a flutter or dart run:

Flutter Tools --> DevTools Server --> DTD
Dart CLI --> DDS --> DTD

jakemac53: Right, but we are just ignoring it right now. My assumption is the IDE also does exactly this and just ignores the DTD instance that comes from flutter/dart run.

kenzieschmoll: The IDEs start Dart or Flutter apps and pass the `--no-serve-devtools` (or `--no-dds`?) which results in DTD not being started. 
https://github.com/dart-lang/sdk/blob/main/pkg/dds/lib/src/dds_impl.dart#L197

We could take the same approach here by passing the right flag but we'd need to check what the IDEs are doing currently.

kenzieschmoll (approved):

PR #48, opened 2025-04-10
[dart_tooling_mcp_server] add Dart CLI tool support

kenzieschmoll:
This PR adds tool support for the following Dart CLI tools:
- Running `dart fix` over a set of roots
- Running `dart format` over a set of roots

This PR also adds the directory `test_fixtures/dart_cli_app` which contains Dart scripts for use in tests.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
@@ -1,0 +1,159 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:dart_mcp/server.dart';
+
+// TODO: migrate the analyze files tool to use this mixin and run the
+// `dart analyze` command instead of the analyzer package.
+
+/// Mix this in to any MCPServer to add support for running Dart CLI commands
+/// like `dart fix` and `dart format`.
+///
+/// The MCPServer must already have the [ToolsSupport] and [LoggingSupport]
+/// mixins applied.
+base mixin DartCliSupport on ToolsSupport, LoggingSupport {
+  @override
+  FutureOr<InitializeResult> initialize(InitializeRequest request) {
+    if (request.capabilities.roots == null) {
+      throw StateError(
+        'This server requires the "roots" capability to be implemented.',
+      );
+    }
+    registerTool(dartFixTool, _runDartFixTool);
+    registerTool(dartFormatTool, _runDartFormatTool);
+    return super.initialize(request);
+  }
+
+  /// Implementation of the [dartFixTool].
+  Future<CallToolResult> _runDartFixTool(CallToolRequest request) async {
+    return _runDartCommandInRoots(request, 'dart fix', ['fix', '--apply']);
+  }
+
+  /// Implementation of the [dartFormatTool].
+  Future<CallToolResult> _runDartFormatTool(CallToolRequest request) async {
+    return _runDartCommandInRoots(request, 'dart format', ['format', '.']);
+  }
+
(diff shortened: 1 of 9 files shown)


pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
```diff
@@ -0,0 +1,159 @@
...
+        runInShell: true,
+      );
+
+      final output = (result.stdout as String).trim();
+      final errors = (result.stderr as String).trim();
+      if (result.exitCode != 0) {
+        return CallToolResult(
+          content: [
+            TextContent(
+              text: '$commandName failed in ${projectRoot.path}:\n$errors',
```
jakemac53, line 99: Probably should write stdout as well here?

jakemac53 (approved):

pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
kenzieschmoll: Won't writing to standard out mess up the stdio output for the MCP client?

kenzieschmoll:
I added an additional test_fixture after your review and fixed up the tests. PTAL.

pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
jakemac53: I just mean adding the `output` variable above to this text - not writing it to actual stdout.

pkgs/dart_tooling_mcp_server/test/dart_tooling_mcp_server_test.dart
```diff
@@ -162,72 +191,115 @@ void main() {
       );
     });
 
+    tearDown(() async {
+      // Delete the copy.
+      await Directory(copyDartCliAppsBin).delete(recursive: true);
```
jakemac53, line 196: nit: I would move this into the `copyDartCliAppAndVerifyContents` function - using `addTearDown`. That way you don't have to remember to do this whenever using that function.

pkgs/dart_tooling_mcp_server/test/dart_tooling_mcp_server_test.dart
```diff
@@ -151,8 +154,34 @@ void main() {
   group('dart cli', () {
     late Tool dartFixTool;
     late Tool dartFormatTool;
+    late String copyDartCliAppsBin;
+    const copyDirectoryName = 'bin_copy';
+
+    Future<void> copyDartCliAppAndVerifyContents() async {
+      // Copy the original `dart_cli_app` directory so that changes from these
```
jakemac53, line 161: Consider just copying the entire test fixture dir, but to a temp directory.

That way if a test gets into an unrecoverable state that has to be force killed you don't have to manually clean up this dir.

kenzieschmoll: For whatever reason I was running into issues with temporary directories. I originally tried to copy the entire `dart_cli_app` dir to a temporary directory but then the `dart fix` command wouldn't detect any fixes. Idk if this is because there wasn't a pubspec.yaml anywhere or what, but I eventually just went back to copy in place. I will add to .gitignore and add a todo in the test in case this becomes an issue later

pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
```diff
@@ -0,0 +1,159 @@
...
+      properties: {
+        'roots': ListSchema(
+          title: 'All projects roots to run dart format in.',
+          description:
+              'These must match a root returned by a call to "listRoots".',
+          items: ObjectSchema(
+            properties: {
+              'root': StringSchema(
+                title: 'The URI of the project root to run `dart format` in.',
+              ),
```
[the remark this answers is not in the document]
kenzieschmoll: done.

PR #49, opened 2025-04-10
start on an example dash chat client

jakemac53:
Adds a basic command line chat bot client example, accepts stdio servers to connect to as command line arguments.

### DIFF
--- a/pkgs/dart_mcp/example/dash_client.dart
+++ b/pkgs/dart_mcp/example/dash_client.dart
@@ -97,7 +97,7 @@ final class DashClient extends MCPClient with RootsSupport {
     }
   }
 
-  // Prints `text` and adds it to the chat history
+  /// Prints `text` and adds it to the chat history
   Future<void> _chatToUser(String text) async {
     final dashSpeakResponse =
         (await model.generateContent([
--- a/pkgs/dart_mcp/example/dash_client.dart
+++ b/pkgs/dart_mcp/example/dash_client.dart
@@ -25,7 +25,7 @@ void main(List<String> args) {
 
   final parsedArgs = argParser.parse(args);
   final serverCommands = parsedArgs['server'] as List<String>;
-  GeminiClient(serverCommands, geminiApiKey: geminiApiKey);
+  DashClient(serverCommands, geminiApiKey: geminiApiKey);
 }
 
 final argParser =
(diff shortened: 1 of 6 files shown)


pkgs/dart_mcp/example/gemini_client.dart
```diff
@@ -0,0 +1,287 @@
...
+        if (part is gemini.TextPart) {
+          chatToUser(part.text);
+        } else if (part is gemini.FunctionCall) {
+          final userResponse = StringBuffer();
+          chatToUser(
+            'It looks like you want to invoke tool ${part.name} with args '
+            '${jsonEncode(part.args)}, is that correct? (y/n)',
+          );
+          final answer = await stdinQueue.next;
+          chatHistory.add(gemini.Content.text('Yes'));
```
jakemac53, line 110: Lmao the risks of AI coding 🤣. I will fix this tomorrow

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -183,7 +185,7 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
 
   @visibleForTesting
   static final analyzeFilesTool = Tool(
-    name: 'analyze files',
+    name: 'analyzeFiles',
```
jakemac53, line 188: Gemini doesn't allow names with spaces

pkgs/dart_mcp/example/gemini_client.dart
```diff
@@ -0,0 +1,291 @@
...
+          case gemini.FunctionCall():
+            await _handleFunctionCall(part);
+          default:
+            throw UnimplementedError('Unhandled response type $modelResponse');
+        }
+      }
+    }
+  }
+
+  // Prints `text` and adds it to the chat history
```
kenzieschmoll, line 106: nit: ///

kenzieschmoll (commented):
can you add readme instructions for how use this client for development?

pkgs/dart_mcp/example/dash_client.dart
```diff
@@ -35,22 +35,23 @@ final argParser =
...
-  GeminiClient(this.serverCommands, {required String geminiApiKey})
-    : stdinQueue = StreamQueue(
-        stdin.transform(utf8.decoder).transform(const LineSplitter()),
-      ),
-      model = gemini.GenerativeModel(
-        model: 'gemini-2.5-pro-exp-03-25',
+  DashClient(this.serverCommands, {required String geminiApiKey})
+    : model = gemini.GenerativeModel(
+        // model: 'gemini-2.5-pro-exp-03-25',
+        model: 'gemini-2.0-flash',
```
kenzieschmoll, line 49: which model do we want to use? should we delete the commented out one?

kenzieschmoll (approved):

pkgs/dart_mcp/example/dash_client.dart
jakemac53: Its handy to be able to swap back and forth.... flash is a lot faster so I find it better for this purpose. But its not as good at analyzing images for instance.

jakemac53: Going to go ahead and land this as is - we should probably have a followup which just allows you to select a model on the fly.

PR #53, opened 2025-04-11
Add skeleton for `dev_mcp_host` DevTools extension

kenzieschmoll:
Note for reviewers: the `build/` directory is generated code and does not need to be reviewed. This is the built assets of the dev_mcp_host flutter web app (pre requisite for loading the DevTools extension in a real DevTools-hosted environment)


pkgs/dev_mcp_host/devtools_options.yaml
```diff
@@ -0,0 +1,4 @@
+description: This file stores settings for Dart & Flutter DevTools.
+documentation: https://docs.flutter.dev/tools/devtools/extensions#configure-extension-enablement-states
+extensions:
+  - dev_mcp_host: true
```
jakemac53, line 4: A fair number of these files are missing trailing newlines, would be nice to get those in

jakemac53 (approved):

PR #62, opened 2025-04-14
[dart_tooling_mcp_server] Persist VmService objects

kenzieschmoll:
This PR:
- persists active VM service objects so that we do not need to recreate them for each DTD tool call
- cleans up a stream subscription disposal for the hot reload command (review with hide whitespace diffs)

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
@@ -56,6 +56,9 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
     final response = await dtd.getDebugSessions();
     final debugSessions = response.debugSessions;
     for (final debugSession in debugSessions) {
+      if (activeVmServices.containsKey(debugSession.vmServiceUri)) {
+        continue;
+      }
       final vmService = await vmServiceConnectUri(debugSession.vmServiceUri);
       activeVmServices[debugSession.vmServiceUri] = vmService;
       unawaited(
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
@@ -29,19 +29,19 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
   ///
   /// [VmService] objects are automatically removed from the Map when the
   /// [VmService] shuts down.
-  final activeVmServices = <String, VmService>{};
+  final _activeVmServices = <String, VmService>{};
 
   /// Called when the DTD connection is lost, resets all associated state.
   Future<void> _resetDtd() async {
     _dtd = null;
     _getDebugSessionsReady = false;
     await Future.wait(
-      activeVmServices.values.map((vmService) => vmService.dispose()),
+      _activeVmServices.values.map((vmService) => vmService.dispose()),
     );
-    activeVmServices.clear();
+    _activeVmServices.clear();
   }
 
-  Future<void> updateActiveVmServices() async {
+  Future<void> _updateActiveVmServices() async {
     final dtd = _dtd;
     if (dtd == null) return;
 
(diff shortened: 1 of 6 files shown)


pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -24,10 +24,50 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
   /// ready to be invoked.
   bool _getDebugSessionsReady = false;
 
+  /// A Map of [VmService] objects by their associated VM Service URI
+  /// (represented as a String).
+  ///
+  /// [VmService] objects are automatically removed from the Map when the
+  /// [VmService] shuts down.
+  final _activeVmServices = <String, VmService>{};
```
kenzieschmoll, line 32: @jakemac53 it doesn't look like there is a good way to test the logic inside of the tool mixins. All that is available from the test harness is calling the registered tools. Do you see any issues with running some tests in process so that we can test the implementation logic inside the server tools? I see [here](https://github.com/dart-lang/ai/blob/main/pkgs/dart_tooling_mcp_server/test/test_harness.dart#L331) that we create the DartToolingMCPServer in process, but this is currently limited to `debugMode` only.

jakemac53: I think it's fine if we want to run some tests this way explicitly so we can inspect the server state - we just don't want to run all tests that way IMO.

I would probably rename the parameter though in that case.

jakemac53: Compiling the MCP server right now is also likely most of the test time, since it's happening for each test.

We might want to consider moving more tests to the in process mode and just having a single end to end test which does some basic interactions (just list the tools for instance). Really we just need to validate that the server actually works as a separate process.

pkgs/dart_tooling_mcp_server/test/tools/dtd_test.dart
```diff
@@ -75,4 +87,55 @@ void main() {
...
+      expect(server.activeVmServices.length, 1);
+      expect(originalVmService, server.activeVmServices.values.single);
+
+      // await testHarness.startDebugSession(
+      //   counterAppPath,
+      //   'lib/main.dart',
+      //   isFlutter: true,
+      // );
+      // await server.updateActiveVmServices();
+      // expect(server.activeVmServices.length, 2);
```
kenzieschmoll, line 121: @jakemac53 when I uncomment this code, I get the following error:

```
  Bad state: Unable to find the context to /Users/kenzieschmoll/develop/dart-lang/ai/pkgs/dart_tooling_mcp_server/test_fixtures/counter_app/.dart_tool/dartpad/web_plugin_registrant.dart
  package:analyzer/src/dart/analysis/analysis_context_collection.dart 145:5  AnalysisContextCollectionImpl.contextFor
  package:dart_tooling_mcp_server/src/mixins/analyzer.dart 98:17             DartAnalyzerSupport._createAnalysisContext.<fn>
```

Is this expected or is this a bug in the test harness? With the current implementation should we expect running a second debug session to fail?

jakemac53: It sounds like possibly an error somewhere... `startDebugSession` does add a new root dir to the client - and the server will then ask for the new roots and set up new file listeners.

I think this might be hitting a race condition where we are getting a file change notification after the context has been disposed and before the new one for that dir has been set up.

jakemac53: Hmm, looking into this a bit and I wonder if its actually that the path is not analyzed by default because its in a hidden dir?

jakemac53: For now maybe its best to just guard this in a try/catch https://github.com/dart-lang/ai/blob/main/pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart#L97

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -24,10 +24,50 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
...
+
   /// Called when the DTD connection is lost, resets all associated state.
-  void _resetDtd() {
+  Future<void> _resetDtd() async {
     _dtd = null;
     _getDebugSessionsReady = false;
+    await Future.wait(
```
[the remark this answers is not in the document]
kenzieschmoll: done.

pkgs/dart_tooling_mcp_server/test/tools/dtd_test.dart
kenzieschmoll: Added a try / catch and filed https://github.com/dart-lang/ai/issues/65.

jakemac53 (approved):

PR #68, opened 2025-04-15
Add resource template support

jakemac53:
See https://modelcontextprotocol.io/specification/2024-11-05/server/resources#resource-templates

The support is pretty low level right now, and we do not provide any assistance for matching the provided templates or extracting out the values.

We likely should add that support at some point later on, but we should be able to do so in a non-breaking manner. We will simply stop invoking the `handlers` for URIs that can't possibly match the template.


jakemac53:
Note that I initially was going to try and use package:uri for this but couldn't get that package to function for the use cases we care about (package: scheme uris).

kenzieschmoll (approved):

PR #71, opened 2025-04-16
[dart_tooling_mcp_server] add tool for getting the selected widget

kenzieschmoll:
This PR adds a tool that gets the selected widget from the Widget Inspector for a running Flutter app. In its current form, this tool calls the `ext.flutter.inspector.getSelectedSummaryWidget` service extension. I opted to use `getSelectedSummaryWidget` instead of `getSelectedWidget` because this tool will likely be useful in the context of the user's code.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
@@ -433,9 +433,7 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
           final widget = result.json?['result'];
           if (widget == null) {
             return CallToolResult(
-              content: [
-                TextContent(text: 'No Widget selected.'),
-              ],
+              content: [TextContent(text: 'No Widget selected.')],
             );
           }
           return CallToolResult(
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
@@ -439,7 +439,7 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
             );
           }
           return CallToolResult(
-            content: [TextContent(text: widget.toString())],
+            content: [TextContent(text: jsonEncode(widget))],
           );
         } catch (e) {
           return CallToolResult(
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
@@ -434,7 +434,7 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
           if (widget == null) {
             return CallToolResult(
               content: [
-                TextContent(text: 'No Widget selected: ${result.json}'),
+                TextContent(text: 'No Widget selected.'),
               ],
             );
           }
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert';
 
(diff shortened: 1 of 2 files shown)


pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -394,11 +408,43 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
...
+  // TODO: support passing a debug session id when there is more than one debug
+  // session.
+  Future<CallToolResult> selectedWidget(CallToolRequest request) async {
+    return _callOnVmService(
+      callback: (vmService) async {
+        final vm = await vmService.getVM();
+        final isolateId = vm.isolates!.first.id;
+        try {
+          final result = await vmService.callServiceExtension(
+            '$_inspectorServiceExtensionPrefix.getSelectedSummaryWidget',
```
jakemac53, line 428: How does this determine what things are "user code"? Do we need to call `addPubRootDirectories`? Or do we assume something else is?

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -394,11 +408,43 @@ base mixin DartToolingDaemonSupport on ToolsSupport {
...
+          final widget = result.json?['result'];
+          if (widget == null) {
+            return CallToolResult(
+              content: [
+                TextContent(text: 'No Widget selected.'),
+              ],
+            );
+          }
+          return CallToolResult(
+            content: [TextContent(text: widget.toString())],
```
jakemac53, line 442: Should we do this instead?

```suggestion
            content: [TextContent(text: jsonEncode(widget)],
```

Or maybe even:

```suggestion
            content: [TextContent(text: JsonEncoder.withIndent(' ').encode(widget)],
```

jakemac53 (commented):
LGTM with some nits - should we also add enabling/disabling the widget inspector?

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
kenzieschmoll: The Dart Code extension / Flutter IntelliJ plugin already do this today. Otherwise we assume the pub root directories is the root of the running app. CC @elliette to confirm this is how things still work today.

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
kenzieschmoll: Will the added new lines caused by using .withIndent have any impact on the LLM trying to parse this information? In terms of using up more tokens than necessary?

kenzieschmoll:
> should we also add enabling/disabling the widget inspector

Good question. It's not immediately clear to me what the use case for this would be. Do you have one in mind?

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
elliette: Yes, that matches my understanding. Though it looks like the call to `addPubRootDirectories` was removed from the IntelliJ plugin: https://github.com/flutter/flutter-intellij/commit/1c1389fd10030487fa9802badcb64ca4aeb84688

FYI @pq @jwren this is another piece of code that might need to be added back

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
jakemac53: It probably is just a waste of tokens yes, no idea how it might affect its ability to reason about it either positively or negatively 🤷‍♂️ .

Just jsonEncode SGTM though

jakemac53 (approved):

jakemac53:
> Good question. It's not immediately clear to me what the use case for this would be. Do you have one in mind?

Basically the idea is it would give the LLM the ability to decide that it should be asking the user to select a widget based on some query, and handle enabling the inspector for them and then prompt them to select the widget.

PR #73, opened 2025-04-18
Add an example client to execute a workflow

kenzieschmoll:


jakemac53 (commented):
Nice, we should move some of this logic out to a shared library before landing though :D.

pkgs/dart_mcp/example/workflow_client.dart
```diff
@@ -405,39 +383,36 @@ final class WorkflowClient extends MCPClient with RootsSupport {
...
+
+Generally, if you are asked to make code changes, you should follow this high
+level process:
+
+1) Write the code and apply the changes to the codebase
+2) Check for static analysis errors and warnings
+3) Fix all errors and warnings that you can
+4) Hot reload the changes to the running app
+5) Check for runtime errors
+6) Fix any runtime errors that you can
```
jakemac53, line 413: Should it ask the user at this point to try out the app, and then it can listen for any additional runtime errors, or the user can give other feedback before proceeding?

kenzieschmoll: Once it is done it does ask if there is anything else the user would like. Often times the steps it generates checks for runtime errors a second time after applying changes.

kenzieschmoll:
> we should move some of this logic out to a shared library before landing though :D.

Filed https://github.com/dart-lang/ai/issues/86 to do this as a follow up.

pkgs/dart_mcp/example/workflow_client.dart
```diff
@@ -66,6 +80,22 @@ final class WorkflowClient extends MCPClient with RootsSupport {
...
+        'The current working directory is '
+        '${Directory.current.absolute.uri.toString()}. Convert all relative '
+        'URIs to absolute using this root. For tools that want a root, use this'
+        'URI.',
+      ),
+    );
+    if (dtdUri != null) {
+      chatHistory.add(
+        gemini.Content.text(
+          'If you need to establish a Dart Tooling Daemon (DTD) connection, '
```
jakemac53, line 94: Fwiw, this could even be just "connect to DTD at $dtdUri"

jakemac53 (approved):

Issue #65, opened 2025-04-14
Bad state: Unable to find the context to hidden file.


kenzieschmoll:
Example failure:
```
Bad state: Unable to find the context to /Users/kenzieschmoll/develop/dart-lang/ai/pkgs/dart_tooling_mcp_server/test_fixtures/counter_app/.dart_tool/dartpad/web_plugin_registrant.dart package:analyzer/src/dart/analysis/analysis_context_collection.dart 145:5 AnalysisContextCollectionImpl.contextFor package:dart_tooling_mcp_server/src/mixins/analyzer.dart 98:17
```

See https://github.com/dart-lang/ai/pull/62/files#r2042859164 for context.

PR #74, opened 2025-04-21
Update analyze tool to use LSP, simplify tool

jakemac53:
- Use `dart language-server` for errors, file watching, etc.
- Drop analyzer dependency.
- Depends on language_server_protocol package from the SDK using a git dependency for now:
  - We can fix this by moving into the SDK
  - Or publishing language_server_protocol
- The tool now just always returns all errors for all known workspaces.
- Errors are returned in a structured format right now instead of as a human readable string.
 - We might want to steal some code from analyzer/DartCode etc to get a human readable string going.
 - Or if this is just for AI input, structured may even be better.
 
 In the future, we might try to use the DTD connection to talk, which won't change this code too much. But right now DTD has very limited support (read/write properties).
 
 Closes https://github.com/dart-lang/ai/issues/65 as well

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/lsp/wire_format.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/lsp/wire_format.dart
@@ -8,6 +8,7 @@ import 'dart:convert';
 import 'package:async/async.dart';
 import 'package:stream_channel/stream_channel.dart';
 
+/// Handles LSP communication with its associated headers.
 StreamChannel<String> lspChannel(
   Stream<List<int>> stream,
   StreamSink<List<int>> sink,
@@ -23,6 +24,9 @@ StreamChannel<String> lspChannel(
   return StreamChannel.withGuarantees(parser.stream, outSink);
 }
 
+/// Writes [data] to [sink], with the appropriate content length header.
+///
+/// Writes the [data] in 1KB chunks.
 void _serialize(String data, EventSink<List<int>> sink) {
   final message = utf8.encode(data);
   final header = 'Content-Length: ${message.length}\r\n\r\n';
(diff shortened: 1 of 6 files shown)


pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -41,214 +54,221 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
...
+    _lspConnection = Peer(channel);
+
+    _lspConnection.registerMethod(
+      lsp.Method.textDocument_publishDiagnostics.toString(),
+      _handleDiagnostics,
+    );
+
+    _lspConnection.registerMethod(r'$/analyzerStatus', _handleAnalyzerStatus);
+
+    _lspConnection.registerFallback((Parameters params) {
```
kenzieschmoll, line 111: nit: use cascades for all these:
```dart
_lspConnection = Peer(channel)
  ..registerMethod(...)
  ..registerMethod(...)
  ..registerFallback(...);
```

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -41,214 +54,221 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
...
+      log(LoggingLevel.debug, 'fallback: ${params.method} - ${params.asMap}');
+    });
+
+    unawaited(_lspConnection.listen());
+
+    log(LoggingLevel.debug, 'Connecting to analyzer lsp server');
+    lsp.InitializeResult? initializeResult;
+    String? error;
+    try {
+      /// Initialize with the server.
```
kenzieschmoll, line 121: // for comments

kenzieschmoll (approved):
LGTM, but would be good to wait for @bwilkerson's review as well.

bwilkerson (approved):

pkgs/dart_tooling_mcp_server/lib/src/lsp/wire_format.dart
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:async/async.dart';
+import 'package:stream_channel/stream_channel.dart';
```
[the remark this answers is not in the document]
jakemac53: Yeah I would probably move it to be a part of the `language_server_protocol` package.

PR #75, opened 2025-04-21
Add `ProcessManagerSupport` mixin to make CLI command end points more testable

kenzieschmoll:
This allows us to not duplicate testing that is already done for CLI tools this server is calling into (like the Dart CLI for example).

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/server.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/server.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 
 import 'package:dart_mcp/server.dart';
 import 'package:meta/meta.dart';
+import 'package:process/process.dart';
 import 'package:stream_channel/stream_channel.dart';
 
 import 'mixins/analyzer.dart';
@@ -16,6 +17,7 @@ import 'utils/process_manager.dart';
 /// An MCP server for Dart and Flutter tooling.
 final class DartToolingMCPServer extends MCPServer
     with
+        ProcessManagerSupport,
         LoggingSupport,
         ToolsSupport,
         DartAnalyzerSupport,
@@ -23,7 +25,8 @@ final class DartToolingMCPServer extends MCPServer
         DartToolingDaemonSupport {
   DartToolingMCPServer({
     required super.channel,
-    @visibleForTesting ProcessManager processManager = const ProcessManager(),
+    @visibleForTesting
+    LocalProcessManager processManager = const LocalProcessManager(),
   }) : super.fromStreamChannel(
          implementation: ServerImplementation(
            name: 'dart and flutter tooling',
@@ -36,11 +39,13 @@ final class DartToolingMCPServer extends MCPServer
     _processManager = processManager;
   }
 
-  static late final ProcessManager _processManager;
-
   static Future<DartToolingMCPServer> connect(
     StreamChannel<String> mcpChannel,
   ) async {
     return DartToolingMCPServer(channel: mcpChannel);
   }
+
+  @override
+  LocalProcessManager get processManager => _processManager;
+  late final LocalProcessManager _processManager;
 }
--- a/pkgs/dart_tooling_mcp_server/lib/src/server.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/server.dart
@@ -5,11 +5,13 @@
 import 'dart:async';
 
 import 'package:dart_mcp/server.dart';
+import 'package:meta/meta.dart';
 import 'package:stream_channel/stream_channel.dart';
 
 import 'mixins/analyzer.dart';
 import 'mixins/dart_cli.dart';
 import 'mixins/dtd.dart';
+import 'utils/process_manager.dart';
 
 /// An MCP server for Dart and Flutter tooling.
 final class DartToolingMCPServer extends MCPServer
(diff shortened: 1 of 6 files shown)


pkgs/dart_tooling_mcp_server/lib/src/server.dart
```diff
@@ -5,34 +5,47 @@
...
 
   static Future<DartToolingMCPServer> connect(
     StreamChannel<String> mcpChannel,
   ) async {
     return DartToolingMCPServer(channel: mcpChannel);
   }
+
+  @override
+  LocalProcessManager get processManager => _processManager;
+  late final LocalProcessManager _processManager;
```
jakemac53, line 50:
```suggestion
  final LocalProcessManager processManager;
```

Plus the associated change above

pkgs/dart_tooling_mcp_server/lib/src/server.dart
```diff
@@ -5,34 +5,47 @@
...
-          version: '0.1.0-wip',
-        ),
-        instructions:
-            'This server helps to connect Dart and Flutter developers to '
-            'their development tools and running applications.',
-      );
+  DartToolingMCPServer({
+    required super.channel,
+    @visibleForTesting
+    LocalProcessManager processManager = const LocalProcessManager(),
```
jakemac53, line 29:
```suggestion
    this.processManager = const LocalProcessManager(),
```

plus change below, and then drop the constructor body entirely.

pkgs/dart_tooling_mcp_server/lib/src/utils/process_manager.dart
```diff
@@ -0,0 +1,19 @@
...
+
+/// A mixin that provides a single getter of type [LocalProcessManager].
+///
+/// The `DartToolingMCPServer` class mixes this in so that [Process] methods
+/// can be easily mocked during testing.
+///
+/// MCP support mixins like `DartCliSupport` that spawn processes should use
+/// [processManager] from this mixin instead of making direct calls to dart:io's
+/// [Process] class.
+mixin ProcessManagerSupport {
```
jakemac53, line 17: nit: This is really an interface not a mixin

```suggestion
abstract interface class ProcessManagerSupport {
```

pkgs/dart_tooling_mcp_server/test/test_harness.dart
```diff
@@ -378,3 +383,29 @@ Root rootForPath(String projectPath) =>
 final counterAppPath = p.join('test_fixtures', 'counter_app');
 
 final dartCliAppsPath = p.join('test_fixtures', 'dart_cli_app');
+
+/// A test wrapper around [LocalProcessManager] that stores commands locally
+/// instead of running them by spawning sub-processes.
+class TestProcessManager extends LocalProcessManager {
+  final commands = <String>[];
```
jakemac53, line 390: nit: commandsRan?

We might eventually want to make this a stream queue or something, with some helpers like `expectCommand` but this should be fine for now.

jakemac53, line 390: Should this be a `List<String>`? It probably doesn't matter much which it is, but I would generally tend to keep it as close to the original representation as possible.

pkgs/dart_tooling_mcp_server/test/tools/dart_cli_test.dart
```diff
@@ -2,68 +2,39 @@
...
+  // This root is arbitrary for these tests since we are not actually running
+  // the CLI commands, but rather sending them through the
+  // [TestProcessManager] wrapper.
+  final testRoot = rootForPath(counterAppPath);
 
   // TODO: Use setUpAll, currently this fails due to an apparent TestProcess
   // issue.
   setUp(() async {
-    testHarness = await TestHarness.start();
+    testHarness = await TestHarness.start(inProcess: true);
```
jakemac53, line 23: Just checking we intentionally are converting this one to run in process (I think its fine, but I want to make sure we do leave at least one test running not in process somewhere).

pkgs/dart_tooling_mcp_server/test/tools/dart_cli_test.dart
```diff
@@ -2,68 +2,39 @@
...
+    testProcessManager =
+        testHarness.serverConnectionPair.server!.processManager
+            as TestProcessManager;
+    addTearDown(() => testProcessManager.reset());
```
jakemac53, line 27: Should we just call `addTearDown(reset)` in the constructor of `TestProcessManager` to do this?

pkgs/dart_tooling_mcp_server/test/tools/dart_cli_test.dart
kenzieschmoll: Yeah, otherwise we won't be able to access the `DartToolingMCPServer.processManager` variable to check the commands that are ran during the test.

pkgs/dart_tooling_mcp_server/test/test_harness.dart
kenzieschmoll: I changed to `List<String>` so that it was easy to check equality in tests.

kenzieschmoll: Changed to `List<Object>`

pkgs/dart_tooling_mcp_server/test/tools/dart_cli_test.dart
jakemac53: Oh right, makes sense 👍

PR #76, opened 2025-04-21
[dart_tooling_mcp_server] Add a dart pub tool

kenzieschmoll:
The initial set of supported pub commands are `add`, `remove`, `get`, and `upgrade`.

Requires https://github.com/dart-lang/ai/pull/75 to land first.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
@@ -28,6 +28,7 @@ base mixin DartCliSupport on ToolsSupport, LoggingSupport
     }
     registerTool(dartFixTool, _runDartFixTool);
     registerTool(dartFormatTool, _runDartFormatTool);
+    registerTool(dartPubTool, _runDartPubTool);
     return super.initialize(request);
   }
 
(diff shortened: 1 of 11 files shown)


pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
```diff
@@ -113,22 +165,62 @@ base mixin DartCliSupport on ToolsSupport, LoggingSupport {
...
+          TextContent(
+            text: '$commandDescription in ${projectRoot.path}:\n$output',
+          ),
         );
       }
     }
     return CallToolResult(content: outputs);
   }
 
+  static final dartPubTool = Tool(
```
jakemac53, line 186: I think we should try and make this a general "pub" tool, and it should detect whether it should use `dart pub` or `flutter pub` in each of the given roots based on project type.

kenzieschmoll: Do `dart pub` and `flutter pub` actually behave differently? Or does `flutter pub` just call into `dart pub` from Flutter tools?

pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
```diff
@@ -182,3 +274,53 @@ base mixin DartCliSupport on ToolsSupport, LoggingSupport {
     ),
   );
 }
+
+/// The set of supported `dart pub` subcommands.
+enum SupportedPubCommand {
+  // This is supported in a simplified form: `dart pub add <package-name>`.
+  add(requiresPackageName: true),
```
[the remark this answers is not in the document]
kenzieschmoll: Added a TODO to add this support: https://github.com/dart-lang/ai/issues/77

pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
jakemac53: They do behave differently, `dart pub` does not work in flutter projects and `flutter pub` does not work in dart projects.

kenzieschmoll: I also just checked on a flutter project and `dart pub` works there as well.

jakemac53 (commented):
LGTM - see comment regarding dart vs flutter "pub" commands.

I am fine with addressing this as a follow-up, but removed auto-submit in case you do want to address it now.

pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
jakemac53:
```
Because counter_app depends on flutter_test from sdk which doesn't exist (the
  Flutter SDK is not available), version solving failed.

Flutter users should use `flutter pub` instead of `dart pub`.
```

jakemac53: (it doesn't work if you have any flutter SDK deps)

kenzieschmoll:
> see comment regarding dart vs flutter "pub" commands.

Responded in the comment above. I'll land as is until we can confirm there is an issue with just always using `dart pub`.

jakemac53:
> Responded in the comment above. I'll land as is until we can confirm there is an issue with just always using `dart pub`.

See my response, it only works if you don't have flutter deps. Or maybe it works if your `dart` is from the flutter SDK? In my case `dart` is just the Dart SDK version.

pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
kenzieschmoll: I used `dart pub add` on a flutter project and it succeeded, but my `dart` is the Dart from the Flutter SDK so maybe that is the difference. I filed https://github.com/dart-lang/ai/issues/81 to track work and discuss the best way to determine whether a project is a dart or flutter project.

PR #79, opened 2025-04-22
implement a tool to resolve workspace symbols based on a query

jakemac53:
This allows you to ask things like "Tell me about <class-name> in my project" and it will tell you where it is located.

Follow on features will allow for looking up more detailed information about the symbol such as its doc comments and API.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
@@ -57,6 +57,7 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
     unsupportedReason ??= await _initializeAnalyzerLspServer();
     if (unsupportedReason == null) {
       registerTool(analyzeFilesTool, _analyzeFiles);
+      registerTool(resolveWorkspaceSymbolTool, _resolveWorkspaceSymbol);
     }
 
     // Don't call any methods on the client until we are fully initialized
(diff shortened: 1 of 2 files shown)


pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -268,7 +331,19 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
...
     name: 'analyze_files',
     description: 'Analyzes the entire project for errors.',
-    inputSchema: ObjectSchema(),
+    inputSchema: Schema.object(),
+  );
+
+  @visibleForTesting
+  static final resolveWorkspaceSymbolTool = Tool(
+    name: 'resolve_workspace_symbol',
+    description: 'Resolves a given symbol or symbols in a workspace.',
```
kenzieschmoll, line 340: does this need to be more descriptive for an LLM to know what this does? This reads like only someone familiar with static analysis will understand what this tool does.

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -268,7 +331,19 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
   static final analyzeFilesTool = Tool(
     name: 'analyze_files',
     description: 'Analyzes the entire project for errors.',
-    inputSchema: ObjectSchema(),
+    inputSchema: Schema.object(),
+  );
+
+  @visibleForTesting
+  static final resolveWorkspaceSymbolTool = Tool(
+    name: 'resolve_workspace_symbol',
```
kenzieschmoll, line 339: same question WRT the tool name. Should this be understandable by a human trying to decide whether to approve or deny the use of this tool?

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: So, I was wondering about that but I have tried a few different prompts and it seems to understand. I think the LLM was trained on enough people familiar with static analysis that it actually "understands" this prompt well?

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -268,7 +331,19 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
...
+    description: 'Resolves a given symbol or symbols in a workspace.',
+    inputSchema: Schema.object(
+      properties: {
+        'query': Schema.string(title: 'The query for the symbol(s) to look up'),
```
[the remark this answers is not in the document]
jakemac53: The LSP spec doesn't have any sort of definition I could find, cc @bwilkerson who might know what is actually supported?

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: With the newest API version there is an additional field called "title" so I could add a user friendly title under there?

We should probably do this to all the tools

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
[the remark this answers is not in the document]
jakemac53: As far as I can tell, this just does a "fuzzy match" by exact name. So it allows for typos but not any sort of patterns or path queries.

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: I updated this a bit, PTAL

pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: Added tool annotations to give it a better name, PTAL

kenzieschmoll (approved):

Issue #83, opened 2025-04-23
Add tool annotations to all tools


jakemac53:
This can involve a user visible title, and hints as to whether the tool is read only, destructive, open world, or idempotent.

PR #84, opened 2025-04-23
Add tool annotations to tools

jakemac53:
Closes https://github.com/dart-lang/ai/issues/83

Also moves us to a consistent style for schemas, using the `Schema.<kind>` style.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/analyzer.dart
@@ -268,7 +268,8 @@ base mixin DartAnalyzerSupport on ToolsSupport, LoggingSupport {
   static final analyzeFilesTool = Tool(
     name: 'analyze_files',
     description: 'Analyzes the entire project for errors.',
-    inputSchema: ObjectSchema(),
+    inputSchema: Schema.object(),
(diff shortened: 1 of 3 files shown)


kenzieschmoll (commented):
This looks like it might be behind the latest on the main branch since it doesn't include changed to the pub tools. Although, may be better to approve and land https://github.com/dart-lang/ai/pull/82 first since the location of those tools are moving to a different file.

jakemac53:
Updated now that #82 is merged, PTAL.

pkgs/dart_tooling_mcp_server/lib/src/mixins/pub.dart
```diff
@@ -77,26 +77,27 @@ base mixin PubSupport on ToolsSupport, LoggingSupport
     description:
         'Runs a dart pub command for the given project roots, like `dart pub '
         'get` or `dart pub add`.',
-    inputSchema: ObjectSchema(
+    annotations: ToolAnnotations(title: 'pub', readOnlyHint: false),
```
jakemac53, line 80: Note that we might want consider splitting up some of these tools into the (mostly) "read only" ones versus the ones that are not.

At least `pub get` should be considered read only probably, even though its technically not it only touches hidden files (other than the pubspec.lock, but it won't alter that if it exists already).

kenzieschmoll (approved):

PR #85, opened 2025-04-23
Add a RootsTrackingSupport mixin for servers, use it from DartAnalyzerSupport

jakemac53:
This mixin will be used by anything that wants to look at what the current roots are, so we can share that logic across them.

Likely we will want to add this to various other tools (like the command line tools).

### DIFF
--- a/pkgs/dart_mcp/lib/src/server/roots_tracking_support.dart
+++ b/pkgs/dart_mcp/lib/src/server/roots_tracking_support.dart
@@ -1,0 +1,108 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of 'server.dart';
+
+/// Mix this in to any MCPServer to add support tracking the [Root]s as given
+/// by the client in an opinionated way.
+///
+/// Listens to change events and updates the set of [roots].
+base mixin RootsTrackingSupport on LoggingSupport {
+  /// All known workspace [Root]s from the last call to [listRoots].
+  ///
+  /// May be a [Future] if we are currently requesting the roots.
+  FutureOr<List<Root>> get roots => switch (_rootsState) {
+    RootsState.upToDate => _roots!,
+    RootsState.pending => _rootsCompleter!.future,
+  };
+
+  /// The current state of [roots], whether it is up to date or waiting on
+  /// updated values.
+  RootsState _rootsState = RootsState.pending;
+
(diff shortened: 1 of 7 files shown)


pkgs/dart_mcp/lib/src/server/roots_tracking_support.dart
```diff
@@ -0,0 +1,108 @@
...
+    initialized.then((_) async {
+      if (!supportsRoots) {
+        log(
+          LoggingLevel.warning,
+          'Client does not support the roots capability, some functionality '
+          'may be disabled.',
+        );
+      } else {
+        if (supportsRootsChanged) {
+          rootsListChanged!.listen((event) {
```
kenzieschmoll, line 57: double checking: does this listener need to be cancelled anywhere?

jakemac53: The controller for this stream gets cancelled when the client disconnects so no it shouldn't be necessary

pkgs/dart_mcp/test/roots_tracking_support_test.dart
```diff
@@ -0,0 +1,76 @@
...
+    expect(client.addRoot(c), isTrue);
+    await pumpEventQueue();
+    expect(
+      server.roots,
+      isA<Future>(),
+      reason: 'Server is waiting to fetch new roots',
+    );
+    expect(
+      server.roots,
+      completion(unorderedEquals([b, c, d])),
```
kenzieschmoll, line 48: what happened to root a?

jakemac53: It gets removed later on line 53 - this is an async expectation for a future that doesn't complete until the completer gets completed on line 35.

pkgs/dart_tooling_mcp_server/lib/src/mixins/dart_cli.dart
```diff
@@ -21,14 +21,21 @@ base mixin DartCliSupport on ToolsSupport, LoggingSupport
     implements ProcessManagerSupport {
   @override
   FutureOr<InitializeResult> initialize(InitializeRequest request) {
+    String? unsupportedReason;
     if (request.capabilities.roots == null) {
```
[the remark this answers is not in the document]
jakemac53: Sure for now I can just remove this instead of making it not throw.

Longer term we might want to still enforce that we will only run these tools inside one of the specified roots, just to keep the AI on the rails. As an example, what if it somehow gets into the pub cache and starts trying to edit files 😱 .

kenzieschmoll (approved):

PR #92, opened 2025-04-29
Add throttling for resource notifications

jakemac53:
This is in anticipation of potentially modeling runtime errors as resources, so they might be updated often. But it is in general a good practice to throttle these.

### DIFF
--- a/pkgs/dart_mcp/lib/src/server/resources_support.dart
+++ b/pkgs/dart_mcp/lib/src/server/resources_support.dart
@@ -45,7 +45,7 @@ base mixin ResourcesSupport on MCPServer {
   final StreamController<void> _resourceListChangedController =
       StreamController<void>();
 
-  /// At most updates for the same resource or list of resources will only be
+  /// At most, updates for the same resource or list of resources will only be
   /// sent once per this [Duration].
   ///
   /// Override this getter in subtypes in order to configure the delay.
--- a/pkgs/dart_mcp/lib/src/server/resources_support.dart
+++ b/pkgs/dart_mcp/lib/src/server/resources_support.dart
@@ -38,7 +38,17 @@ base mixin ResourcesSupport on MCPServer {
   _resourceTemplates = [];
 
   /// The list of currently subscribed resources by URI.
-  final Set<String> _subscribedResources = {};
+  final Map<String, StreamController<ResourceUpdatedNotification>>
+  _subscribedResources = {};
+
+  final StreamController<void> _resourceListChangedController =
+      StreamController<void>();
+
+  /// At most updates for the same resource or list of resources will only be
+  /// sent once per this [Duration].
+  ///
+  /// Override this getter in subtypes in order to configure the delay.
+  Duration get resourceUpdateThrottleDelay => const Duration(milliseconds: 500);
 
   /// Invoked by the client as a part of initialization.
   ///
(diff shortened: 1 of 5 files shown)


pkgs/dart_mcp/test/resources_support_test.dart
```diff
@@ -136,6 +136,81 @@ void main() {
...
+    final serverConnection = environment.serverConnection;
+    final server = environment.server;
+
+    final resourceListChangedQueue = StreamQueue(
+      serverConnection.resourceListChanged,
+    );
+
+    final resources = [
+      for (var i = 0; i < 5; i++) Resource(name: '$i', uri: 'foo://$i'),
+    ];
```
kenzieschmoll, line 155: nit: using List.generate might be more terse

pkgs/dart_mcp/lib/src/server/resources_support.dart
```diff
@@ -37,8 +37,19 @@ base mixin ResourcesSupport on MCPServer {
   final List<({ResourceTemplate template, ReadResourceHandler handler})>
   _resourceTemplates = [];
 
-  /// The list of currently subscribed resources by URI.
-  final Set<String> _subscribedResources = {};
+  /// The currently subscribed resource [StreamController]s by URI.
```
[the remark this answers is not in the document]
jakemac53: These can be any URI, as defined by the server.

pkgs/dart_mcp/lib/src/server/resources_support.dart
```diff
@@ -37,8 +37,19 @@ base mixin ResourcesSupport on MCPServer {
...
+  final Map<String, StreamController<ResourceUpdatedNotification>>
+  _subscribedResources = {};
+
+  /// The [StreamController] which controls [ResourceListChangedNotification]s.
+  final StreamController<void> _resourceListChangedController =
+      StreamController<void>();
+
+  /// At most updates for the same resource or list of resources will only be
```
jakemac53, line 48:
```suggestion
  /// At most, updates for the same resource or list of resources will only be
```

pkgs/dart_mcp/test/resources_support_test.dart
jakemac53, line 155:
```suggestion
    final resources = List.generate(
      5,
      (i) => Resource(name: '$i', uri: 'foo://$i'),
    );
```

pkgs/dart_mcp/lib/src/server/resources_support.dart
```diff
@@ -65,9 +76,26 @@ base mixin ResourcesSupport on MCPServer {
...
     return result;
   }
 
+  @override
+  Future<void> shutdown() async {
+    await super.shutdown();
+    await _resourceListChangedController.close();
+    final subscribed = _subscribedResources.values.toList();
+    _subscribedResources.clear();
+    await Future.wait([for (var controller in subscribed) controller.close()]);
```
natebosch, line 96: [optional] It's little shorter to use use the `Iterable<Future>.wait` extension.

```suggestion
    await subscribed.map((c) => c.close()).wait;
```

pkgs/dart_mcp/test/resources_support_test.dart
jakemac53: Because of the new formatter its actually longer 🤣 . Not sure which is easier to read, thoughts?

jakemac53: see my comment below, not sure which I like more

natebosch: Readability guidelines generally prefer using collection literals, I think this looks good as is.

pkgs/dart_mcp/test/resources_support_test.dart
```diff
@@ -136,6 +136,81 @@ void main() {
...
+    // Should get exactly two notifications even though we have more resources,
+    // one initial notification and one after the throttle delay.
+    await resourceListChangedQueue.take(2);
+    expect(
+      resourceListChangedQueue.hasNext.timeout(
+        const Duration(milliseconds: 10),
+        onTimeout: () => false,
+      ),
+      completion(false),
+    );
```
natebosch, line 176: I don't think it adds much safety to check for no-events within a bound window of time. We're limited by overall test runtime, but we can make a stronger expectation that no further events are emitted here. How about

```suggestion
    expect(
      resourceListChangedQueue.hasNext,
      doesNotComplete,
    );
```

natebosch (approved):

kenzieschmoll (approved):

pkgs/dart_mcp/test/resources_support_test.dart
jakemac53: Updated this to `completion(false)` with a zero delay for the throttling and `pumpEventQueue` to ensure delivery of events.

PR #94, opened 2025-04-29
Add runtime errors resource and tool to clear errors.

jakemac53:
This ended up being a fair bit more complicated than I wanted, as it also refactors things and moves us towards a direction where we can support multiple connected apps (which was a natural extension of creating the resource URI, to include the debug session ID in it).

Functionally, this:

- Starts collecting errors as soon as a debug session is connected, and creates a runtime errors resource for each session.
- Allows us to fire change notifications for when new errors come in.
- Also added a `clear_runtime_errors` tool as well as an argument to the `hotReload` tool to clear the errors prior to the hot reload.
- Drops the support for `since` in favor of just clearing the errors.

Eventually we might want to support pagination of the resource instead (it is already modeled as many parts), I filed a feature request on the MCP repo for this, but it likely won't land for some time if at all.


jakemac53:
cc @gspencergoog

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -83,24 +86,46 @@ base mixin DartToolingDaemonSupport on LoggingSupport, ToolsSupport {
...
+        );
+      }
     }
   }
 
   @override
   FutureOr<InitializeResult> initialize(InitializeRequest request) async {
     registerTool(connectTool, _connect);
     registerTool(getRuntimeErrorsTool, runtimeErrors);
+    registerTool(clearRuntimeErrorsTool, clearRuntimeErrors);
```
gspencergoog, line 128: Instead of making this a separate tool, could we just keep it as an argument to the `hotReload` tool or the `getRuntimeErrors` tool (to clear *after* returning the errors)?  Each call to a tool requires a separate LLM inference, so having something that doesn't produce output, and only affects future results from other calls seems like a waste of an inference.

gspencergoog (approved):

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -31,7 +32,7 @@ base mixin DartToolingDaemonSupport on LoggingSupport, ToolsSupport {
   /// [VmService] objects are automatically removed from the Map when the
   /// [VmService] shuts down.
   @visibleForTesting
-  final activeVmServices = <String, VmService>{};
+  final activeVmServices = <String, Future<VmService>>{};
```
kenzieschmoll, line 35: the dart doc needs to be updated since the key for this map changed (debug session id vs vm service uri).

Generally, why the change from vm service uri as the key to debug session id?

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -186,6 +210,21 @@ base mixin DartToolingDaemonSupport on LoggingSupport, ToolsSupport {
...
+    dtd.onEvent('Editor').listen((e) async {
+      log(LoggingLevel.debug, e.toString());
+      switch (e.kind) {
+        case 'debugSessionStarted':
+        case 'debugSessionChanged':
+          await updateActiveVmServices();
+        case 'debugSessionStopped':
+          await activeVmServices
+              .remove((e.data['debugSession'] as DebugSession).id)
+              ?.then((service) => service.dispose());
```
kenzieschmoll, line 223: we are already doing this above when the vm service shuts down, do we need to do it here too? will calling dispose twice be a problem? Also, we'd want to avoid calling this here if if doesn't allow us to remove the resource in the onDone handler above

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -541,7 +550,17 @@ base mixin DartToolingDaemonSupport on LoggingSupport, ToolsSupport {
         'This is to apply the latest code changes to the running application. '
         'Requires "${connectTool.name}" to be successfully called first.',
     annotations: ToolAnnotations(title: 'Hot reload', destructiveHint: true),
-    inputSchema: Schema.object(),
+    inputSchema: Schema.object(
+      properties: {
+        'clearRuntimeErrors': Schema.bool(
```
kenzieschmoll, line 555: can we use a shared const for this string since it is used in multiple tools?

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -606,6 +625,100 @@ base mixin DartToolingDaemonSupport on LoggingSupport, ToolsSupport {
       ),
     ],
   );
+
+  static final runtimeErrorsScheme = 'runtime-errors';
+}
+
+/// Listens on a VM service for errors.
+class _AppErrorsListener {
```
kenzieschmoll, line 633: nit: rename to _FlutterAppErrorsListener to make it clear this is only for flutter apps and not pure dart apps

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -678,28 +791,19 @@ extension type DebugSession.fromJson(Map<String, Object?> _value)
...
     required String projectRootPath,
-    required String vmServiceUri,
+    required String? vmServiceUri,
   }) => DebugSession.fromJson({
     'debuggerType': debuggerType,
     'id': id,
     'name': name,
     'projectRootPath': projectRootPath,
-    'vmServiceUri': vmServiceUri,
+    if (vmServiceUri != null) 'vmServiceUri': vmServiceUri,
```
kenzieschmoll, line 807: what type of debug sessions do not have a vm service uri?

pkgs/dart_tooling_mcp_server/test/tools/dtd_test.dart
```diff
@@ -279,6 +221,173 @@ void main() {
...
+            (runtimeErrorsResult.content[1] as TextContent).text,
+            contains('A RenderFlex overflowed by'),
+          );
+
+          final nextResult = await testHarness.callToolWithRetry(
+            CallToolRequest(name: runtimeErrorsTool.name),
+          );
+          expect(
+            (nextResult.content.first as TextContent).text,
+            contains('No runtime errors found'),
```
kenzieschmoll, line 275: I'm surprised there are no errors even though the clearRuntimeErrors arg was not passed, but maybe I'm not following the logic here.

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
jakemac53: When we get the event for a debug session ending, a VM service URI is not included, only the debug session ID

pkgs/dart_tooling_mcp_server/test/tools/dtd_test.dart
jakemac53: The logic is a bit hard to follow - we are passing `clearRuntimeErrors` in the retry loop above (which is waiting for the error to show up, this test was flaking without that since there is no explicit 1 second delay any more built into this tool).

This then asks for the errors again - to check that the errors were in fact cleared.

jakemac53: Added a comment above this tool call

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
jakemac53: It is safe to call `dispose` multiple times.

This also doesn't affect our ability to dispose of the resource (but will trigger it by triggering the `onDone` future).

Really, this is just defensive, in case somehow the VM service remained active but the debug session was completed. We will shut it down if either DTD tells us to, or the VM service connection closes.

jakemac53: I did remove the (should be) unnecessary `dispose` call form the `onDone` callback though, that only gets completed when the VM service is disposed anyways.

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
jakemac53: Yeah, I will just add it to the bag of constants I created in the other PR, once that lands

pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
jakemac53: When we get the `started` notification, they don' that's a VM service URI yet. Then we get a follow-up `changed` notification that does contain the VM service URI.

kenzieschmoll (approved):

PR #101, opened 2025-05-01
Require roots for all CLI tools

jakemac53:
Updates all CLI tools to only allow them to be ran under one of the client specified roots.

- If paths are specified, those also must not escape the roots.
- Make roots not required any longer, use the client specified roots by default for all tools.
- Create some shared constants for argument names and a shared function to create the schemas for the roots parameters.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/utils/cli_utils.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/utils/cli_utils.dart
@@ -5,7 +5,10 @@
 import 'dart:io';
 
 import 'package:dart_mcp/server.dart';
+import 'package:path/path.dart' as p;
 import 'package:process/process.dart';
+
+import 'constants.dart';
 
 /// Runs [command] in each of the project roots specified in the [request].
 ///
@@ -15,6 +18,10 @@ import 'package:process/process.dart';
 /// The [commandDescription] is used in the output to describe the command
 /// being run. For example, if the command is `['dart', 'fix', '--apply']`, the
 /// command description might be `dart fix`.
+///
+/// The [knownRoots] are used by default if no roots are provided as an
+/// argument on the [request]. Otherwise, all roots provided in the request
+/// arguments must still be encapsulated by the [knownRoots].
 ///
 /// [defaultPaths] may be specified if one or more path arguments are required
 /// for the command (e.g. `dart format <default paths>`). The paths can be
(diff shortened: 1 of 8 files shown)


jakemac53:
cc @domesticmouse I know you said you were giving a demo soon, this could affect that so I will wait to land this until your demo is done, please let me know when its safe :)

pkgs/dart_tooling_mcp_server/lib/src/utils/cli_utils.dart
```diff
@@ -102,3 +144,30 @@ Future<CallToolResult> runCommandInRoots(
   }
   return CallToolResult(content: outputs);
 }
+
+/// Returns whether or not [rootUri] is an allowed root, either exactly matching
+/// or under on of the [knownRoots].
+bool _isAllowedRoot(String rootUri, List<Root> knownRoots) =>
+    knownRoots.any((knownRoot) => knownRoot.uri.startsWith(rootUri));
```
kenzieschmoll, line 151: shouldn't this check that `rootUri.startsWith(knownRoot.uri)` instead?

pkgs/dart_tooling_mcp_server/lib/src/utils/constants.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// A namespace for all the parameter names.
+extension ParameterNames on Never {
```
kenzieschmoll, line 6: why extend Never rather than just creating a new abstract class?

kenzieschmoll (approved):
a couple comments but overall lgtm

pkgs/dart_tooling_mcp_server/lib/src/utils/constants.dart
jakemac53: Extensions are a bit more restricted (can't be implemented/extended etc), and don't create a real type (they disappear at runtime entirely) so they are technically cheaper.

jakemac53: Note also that static methods on extensions are not actually added to `Never` either, so this has no impact on the `Never` type anywhere it appears.

pkgs/dart_tooling_mcp_server/lib/src/utils/cli_utils.dart
jakemac53: I rewrote this to also make sure you can't circumvent it using `../` and added tests

Issue #4, opened 2025-03-24
Add logging option to MCP client and server


jakemac53:
It would be nice to be able to get a full transcript of all messages sent.

We have to be careful (especially on the server side) not to just call `print` or write to stdout, since that will break the protocol, so having a logging framework makes sense.

On the client side, we should distinguish the logs between each server connection - maybe with separate files per server connection.

DanTup:
@jakemac53 is there (or could there) be a way to log to a file by providing a path on the command line?

For VS Code, we don't control the process, but it'd be nice to have [a setting like the others](https://dartcode.org/docs/logging/#individual-logs) that allow logging to disk to aid debugging.

jakemac53:
Yeah we definitely could do that

DanTup:
lmk if I should file an issue for it (it's not urgent, but it'd be nice to help debug/test things).

DanTup:
Filed https://github.com/dart-lang/ai/issues/171, thanks!

PR #102, opened 2025-05-01
add option to log protocol messages to a Sink<String>

jakemac53:
Closes https://github.com/dart-lang/ai/issues/4.

I also used this functionality to test the invalid protocol message flow, which does appear to work fine as is (json_rpc_2 handles it well for us), but is now tested.

### DIFF
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -9,7 +9,6 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:async/async.dart' hide Result;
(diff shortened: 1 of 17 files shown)


pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -52,10 +51,16 @@ base class MCPClient {
 
   /// Connect to a new MCP server by invoking [command] with [arguments] and
   /// talking to that process over stdin/stdout.
+  ///
+  ///
```
kenzieschmoll, line 55: remove extra blank line

kenzieschmoll (approved):

Issue #96, opened 2025-04-30
Invalid MCP Server instantiation fails poorly


domesticmouse:
I was attempting to run a MCP server incorrectly. It exited with an error message. This error message was not surfaced by the MCP Client sample code.

```console
$ dart run bin/mcp_client.dart
connecting to server
server started
initializing server
Unhandled exception:
Bad state: The client closed with pending request "initialize".
```

The simplified code is as follows. For context, the `mcp_server` requires a command line argument that is not supplied here.

```dart
import 'package:dart_mcp/client.dart';

void main() async {
  final client = MCPClient(
    ClientImplementation(name: 'example dart client', version: '0.1.0'),
  );
  print('connecting to server');
  final server = await client.connectStdioServer('dart', [
    'run',
    '../mcp_server/bin/mcp_server.dart',
  ]);
  print('server started');

  print('initializing server');
  final initializeResult = await server.initialize(
    InitializeRequest(
      protocolVersion: protocolVersion,
      capabilities: client.capabilities,
      clientInfo: client.implementation,
    ),
  );
  print('initialized: $initializeResult');
  if (initializeResult.protocolVersion != protocolVersion) {
    throw StateError(
      'Protocol version mismatch, expected $protocolVersion, '
      'got ${initializeResult.protocolVersion}',
    );
  }

  await client.shutdown();
}
```

The `pubspec.yaml` looks like the following:

```yaml
name: mcp_client
description: A sample command-line application.
version: 1.0.0

environment:
  sdk: ^3.7.2

dependencies:
  dart_mcp: ^0.1.0

dev_dependencies:
  lints: ^5.0.0
  test: ^1.24.0
```

Can we please add an exception or similar when the MCP server dies on initialization?

jakemac53:
https://github.com/dart-lang/ai/issues/4 is likely related to this as well

jakemac53:
In this case, the first message from the server is an invalid one (not JSON), I am a bit surprised we aren't seeing some other sort of error bubble up here.

jakemac53:
Ok, so json_rpc_2 automatically handles invalid messages by sending a notification message back to the server that it had sent an invalid request.

So, the client never sees this message and it sends an initialize request. The server then exits, and then we get the error about the connection closing with the pending initialize request.

We don't really have a good opportunity to see the actual bad output at all because it's intercepted by the json_rpc_2 package, we may want to add something to that package to allow you to see these.

[...]

PR #105, opened 2025-05-02
add test for server closing early to validate behavior

jakemac53:
Related to https://github.com/dart-lang/ai/issues/96

- Tests the current behavior, ensures we aren't leaking exceptions beyond what is expected.
- Also documents explicitly the behavior in initialize to encourage exception handling, and helping you to debug.

### DIFF
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -264,6 +264,11 @@ base class ServerConnection extends MCPBase {
   ///
   /// The client must call [notifyInitialized] after receiving and accepting
   /// this response.
+  ///
+  /// Throws a [StateError] if initialization fails for unknown reasons (usually
+  /// the server connection closes prematurely due to misconfiguration). To
+  /// debug these errors you should pass a `protocolLogSink` when creating these
+  /// connection
   Future<InitializeResult> initialize(InitializeRequest request) async {
     final response = await sendRequest<InitializeResult>(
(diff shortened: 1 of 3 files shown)


pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -264,6 +264,11 @@ base class ServerConnection extends MCPBase {
   ///
   /// The client must call [notifyInitialized] after receiving and accepting
   /// this response.
+  ///
+  /// Throws a [StateError] if initialization fails for unknown reasons (usually
+  /// the server connection closes prematurely due to misconfiguration). To
+  /// debug these errors you should pass a `protocolLogSink` when creating these
+  /// connection
```
kenzieschmoll, line 271: r/connection/connections
or r/these/this. Whichever matches the intent of the comment.

pkgs/dart_mcp/test/client_and_server_test.dart
```diff
@@ -287,6 +288,45 @@ void main() {
...
+        throwsA(
+          isA<StateError>().having(
+            (e) => e.message,
+            'message',
+            'The client closed with pending request "initialize".',
+          ),
+        ),
+      );
+
+      serverChannel.sink.close();
```
kenzieschmoll, line 323: does this need to be before the initialize request? Maybe add a comment to the line of code that is closing the client

kenzieschmoll (approved):
lgtm with a couple comments

PR #110, opened 2025-05-05
add signature_help tool

jakemac53:
This forwards directly to the [LSP signature help](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_signatureHelp) functionality.

Allows requesting information about the signature of some API being _used_ at a specific cursor location.

**Note**: Does not seem to give useful information if the location is a _definition_ site (such as one returned by a workspace_symbol lookup). But this looks like it is by design probably.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/README.md
+++ b/pkgs/dart_tooling_mcp_server/README.md
@@ -9,6 +9,7 @@ WIP. This package is still experimental and is likely to evolve quickly.
 | Tool Name | Feature Group | Description |
 | --- | --- | --- |
 | `analyze_files` | `static analysis` | Analyzes the entire project for errors. |
+| `signature_help` | `static_analysis` | Gets signature information for usage at a given cursor position. |
 | `resolve_workspace_symbol` | `static analysis` | Look up a symbol or symbols in all workspaces by name. |
 | `dart_fix` | `static tool` | Runs `dart fix --apply` for the given project roots. |
 | `dart_format` | `static tool` | Runs `dart format .` for the given project roots. |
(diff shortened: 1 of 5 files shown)


bwilkerson (approved):

bwilkerson:
> Does not seem to give useful information if the location is a definition site ...

I think the primary use that VS Code has for this is to display information about the parameters while the user is attempting to edit a call site. I don't think it helps much at the declaration site. But it would be good to double check with Danny to make sure my understanding is correct.

jakemac53:
> I think the primary use that VS Code has for this is to display information about the parameters while the user is attempting to edit a call site. I don't think it helps much at the declaration site. But it would be good to double check with Danny to make sure my understanding is correct.

Yes, this is my understanding as well, and I tried to describe that in the text for the tool. I just wanted to call it out in the review, because we are also going to want that functionality (give me the docs for declaration X).

bwilkerson:
The docs are included in the hover text (even at the declaration site). Not sure if the hover text is in a good format for LLMs, though. Hovers also include the docs and signature information at call sites, so it might be a better all-purpose tool in some ways.

Also, all of the docs we currently return are minimally pre-processed. We remove dartdoc tags that we don't understand how to render, so if that data is useful to the LLMs then we might need a dedicated protocol just for that purpose.

jakemac53:
> The docs are included in the hover text (even at the declaration site). Not sure if the hover text is in a good format for LLMs, though. Hovers also include the docs and signature information at call sites, so it might be a better all-purpose tool in some ways.

Ah interesting, maybe that would be better. Most of the time though I think this will be at call sites I validated that the existing response data is sufficient for an LLM to use, given enough prompting.

PR #112, opened 2025-05-06
add basic file system server example

jakemac53:
This is useful for testing the workflow client with more complicated flows to give it file system access, without needing to depend on a third party server for that.

### DIFF
--- a/pkgs/dart_mcp/example/file_system_server.dart
+++ b/pkgs/dart_mcp/example/file_system_server.dart
@@ -44,6 +44,7 @@ final class SimpleFileSystemServer extends MCPServer
     registerTool(readFileTool, _readFile);
     registerTool(writeFileTool, _writeFile);
     registerTool(deleteFileTool, _deleteFile);
+    registerTool(listFilesTool, _listFiles);
     return super.initialize(request);
   }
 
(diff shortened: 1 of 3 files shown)


pkgs/dart_mcp/example/file_system_server.dart
```diff
@@ -0,0 +1,150 @@
...
+    return super.initialize(request);
+  }
+
+  /// Checks if [path] is under any of the known [roots], and if not returns
+  /// an error result.
+  ///
+  /// Returns `null` if [path] is valid.
+  Future<CallToolResult?> _checkAllowedPath(String path) async {
+    for (var root in await roots) {
+      if (root.uri == path || p.isWithin(root.uri, path)) {
```
domesticmouse, line 56: I'm guessing `path.isWithin` deals with canonicalizing the incoming path to prevent `../..` style escape attacks?

pkgs/dart_mcp/example/file_system_server.dart
```diff
@@ -0,0 +1,150 @@
...
+          name: 'file system',
+          version: '0.0.1',
+        ),
+      );
+
+  @override
+  FutureOr<InitializeResult> initialize(InitializeRequest request) {
+    registerTool(readFileTool, _readFile);
+    registerTool(writeFileTool, _writeFile);
+    registerTool(deleteFileTool, _deleteFile);
```
domesticmouse, line 46: Can we please add a tool for `ls`ing a directory within the allowed path?

domesticmouse (approved):

pkgs/dart_mcp/example/file_system_server.dart
jakemac53: Yes this is most of what it does :)

pkgs/dart_mcp/example/file_system_server.dart
jakemac53: done

kenzieschmoll (approved):

PR #113, opened 2025-05-06
add a test command and project detection

jakemac53:
- Rename `DartCliSupport` to `DashCliSupport`
- Add support for choosing the command to run based on the root it is ran in
  - Defaults to project detection, calling `flutter` or `dart` as appropriate.
    - Updated the dart only commands to explicitly use `dart` always
- Adds a tool for running tests, which requires this support.

TODO: Add support for some `test` command arguments such as `--name`, and possibly platform or other arguments. We may just want to allow general arguments.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/pubspec.yaml
+++ b/pkgs/dart_tooling_mcp_server/pubspec.yaml
@@ -18,6 +18,7 @@ dependencies:
   dds_service_extensions: ^2.0.1
   devtools_shared: ^11.2.0
   dtd: ^2.4.0
+  file: ^7.0.1
   json_rpc_2: ^3.0.3
   # TODO: Get this another way.
   language_server_protocol:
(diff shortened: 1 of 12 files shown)


pkgs/dart_tooling_mcp_server/test_fixtures/dart_cli_app/pubspec.yaml
```diff
@@ -0,0 +1,4 @@
+name: exampl_dart_cli_app
```
kenzieschmoll, line 1: sp: example

pkgs/dart_tooling_mcp_server/pubspec.yaml
```diff
@@ -28,6 +29,7 @@ dependencies:
   meta: ^1.16.0
   path: ^1.9.1
   process: ^5.0.3
+  pubspec_parse: ^1.5.0
```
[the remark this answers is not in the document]
jakemac53: I don't see usage but it is there, because its part of the dart-lang/tools mono repo.

pkgs/dart_tooling_mcp_server/test_fixtures/dart_cli_app/pubspec.yaml
jakemac53, line 1:
```suggestion
name: example_dart_cli_app
```

PR #116, opened 2025-05-07
Add a tool to get the active cursor location

jakemac53:
Similar to the selected widget tool this should help the AI get some context as to what the user is talking about.

Just directly forwards the last event received from DTD.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
@@ -171,7 +171,7 @@ base mixin DartToolingDaemonSupport
       );
       unawaited(_dtd!.done.then((_) async => await _resetDtd()));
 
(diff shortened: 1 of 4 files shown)


pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -220,8 +225,10 @@ base mixin DartToolingDaemonSupport
           await updateActiveVmServices();
         case 'debugSessionStopped':
           await activeVmServices
-              .remove((e.data['debugSession'] as DebugSession).id)
+              .remove(e.data['debugSessionId'] as String)
```
jakemac53, line 228: This is an unrelated bug fix that I snuck in here, I can separate it out if we want

kenzieschmoll (approved):
add to list of tools in the readme

PR #120, opened 2025-05-09
Fix notification bug where null parameters caused type error

gspencergoog:
This fixes a bug in notification handling where if the parameters were left off, as in:

```json
{"jsonrpc":"2.0","method":"notifications/initialized"}
```

instead of:

```json
{"jsonrpc":"2.0","method":"notifications/initialized", "params": {}}
```

A type error would get silently ignored and the notification would never be received.


jakemac53 (approved):
thanks!

jakemac53:
retrying the jobs, there is one test that has gotten very flaky for some reason 😮‍💨

PR #126, opened 2025-05-13
Add Schema validation

gspencergoog:
## Description

This adds schema validation to the MCP Schema extension so that we can validate the returned values in tests.  It could also be used to validate results from LLMs and make them try again if they don't match the schema.

Also moved the tests into subdirs matching the lib structure.

## Tests
 - Added tests for all the different validation failure modes.


pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -288,6 +331,313 @@ extension type Schema.fromMap(Map<String, Object?> _value) {
...
+          _validateSchema(
+            entry.value,
+            data[entry.key],
+          ).map((_) => ValidationFailureType.propertyValueInvalid),
+        );
+      }
+    }
+  }
+  if (schema.patternProperties != null) {
+    for (final entry in schema.patternProperties!.entries) {
```
jakemac53, line 495: I think some comments around these sections might be good, to explain what is being checked exactly (same for the list validator).

pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -288,6 +331,313 @@ extension type Schema.fromMap(Map<String, Object?> _value) {
...
+}
+
+List<ValidationFailureType> _validateObject(
+  ObjectSchema schema,
+  Map<String, Object?> data,
+) {
+  final failures = <ValidationFailureType>[];
+
+  if (schema.minProperties != null &&
+      data.keys.length < schema.minProperties!) {
```
jakemac53, line 466: I would use if-case throughout these as well, same for the list validation below

```suggestion
  if (schema.minProperties case final min? when data.keys.length < min) {
```

pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -288,6 +331,313 @@ extension type Schema.fromMap(Map<String, Object?> _value) {
...
+// These have to be external to the Schema extension because of clashes with
+// type-named members like bool, int, and num.
+List<ValidationFailureType> _validateSchema(Schema schema, Object? data) {
+  final failures = <ValidationFailureType>[];
+
+  // 1. Handle schema combinators
+  if (schema.allOf != null) {
+    var currentAllOfValid = true;
+    final allOfDetailedFailures = <ValidationFailureType>[];
+    for (final subSchema in schema.allOf!) {
```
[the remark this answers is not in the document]
gspencergoog: I evolved the code a lot, and I'm not sure if this applies anymore.

pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -197,6 +197,78 @@ enum JsonType {
...
+
+/// A validation error with detailed information about the location of the
+/// error.
+extension type ValidationError.fromMap(Map<String, Object?> _value) {
+  factory ValidationError(
+    ValidationErrorType error, {
+    List<String>? path,
+    String? details,
+  }) => ValidationError.fromMap({
+    'error': error,
```
jakemac53, line 251: This needs to be the serialized form of the error enum (either its name or index usually), otherwise this map isn't serializable

pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -197,6 +197,78 @@ enum JsonType {
...
+    if (path != null) 'path': path,
+    if (details != null) 'details': details,
+  });
+
+  /// The type of validation error that occurred.
+  ValidationErrorType? get error => _value['error'] as ValidationErrorType?;
```
jakemac53, line 257: This getter should then do the deserialization so it does have the nice return type

pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -197,6 +197,78 @@ enum JsonType {
...
+
+  /// The path to the object that had the error.
+  List<String>? get path => _value['path'] as List<String>?;
```
jakemac53, line 260: When reading from the map, if this is in a deserialized form, it will be a `List<Object?>`, so we need to cast it.

```suggestion
  List<String>? get path => (_value['path'] as List?)?.cast<String>();
```

pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -290,7 +362,315 @@ extension type Schema.fromMap(Map<String, Object?> _value) {
   List<Schema>? get not => (_value['not'] as List?)?.cast<Schema>();
 }
 
+extension SchemaValidation on Schema {
+  /// Validates the given [data] against this schema.
+  ///
+  /// Returns a list of [ValidationError] if validation fails,
+  /// or an empty list if validation succeeds.
+  List<ValidationError> validate(Object? data) {
+    final failures = <ValidationError>[];
```
jakemac53, line 371: How about just making this be a custom HashSet?

```
    final failures = HashSet<ValidationError>(
      equals: (a, b) => ...,
      hashCode: (a) => ...,
    );
```

pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -290,7 +362,315 @@ extension type Schema.fromMap(Map<String, Object?> _value) {
...
+
+    // Calling .toSet().toList() isn't enough to de-duplicate, since even maps
+    // with identical members can be different maps, so we do it the hard, slow
+    // way. Duplicates arise because the same message can occur for multiple
+    // layers in the hierarchy.
+    final dedupedSet = <ValidationError>{};
+    for (final failure in failures.toSet()) {
+      var alreadyExists = false;
+      for (final dedupedFailure in dedupedSet) {
+        if (failure.path == dedupedFailure.path &&
```
jakemac53, line 382: These lists will only be equal if they are identical instances, we probably need a deep comparison?

pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -329,34 +709,297 @@ extension type ObjectSchema.fromMap(Map<String, Object?> _value)
...
+    // Check for property values that match any defined properties in the
+    // `properties` map. If a property name exists in the `properties` map, the
+    // value of that property is validated against the schema associated with
+    // that property.
+    final evaluatedKeys = <String>{};
+    if (properties case final props?) {
+      for (final entry in props.entries) {
+        if (data.containsKey(entry.key)) {
+          evaluatedKeys.add(entry.key);
+          final propertyPath = [...currentPath, entry.key];
```
jakemac53, line 879: We may want to consider just adding to the currentPath and then removing from it instead of copying all the way down. And then just do the copy when creating an error object.

pkgs/dart_mcp/lib/src/api/tools.dart
gspencergoog: That's a good idea: would elimnate the duplicate checks.

pkgs/dart_mcp/lib/src/api/tools.dart
gspencergoog: Done.  I now only copy it when I create a new `ValidationError`.

pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -528,5 +1441,158 @@ extension type ListSchema.fromMap(Map<String, Object?> _value)
...
+
+HashSet<ValidationError> _createHashSet() {
+  return HashSet<ValidationError>(
+    equals: (ValidationError a, ValidationError b) {
+      return const ListEquality<String>().equals(a.path, b.path) &&
+          a.details == b.details &&
+          a.error == b.error;
+    },
+    hashCode: (ValidationError error) {
+      return Object.hashAll([...error.path ?? [], error.details, error.error]);
```
jakemac53, line 1595:
```suggestion
      return Object.hashAll([...error.path ?? const [], error.details, error.error]);
```

jakemac53: I would also expect the total amount of errors here to be low. That won't always be the case though, but I am more concerned about the performance of checking for the errors, than the actual handling of the errors, and in theory these hash functions should not get called a bunch of times on any given instance.

jakemac53 (approved):

gspencergoog:
Okay, this is ready to be committed once the tests pass.  I added one more [minor change](https://github.com/dart-lang/ai/pull/126/commits/2fbe4fe0dd4b50464cf9bba97c04cf5e2b73f11d) after you approved.

PR #128, opened 2025-05-14
migrate several tests off of using the counter app and onto a test descriptor app

jakemac53:
The DTD tests are still using the counter app.

Also added capturing the working directory to the test process manager so we can confirm the working dirs are correct.

### DIFF
--- a/pkgs/dart_tooling_mcp_server/test/test_harness.dart
+++ b/pkgs/dart_tooling_mcp_server/test/test_harness.dart
@@ -483,3 +483,7 @@ class _CommandMatcher extends Matcher {
     return true;
   }
 }
+
+extension RootPath on Root {
+  String get path => Uri.parse(uri).path;
+}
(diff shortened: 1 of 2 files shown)


pkgs/dart_tooling_mcp_server/test/test_harness.dart
```diff
@@ -445,11 +445,45 @@ class TestProcessManager extends LocalProcessManager {
...
+  bool matches(Object? item, Map matchState) {
+    if (item is! ({List<Object> command, String? workingDirectory})) {
+      return false;
+    }
+    if (item.workingDirectory != value.workingDirectory) {
+      return false;
+    }
+    if (item.command.length != value.command.length) {
+      return false;
+    }
```
kenzieschmoll, line 479: nit: combine these into a single if statement using ||

kenzieschmoll (approved):

PR #130, opened 2025-05-15
migrate example dir to be a package, add README

jakemac53:
This drops some unwanted deps from the main package pubspec.

cc @devoncarew are you fine with this strategy for the SDK merge? The example package created here will NOT be a part of the SDK workspace, it would be totally separate and get its dependencies from pub.


jakemac53:
One downside I am just realizing is that we won't be able to analyze the examples on the bots....

devoncarew:
> The example package created here will NOT be a part of the SDK workspace, it would be totally separate and get its dependencies from pub.

Hmm, is this how we have the workspace set up - `examples/` are not included?

> One downside I am just realizing is that we won't be able to analyze the examples on the bots....

All of `pkg/` is analyzed on the bots, so I think this will be an issue.

It does look like https://github.com/dart-lang/ai/blob/main/pkgs/dart_mcp/example/workflow_client.dart is pretty extensive - ~500 lines long. Perhaps a simpler version of that would mean that we could have a simpler gemini client library (a json REST wrapper)?

jakemac53:
> Hmm, is this how we have the workspace set up - `examples/` are not included?

No idea if other things do this (having examples be packages is rare in general for dart projects though, but common for flutter projects).

But, it should be trivial to do (you just omit it from the workspace pubspec and don't put the workspace config in the example pubspec).

> It does look like https://github.com/dart-lang/ai/blob/main/pkgs/dart_mcp/example/workflow_client.dart is pretty extensive - ~500 lines long. Perhaps a simpler version of that would mean that we could have a simpler gemini client library (a json REST wrapper)?

The interactions with gemini are not complicated. We likely could have our own rest wrapper.

> All of `pkg/` is analyzed on the bots, so I think this will be an issue.

Yeah sounds like it. We would have to exempt it from analysis entirely which would be less than ideal.

jakemac53:
I looked into a "quick" shim around gemini and it quickly ballooned into >500 lines of code, and I didn't complete it. I don't think we want to maintain something like that.

I think maybe we should just keep this example in this repo, as a separate package.

jakemac53:
Extracted these examples all the way out to an `mcp_examples` dir, and updated readmes and such

pkgs/dart_mcp/README.md
```diff
@@ -44,10 +44,6 @@ doc comments explaining how to use it - some may require you to provide
 implementations of methods, while others may just expose new methods that you
 can call.
 
-See the [client example](example/client.dart) for some basic example code, or
```
kenzieschmoll, line 47: can we still link to the examples that will stay in the dart-lang/ai repo?

kenzieschmoll (approved):

Issue #111, opened 2025-05-05
Make roots support optional for MCP server tools


kenzieschmoll:
Many MCP clients do not support roots, so by requiring roots for things like the `analyze_files` tool, we are limiting the surfaces where the tool can be used.

Proposal: support a roots input schema for tools that require roots that defaults to use the roots set by the client when they are available.

Related to https://github.com/firebase/genkit/issues/2867

jakemac53:
We could also expose a way of setting roots on the command line or as a tool invocation.

But, in general allowing setting roots either through a tool or inputs to a tool does open up some security concerns because other servers can do prompt injection to set any root and get us to start doing things in there.

I am also slightly concerned that the LLM might start setting bad roots (like pub cache dirs it anything it sees in a stack trace).

So from that perspective actual roots are I think better because the client directly controls them instead of the LLM.

PR #132, opened 2025-05-16
Add RootsFallbackSupport mixin

jakemac53:
This mixin can be used to conditionally add `addRoots` and `removeRoots` tools, when the client doesn't support roots.

It overrides all the `RootsTrackingSupport` functionality as well as the `listRoots` method, to use the custom behavior instead.

Also adds a flag to force this fallback mode to the server, which is required for some clients who claim to support roots but do not actually (like cursor).

Closes https://github.com/dart-lang/ai/issues/111

### DIFF
--- a/pkgs/dart_tooling_mcp_server/bin/main.dart
+++ b/pkgs/dart_tooling_mcp_server/bin/main.dart
@@ -6,18 +6,17 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io' as io;
 
+import 'package:args/args.dart';
 import 'package:async/async.dart';
 import 'package:dart_mcp/server.dart';
 import 'package:dart_tooling_mcp_server/dart_tooling_mcp_server.dart';
 import 'package:stream_channel/stream_channel.dart';
 
 void main(List<String> args) async {
-  if (args.isNotEmpty) {
-    io.stderr
-      ..writeln('Expected no arguments but got ${args.length}.')
-      ..writeln()
-      ..writeln('Usage: dart_tooling_mcp_server');
-    io.exit(1);
+  final parsedArgs = argParser.parse(args);
+  if (parsedArgs.flag('help')) {
+    print(argParser.usage);
+    io.exit(0);
   }
 
   DartToolingMCPServer? server;
@@ -34,6 +33,7 @@ void main(List<String> args) async {
                 },
               ),
             ),
+        forceRootsFallback: parsedArgs.flag('force-roots-fallback'),
       );
     },
     (e, s) {
@@ -62,3 +62,16 @@ void main(List<String> args) async {
     ),
   );
 }
+
+final argParser =
+    ArgParser(allowTrailingOptions: false)
+      ..addFlag(
+        'force-roots-fallback',
+        negatable: true,
+        defaultsTo: false,
+        help:
+            'Forces a behavior for project roots which uses MCP tools instead of '
+            'the native MCP roots. This can be helpful for clients like cursor '
+            'which claim to have roots support but do not actually support it.',
+      )
+      ..addFlag('help', abbr: 'h', help: 'Show usage text');
--- a/pkgs/dart_tooling_mcp_server/bin/main.dart
+++ b/pkgs/dart_tooling_mcp_server/bin/main.dart
@@ -14,7 +14,7 @@ import 'package:stream_channel/stream_channel.dart';
 
 void main(List<String> args) async {
   final parsedArgs = argParser.parse(args);
-  if (parsedArgs.flag('help')) {
+  if (parsedArgs.flag(help)) {
     print(argParser.usage);
     io.exit(0);
   }
@@ -33,7 +33,7 @@ void main(List<String> args) async {
                 },
               ),
             ),
-        forceRootsFallback: parsedArgs.flag('force-roots-fallback'),
+        forceRootsFallback: parsedArgs.flag(forceRootsFallback),
       );
     },
     (e, s) {
@@ -66,7 +66,7 @@ void main(List<String> args) async {
 final argParser =
     ArgParser(allowTrailingOptions: false)
       ..addFlag(
-        'force-roots-fallback',
+        forceRootsFallback,
         negatable: true,
         defaultsTo: false,
         help:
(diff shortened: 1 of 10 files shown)


jakemac53:
Moving into draft - it looks like cursor might be doing something weird here, I can't quite figure out what though.

jakemac53:
Ugh, so cursor actually claims to support roots even though it doesn't.

So, this fallback doesn't work for cursor.

I will add an argument to force the server to use the fallback support :/.

kenzieschmoll:
Generally, do we want to consider using the roots that DTD has if we already have a DTD connection established? This would be one less tool call and would provide a smoother experience for the user when DTD is available.

pkgs/dart_tooling_mcp_server/bin/main.dart
```diff
@@ -62,3 +62,17 @@ void main(List<String> args) async {
     ),
   );
 }
+
+final argParser =
+    ArgParser(allowTrailingOptions: false)
+      ..addFlag(
+        'force-roots-fallback',
```
kenzieschmoll, line 69: nit: use a named const for this string here and above

pkgs/dart_tooling_mcp_server/lib/src/mixins/roots_fallback_support.dart
```diff
@@ -0,0 +1,171 @@
...
+/// on the tool calls.
+base mixin RootsFallbackSupport on ToolsSupport, RootsTrackingSupport {
+  /// Set of custom roots
+  final Set<Root> _customRoots = HashSet<Root>(
+    equals: (a, b) => a.uri == b.uri,
+    hashCode: (root) => root.uri.hashCode,
+  );
+
+  /// Whether or not to force the fallback mode for roots, regardless of the
+  /// clients reported support.
```
kenzieschmoll, line 28:
```suggestion
  /// client's reported support.
```

pkgs/dart_tooling_mcp_server/lib/src/mixins/roots_fallback_support.dart
```diff
@@ -0,0 +1,171 @@
...
+      );
+    }
+
+    final roots =
+        (request.arguments![ParameterNames.roots] as List).cast<Root>();
+    for (final root in roots) {
+      _customRoots.add(root);
+    }
+    _rootsListChangedFallbackController?.add(RootsListChangedNotification());
+    return CallToolResult(content: [Content.text(text: 'Success')]);
```
kenzieschmoll, line 98: nit: should we have a common `successResponse` somewhere that we can reuse?

pkgs/dart_tooling_mcp_server/lib/src/mixins/roots_fallback_support.dart
```diff
@@ -0,0 +1,171 @@
...
+  /// otherwise returns our own custom roots.
+  @override
+  Future<ListRootsResult> listRoots(ListRootsRequest request) async =>
+      _fallbackEnabled
+          ? ListRootsResult(roots: _customRoots.toList())
+          : super.listRoots(request);
+
+  /// Adds the roots in [request] the custom roots and calls [updateRoots].
+  ///
+  /// Should only be called if [supportsRoots] is false.
```
kenzieschmoll, line 84: is this sentence accurate (here and elsewhere)? `supportsRoots` is always true in this mixin

pkgs/dart_tooling_mcp_server/lib/src/mixins/roots_fallback_support.dart
```diff
@@ -0,0 +1,171 @@
...
+  @visibleForTesting
+  static final removeRootsTool = Tool(
+    name: 'remove_roots',
+    description:
+        'Removes one or more project roots previously added via '
+        'the add_roots tool.',
+    annotations: ToolAnnotations(title: 'Remove roots', readOnlyHint: false),
+    inputSchema: Schema.object(
+      properties: {
+        ParameterNames.uris: Schema.list(
```
kenzieschmoll, line 164: it was initially surprising to me that add_roots and remove_roots have a different scheme for specifying the list of roots. Should we merge these? Or at least use the same parameter name `ParameterNames.roots` for the top level field in both?

pkgs/dart_tooling_mcp_server/lib/src/mixins/roots_fallback_support.dart
```diff
@@ -0,0 +1,171 @@
...
+      throw StateError(
+        'This tool should not be invoked if the client supports roots',
+      );
+    }
+
+    final roots =
+        (request.arguments![ParameterNames.roots] as List).cast<Root>();
+    for (final root in roots) {
+      _customRoots.add(root);
+    }
```
kenzieschmoll, line 96: nit this could be condensed:
`.forEach(_customRoots.add)`

pkgs/dart_tooling_mcp_server/test/tools/roots_fallback_support_test.dart
```diff
@@ -0,0 +1,232 @@
...
+        expect(
+          (await server.listRoots(ListRootsRequest())).roots,
+          unorderedEquals([rootA]),
+        );
+      });
+    });
+  });
+}
+
+// A test client that does not support roots
```
kenzieschmoll, line 189: /// for dart doc. here and below

pkgs/dart_tooling_mcp_server/lib/src/mixins/roots_fallback_support.dart
jakemac53: Updated to `super.supportsRoots`

pkgs/dart_tooling_mcp_server/lib/src/mixins/roots_fallback_support.dart
jakemac53: sure, done

pkgs/dart_tooling_mcp_server/lib/src/mixins/roots_fallback_support.dart
jakemac53: They are removed just by their URI, so I was trying to just keep it concise, and I wanted to make the name different just so the LLM doesn't  get confused about two identically named parameters but with different structures. Also, it is a list of `uri` strings so it just feels the most descriptive.

jakemac53:
> Generally, do we want to consider using the roots that DTD has if we already have a DTD connection established? This would be one less tool call and would provide a smoother experience for the user when DTD is available.

There are two reasons I ended up deciding not to go this route:

- I wasn't confident that roots would be set on DTD in the non-IDE scenarios.
- It would mean that tools which don't require a connected app now require a DTD connection, which they don't today.

PR #133, opened 2025-05-19
Move more tests to dart cli tests

jakemac53:
Specifically, this moves the hot reload and runtime errors tests to dart CLI tests.

All remaining tests are truly flutter specific.

Also, fixed hot reload for dart CLI apps.


pkgs/dart_tooling_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -320,43 +322,38 @@ base mixin DartToolingDaemonSupport
...
+        if (report.success == true) {
+          return CallToolResult(
+            content: [TextContent(text: 'Hot reload succeeded')],
+          );
+        } else {
+          return CallToolResult(
+            isError: true,
+            content: [TextContent(text: 'Hot reload failed')],
+          );
+        }
```
kenzieschmoll, line 356: nit: this can be made more terse
```dart
final success = report.success == true;
return CallToolResult(
  isError: !success,
  content: [TextContent(text: 'Hot reload ${success ? 'succeeded' : 'failed'}.'),
);
```

jakemac53:
Very weird but it looks like on stable we get all stderr events twice...

pkgs/dart_tooling_mcp_server/test/test_harness.dart
```diff
@@ -204,7 +204,7 @@ final class AppDebugSession {
     final process = await TestProcess.start(isFlutter ? 'flutter' : 'dart', [
       'run',
       '--no${isFlutter ? '' : '-serve'}-devtools',
-      if (!isFlutter) '--enable-vm-service',
+      if (!isFlutter) ...['--enable-vm-service=0'],
```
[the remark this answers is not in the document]
jakemac53: Oh, this is cruft I had separated them at one point into different args

pkgs/dart_tooling_mcp_server/test/tools/dtd_test.dart
```diff
@@ -25,65 +27,102 @@ void main() {
...
-        );
-        final tools = (await testHarness.mcpServerConnection.listTools()).tools;
-        final getWidgetTreeTool = tools.singleWhere(
-          (t) => t.name == DartToolingDaemonSupport.getWidgetTreeTool.name,
-        );
-        final getWidgetTreeResult = await testHarness.callToolWithRetry(
-          CallToolRequest(name: getWidgetTreeTool.name),
-        );
+      group('dart cli tests', () {
+        test('can perform a hot reload', () async {
```
kenzieschmoll, line 76: seems like we should still have testing that we can hot reload a flutter app since this is a different code path than hot reloading a dart app (detecting the hot reload service name)

jakemac53: I need to separately validate that the custom flutter logic is actually necessary as well.

But I do generally agree, I am still not sure how we are going to manage these flutter tests exactly though.

jakemac53: I will add it back for now, we have other ones anyways

kenzieschmoll (approved):

Issue #134, opened 2025-05-19
Runtime errors resource/tool can easily exceed context window limits


jakemac53:
We should be paginating or trimming these results so we don't use an unreasonable amount of tokens.

jakemac53:
Going to go ahead and just call this a bug as it easily gets you in bad situations.

jakemac53:
For instance you can get errors that trigger on every frame and quickly get into the millions of tokens

PR #139, opened 2025-05-21
Limit the tokens returned by the runtime errors tool/resource

jakemac53:
Fixes https://github.com/dart-lang/ai/issues/134

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -966,11 +966,11 @@ extension type DebugSession.fromJson(Map<String, Object?> _value)
 class ErrorLog {
   Iterable<String> get errors => _errors;
   final List<String> _errors = [];
-  int _size = 0;
+  int _characters = 0;
 
   /// The number of characters used by all errors in the log.
   @visibleForTesting
-  int get size => _size;
+  int get characters => _characters;
 
   final int _maxSize;
 
@@ -987,16 +987,16 @@ class ErrorLog {
       // all other errors.
       final trimmed = error.substring(0, _maxSize);
       _errors.clear();
-      _size = trimmed.length;
+      _characters = trimmed.length;
       _errors.add(trimmed);
     } else {
       // Otherwise, we append the error and then remove as many errors from the
       // front as we need to in order to get under the max size.
-      _size += error.length;
+      _characters += error.length;
       _errors.add(error);
       var removeCount = 0;
-      while (_size > _maxSize) {
-        _size -= _errors[removeCount].length;
+      while (_characters > _maxSize) {
+        _characters -= _errors[removeCount].length;
         removeCount++;
       }
       _errors.removeRange(0, removeCount);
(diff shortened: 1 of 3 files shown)


devoncarew:
> cc @devoncarew are you aware of this labeler permissions issue?
> 
> **EDIT:**: I added issues: write per [actions/labeler#870](https://github.com/actions/labeler/issues/870) and it seems to be happy now

I believe it now has trouble creating labels, but can apply if the label exists. That might be new; is that would you're seeing? Or trouble applying a label? If the solution is that permission update, we should apply across our various monorepos.

jakemac53:
> I believe it now has trouble creating labels, but can apply if the label exists. That might be new; is that would you're seeing? Or trouble applying a label? If the solution is that permission update, we should apply across our various monorepos.

The error is not descriptive, it seems like it fails to even attempt to do anything after seeing it doesn't have write permissions for labels. I manually had to add the label here.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -383,14 +385,14 @@ base mixin DartToolingDaemonSupport
             content: [
               TextContent(
                 text:
-                    'Found ${errors.length} '
-                    'error${errors.length == 1 ? '' : 's'}:\n',
+                    'Found ${errorLog.errors.length} '
+                    'error${errorLog.errors.length == 1 ? '' : 's'}:\n',
               ),
-              ...errors.map((e) => TextContent(text: e.toString())),
+              ...errorLog.errors.map((e) => TextContent(text: e.toString())),
```
natebosch, line 391: [nit]

```suggestion
              for (final e in errorLog.errors)
                TextContent(text: e.toString()),
```

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -958,3 +960,52 @@ extension type DebugSession.fromJson(Map<String, Object?> _value)
     if (vmServiceUri != null) 'vmServiceUri': vmServiceUri,
   });
 }
+
+@visibleForTesting
+/// Manages a log of errors with a maximum size in terms of total characters.
+class ErrorLog {
```
natebosch, line 966: [nit] https://dart.dev/effective-dart/documentation#do-put-doc-comments-before-metadata-annotations

```suggestion

/// Manages a log of errors with a maximum size in terms of total characters.
@visibleForTesting
class ErrorLog {
```

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -958,3 +960,52 @@ extension type DebugSession.fromJson(Map<String, Object?> _value)
...
+  Iterable<String> get errors => _errors;
+  final List<String> _errors = [];
+  int _size = 0;
+
+  /// The number of characters used by all errors in the log.
+  @visibleForTesting
+  int get size => _size;
```
natebosch, line 973: [optional] Consider renaming to `characters` or `totalCharacters`.

natebosch (approved):

PR #140, opened 2025-05-21
Flutter/Dart create tool

jakemac53:
Adds a `create_project` tool, which takes a project kind (flutter or dart) and runs the appropriate command.

Requires a root to keep things on the rails, but accepts a relative path anywhere under the root to create the app in.

Also accepts a template configuration.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
@@ -158,7 +158,10 @@ base mixin DashCliSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
   static final createProjectTool = Tool(
     name: 'create_project',
     description: 'Creates a new Dart or Flutter project.',
-    annotations: ToolAnnotations(title: 'Create project'),
+    annotations: ToolAnnotations(
+      title: 'Create project',
+      destructiveHint: true,
+    ),
     inputSchema: Schema.object(
       properties: {
         ParameterNames.root: rootSchema,
--- a/pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:dart_mcp/server.dart';
+import 'package:path/path.dart' as p;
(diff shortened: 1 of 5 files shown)


pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
```diff
@@ -71,6 +73,61 @@ base mixin DashCliSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
     );
   }
 
+  /// Implementation of the [createProjectTool].
+  Future<CallToolResult> _runCreateProjectTool(CallToolRequest request) async {
+    final args = request.arguments;
+
+    final errors = createProjectTool.inputSchema.validate(args);
+    final projectType = args?[ParameterNames.projectType] as String?;
+    if (projectType == null ||
```
kenzieschmoll, line 82: nit: we can remove the null check since a null value for `projectType` will fail the following two checks too

pkgs/dart_mcp_server/lib/src/utils/cli_utils.dart
```diff
@@ -91,103 +91,150 @@ Future<CallToolResult> runCommandInRoots(
...
+  List<String> arguments = const [],
+  required String commandDescription,
+  required FileSystem fileSystem,
+  required ProcessManager processManager,
+  required List<Root> knownRoots,
+  List<String> defaultPaths = const <String>[],
+}) async {
+  rootConfig ??= request.arguments;
+  final rootUriString = rootConfig?[ParameterNames.root] as String?;
+  if (rootUriString == null) {
```
kenzieschmoll, line 148: nit: we could check `rootConfig?.containsKey(ParameterNames.root) ?? false` instead

pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
```diff
@@ -71,6 +73,61 @@ base mixin DashCliSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
...
+        projectType != 'dart' && projectType != 'flutter') {
```
jakemac53, line 83:
```suggestion
    if (projectType != 'dart' && projectType != 'flutter') {
```

kenzieschmoll (approved):

Issue #145, opened 2025-05-26
`pub_dev_search` returns a lot of content under `api.qualifiedNames`


domesticmouse:
I'm looking at the returned result for `pub_dev_search` for the term `riverpod`. Thinking about the token burn, I don't see the value in the content returned under `api.qualifiedNames`. Can we elide this information?

```json
{
  "packageName": "riverpod",
  "latestVersion": "2.6.1",
  "description": "A reactive caching and data-binding framework. Riverpod makes working with asynchronous code a breeze.\n",
  "scores": {
    "pubPoints": 160,
    "maxPubPoints": 160,
    "likes": 3665,
    "downloadCount": 1812054
  },
  "topics": [],
  "licenses": [
    "license:mit",
    "license:fsf-libre",
    "license:osi-approved"
  ],
  "publisher": "publisher:dash-overflow.net",
  "api": {
    "qualifiedNames": [
      "riverpod",
      "riverpod.AlwaysAliveProviderBase",
      "riverpod.AlwaysAliveProviderBase.select",
      "riverpod.AlwaysAliveProviderListenable",
      "riverpod.AlwaysAliveProviderListenable.select",
      "riverpod.AlwaysAliveRefreshable",
      "riverpod.AlwaysAliveRefreshable.AlwaysAliveRefreshable.new",
      "riverpod.AsyncData",
      "riverpod.AsyncData.AsyncData.new",
      "riverpod.AsyncData.copyWithPrevious",
      "riverpod.AsyncData.error",
      "riverpod.AsyncData.hasValue",
      "riverpod.AsyncData.isLoading",
      "riverpod.AsyncData.map",
      "riverpod.AsyncData.stackTrace",
      "riverpod.AsyncData.value",
      "riverpod.AsyncError",
      "riverpod.AsyncError.AsyncError.new",
      "riverpod.AsyncError.copyWithPrevious",
      "riverpod.AsyncError.error",
      "riverpod.AsyncError.hasValue",
      "riverpod.AsyncError.isLoading",
      "riverpod.AsyncError.map",
      "riverpod.AsyncError.stackTrace",
      "riverpod.AsyncError.value",
      "riverpod.AsyncLoading",
      "riverpod.AsyncLoading.AsyncLoading.new",
      "riverpod.AsyncLoading.copyWithPrevious",
      "riverpod.AsyncLoading.error",
      "riverpod.AsyncLoading.hasValue",
      "riverpod.AsyncLoading.isLoading",
[...]

PR #146, opened 2025-05-27
remove qualifiedNames from the pub dev api search

jakemac53:
Closes https://github.com/dart-lang/ai/issues/145

I did some testing myself and we easily get >10k tokens added to the context for a single query which is definitely too much for the value provided here.


jakemac53:
We may want to consider a separate tool to get API information for a specific package or library in the future though.

kenzieschmoll (approved):

domesticmouse (approved):

sigurdm (approved):

PR #147, opened 2025-05-29
create the working dir if it doesn't exist

jakemac53:
You can't set a working dir if it doesn't exist, this ensures the dir does actually exist. LLMs like to sometimes set a root that is a subdirectory of an actual root, which is generally fine, but we need to make sure to create it.


gspencergoog (approved):

gspencergoog:
Ahh, nevermind... :-)

jakemac53:
> You might add a test to make sure the directory gets created if it didn't exist before.

Heh yeah I was in the middle of writing it when you sent this :)

PR #148, opened 2025-05-29
Be more flexible about roots by comparing canonicalized paths

jakemac53:
Instead of using `isWithin` and also checking for exact equality, we compare the scheme/authority, and then compare the canonicalized paths.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/utils/cli_utils.dart
+++ b/pkgs/dart_mcp_server/lib/src/utils/cli_utils.dart
@@ -262,7 +262,7 @@ Future<String> defaultCommandForRoot(
     ),
 };
 
-/// Returns whether [uri] is under [root].
+/// Returns whether [uri] is under or exactly equal to [root].
 ///
 /// Relative uris will always be under [root] unless they escape it with `../`.
 bool _isUnderRoot(Root root, String uri) {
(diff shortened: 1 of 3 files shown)


pkgs/dart_mcp_server/lib/src/utils/cli_utils.dart
```diff
@@ -268,8 +268,18 @@ Future<String> defaultCommandForRoot(
...
+    // match.
+    if (knownRootUri.scheme != resolvedRootUri.scheme ||
+        knownRootUri.authority != resolvedRootUri.authority) {
+      continue;
+    }
+    // Canonicalizing the paths handles any `../` segments and also deals with
+    // trailing slashes versus no trailing slashes.
+    final canonicalKnownPath = p.canonicalize(knownRootUri.path);
+    final canonicalRootPath = p.canonicalize(resolvedRootUri.path);
+    if (canonicalRootPath.startsWith(canonicalKnownPath)) {
```
gspencergoog, line 282: Why not use `isWithin`? Seems like it would be more exhaustive, especially for non-Linux systems.

pkgs/dart_mcp_server/lib/src/utils/cli_utils.dart
```diff
@@ -263,17 +262,23 @@ Future<String> defaultCommandForRoot(
...
-  for (final root in knownRoots) {
-    final knownRootUri = Uri.parse(root.uri);
-    final resolvedRoot = knownRootUri.resolve(rootUri).toString();
-    if (root.uri == resolvedRoot || p.isWithin(root.uri, resolvedRoot)) {
-      return root;
-    }
+/// Returns whether [uri] is under [root].
+///
+/// Relative uris will always be under [root] unless they escape it with `../`.
+bool _isUnderRoot(Root root, String uri) {
```
gspencergoog, line 268: So, this means that "under" always includes the root directory itself.  Is that OK for all the tools?  Would there be some where the operation shouldn't be allowed on the root directory itself?

I can't think of any, but I just wanted to point that out.

Maybe just point out in the docs that it always considers the root itself to be "under" the root.

gspencergoog (approved):

Issue #32, opened 2025-04-04
Add macos/windows testing support for dart_tooling_mcp_server


jakemac53:
https://github.com/dart-lang/ai/pull/31 adds linux support for integration tests, mac and windows should also be doable but we will have to figure out what needs to be installed in order to be able to run on those platforms.

PR #149, opened 2025-05-29
run tests on windows, fix executable paths for dart/flutter

jakemac53:
Closes https://github.com/dart-lang/ai/issues/32

I believe this should just work now that we are using headless flutter

Initially also added macos tests but those bots tend to wait queued for a long time, so I have removed them.

### DIFF
--- a/pkgs/dart_mcp_server/test/test_harness.dart
+++ b/pkgs/dart_mcp_server/test/test_harness.dart
@@ -498,6 +498,6 @@ class _CommandMatcher extends Matcher {
 }
 
 extension RootPath on Root {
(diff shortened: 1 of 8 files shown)


pkgs/dart_mcp_server/test/test_harness.dart
```diff
@@ -497,5 +498,6 @@ class _CommandMatcher extends Matcher {
 }
 
 extension RootPath on Root {
-  String get path => Uri.parse(uri).path;
+  // Get the OS specific file path for this root.
```
kenzieschmoll: /// for dart doc

PR #152, opened 2025-06-04
handle relative paths under roots without trailing slashes

jakemac53:
This resolves an issue I was seeing where if an LLM sets a root without a trailing slash, then tools which accept paths would think the path was escaping the root, because `Uri.parse('/foo/bar').resolve('zap')` results in a uri with the path `/foo/zap`).


gspencergoog (approved):

jakemac53:
> Actually, I noticed that `_isUnderRoot` uses `p` to resolve paths, and it should probably use `fileSystem.path` instead, to keep it using the right path types.

Interesting, I didn't know that package:file exposed the path context like that, updated

Issue #154, opened 2025-06-05
[feat] provide the link to the docs and the repo from the pub.dev search tool


csells:
I wanted Cursor to look up a package and show me some sample code, but it the pub.dev search tool didn't provide the link to the docs, the README or the github repo. Can you add that?

Here's the chat transcript:

## Chat Transcript

**User**

can you search pub.dev for all of the mcp-related packages and compare and contrast them for me?

---

**Cursor**

I am Claude Sonnet 4, currently powered by Anthropic's AI technology.

I'll search pub.dev for MCP-related packages and compare them for you.

Let me also search for "model context protocol" to see if there are any additional packages:

Based on the search results, here's a comprehensive comparison of the MCP-related packages available on pub.dev:

### **1. mcp_dart** (Most Popular) ⭐
- **Version**: 0.5.1  
- **Downloads**: 1,873 (highest)  
- **Likes**: 32 (highest)  
- **Publisher**: leehack.com  
- **License**: MIT  
- **Description**: Full Dart SDK implementation of Model Context Protocol
- **Best for**: General MCP development in Dart

### **2. dart_mcp** (Official Dart Team)
- **Version**: 0.2.1  
- **Downloads**: 674  
- **Likes**: 33  
- **Publisher**: labs.dart.dev (Official Dart team)  
- **License**: BSD-3-Clause  
- **Description**: Package for making MCP servers and clients
- **Best for**: Official support and stability

### **3. straw_mcp**
- **Version**: 0.5.0  
- **Downloads**: 35  
- **Likes**: 2  
- **License**: Apache-2.0  
- **Description**: Dart MCP implementation for LLM service integration
- **Best for**: Alternative implementation approach

### **4. mcp_server**
- **Version**: 0.2.0  
- **Downloads**: 221  
- **Likes**: 4  
- **License**: MIT  
- **Description**: Focused on building MCP servers
- **Best for**: Server-side MCP implementations

### **5. mcp_client**
- **Version**: 0.1.8  
- **Downloads**: 214  
- **Likes**: 4  
- **License**: MIT  
- **Description**: Focused on building MCP clients
- **Best for**: Client-side MCP implementations

### **6. mcp_bridge**
- **Version**: 0.1.0  
- **Downloads**: 31  
- **Likes**: 2  
- **License**: MIT  
- **Description**: MCP bridge implementation
- **Best for**: Bridging different MCP systems

### **7. flutter_mcp_client_lib**
- **Version**: 1.0.3  
[...]

PR #155, opened 2025-06-05
add homepage, repository, and documentation links to the pub result

jakemac53:
Closes https://github.com/dart-lang/ai/issues/154

This won't automatically include examples, as that would be a fair bit more involved, but it will include the documentation link if provided by the package.

### DIFF
--- a/pkgs/dart_mcp_server/CHANGELOG.md
+++ b/pkgs/dart_mcp_server/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Version 0.1.0 (Dart SDK 3.8.0)
 
+* Add documentation/homepage/repository links to pub results.
 * Handle relative paths under roots without trailing slashes.
 * Fix executable paths for dart/flutter on windows.
 * Pass the provided root instead of the resolved root for project type detection.
--- a/pkgs/dart_mcp_server/lib/src/mixins/pub_dev_search.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/pub_dev_search.dart
@@ -74,6 +74,9 @@ base mixin PubDevSupport on ToolsSupport {
             (packageName) => (
               versionListing: retrieve('api/packages/$packageName'),
               score: retrieve('api/packages/$packageName/score'),
+              docIndex: retrieve(
+                'documentation/$packageName/latest/index.json',
+              ),
             ),
           )
           .toList();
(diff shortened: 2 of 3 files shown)


jakemac53:
cc @csells

jakemac53:
One open question, should we default the documentation link to `https://pub.dev/documentation/<package>/latest/`?

Also, should we link to the actual pub page? That would show the readme content at least

gspencergoog:
> One open question, should we default the documentation link to `https://pub.dev/documentation/<package>/latest/`?

I think yes.  The latest one is most likely the one most in use (especially for the majority of packages which haven't released in a while).

> Also, should we link to the actual pub page? That would show the readme content at least

Maybe linking to the README itself might be more helpful?  It'll be in Markdown, which the LLM would grok better anyhow, and because it's called "README" might be more attractive to the LLM to actually fetch and read.

jakemac53:
> Maybe linking to the README itself might be more helpful? It'll be in Markdown, which the LLM would grok better anyhow, and because it's called "README" might be more attractive to the LLM to actually fetch and read.

I am not sure we can reliably link to the README is the issue. We could potentially do that if the repo/homepage are GitHub links, as we know the rough structure of GitHub.

jakemac53:
> I think yes. The latest one is most likely the one most in use (especially for the majority of packages which haven't released in a while).

Separate but related question, should we just always include this, separate from the `documentation` link? I have noticed that if packages include a `documentation` link, it tends to be a custom page.

gspencergoog:
> I am not sure we can reliably link to the README is the issue. We could potentially do that if the repo/homepage are GitHub links, as we know the rough structure of GitHub.

In that case, I think the pub page is probably the next best thing.

jakemac53:
> Would `https://pub.dev/documentation/<package>/latest/<package>/` make sense (e.g. [this one for Riverpod](https://pub.dev/documentation/riverpod/latest/riverpod/)), since it's the "top" of the documentation tree for the API, and would list many more symbols?

That won't always work for all packages because it assumes they have a file at `lib/<package-name>.dart`. Yes, most packages do have that, but not all.

jakemac53:
cc @sigurdm do you know if the library names correspond to import paths (can we do `package:<package-name>/<library-name>.dart` always)? Or is there another way to get the import path?

gspencergoog:
> Ok @gspencergoog I added docs links for all the libraries now :)

Nice!

gspencergoog:
> **Update**: Looks like it won't work out of the box today, [pub.dev/documentation/test/latest/bootstrap_browser](https://pub.dev/documentation/test/latest/bootstrap_browser/) is an example, we just get the name `browser` for the library...

The pub.dev site must have a way to figure out all of the library links: it displays them.  Can we ask them to add that info to the API?

jakemac53:
> The pub.dev site must have a way to figure out all of the library links: it displays them. Can we ask them to add that info to the API?

That is why I have cc'd @sigurdm on here :). It might be harder than you would think, those pages are dartdoc output which afaik you cannot get in a structured form. Clearly pub is already doing _something_ here though to extract all the symbols in a package so hopefully it is possible to also provide the library imports.

sigurdm (commented):
LGTM

PR #160, opened 2025-06-05
Always use the failures-only reporter for testing

natebosch:
The default reporter has verbose output which is not useful but eats up
a lot of input tokens once it's in the history. Always run tests with
`--reporter=failures-only` to reduce the output size. This output format
still includes a count of passing tests.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
@@ -68,7 +68,7 @@ base mixin DashCliSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
   Future<CallToolResult> _runTests(CallToolRequest request) async {
     return runCommandInRoots(
       request,
-      arguments: ['test'],
+      arguments: ['test', '--reporter=failures-only'],
       commandDescription: 'dart|flutter test',
       processManager: processManager,
       knownRoots: await roots,
--- a/pkgs/dart_mcp_server/test/tools/dart_cli_test.dart
+++ b/pkgs/dart_mcp_server/test/tools/dart_cli_test.dart
@@ -166,13 +166,19 @@ dependencies:
           command: [
             endsWith(flutterExecutableName),
             'test',
+            '--reporter=failures-only',
             'foo_test.dart',
             'bar_test.dart',
           ],
           workingDirectory: exampleFlutterAppRoot.path,
         )),
         equalsCommand((
-          command: [endsWith(dartExecutableName), 'test', 'zip_test.dart'],
+          command: [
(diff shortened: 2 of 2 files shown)


natebosch:
I have not had much success so far in prodding clients into following specific instructions for using any of the tools. I also can't consistently get clients to run the actual CLI interfaces following specific instructions provided by the MCP server. It's all pretty easy to handle with _user_ specified config, but difficult to make a responsibility of the MCP tooling.

I _think_ we'll have better luck getting the clients to interact with the MCP tools in the ways we want than the direct CLI so we should try to make each of them as useful for model input as possible. An easy win is to reduce the output size for tests, so land that first.

kenzieschmoll (approved):

pkgs/dart_mcp_server/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # 0.1.1-wip
 
 * Instruct clients to prefer MCP tools over running tools in the shell.
+* Reduce output size of `run_tests` tool to save on input tokens.
```
jakemac53: Same comment as other PR, I would move this into the section below as we will do more rolls before the last beta release for 3.8.

jakemac53 (approved):

pkgs/dart_mcp_server/CHANGELOG.md
```diff
@@ -1,9 +1,4 @@
-# 0.1.1-wip
-
-* Instruct clients to prefer MCP tools over running tools in the shell.
-* Reduce output size of `run_tests` tool to save on input tokens.
-
-# 0.1.0 (Dart SDK 3.8.0)
+# 0.1.0-wip (Dart SDK 3.8.0)
```
jakemac53: The `-wip` is a bit odd, I do get that it is more clear that new entries should be added to it. But, it is already "released" in the sense that it is rolled into the SDK. I don't really want to push a last minute merge into the SDK just before the final beta cut just to change the version. So, that is why I marked it as not -wip, even though it is wip 🤣 .

Anyways, I don't feel strongly here, we could just remove the -wip at some point later after the beta cut happens, and just accept that in the SDK the version will always be wip.

jakemac53: We could do several version bumps, versioning it each time we want to merge into the SDK, but that will add extra process to doing those SDK merges (an extra PR each time) and for little value.

We could also just remove the version number here entirely.

natebosch: As discussed separately - I dropped the package level version number and made the header of the changelog just the SDK version reference with a `WIP` marker to prioritize easy edits with an (optional) step to remove it for the stable SDK publish.

PR #161, opened 2025-06-05
Instruct clients to prefer MCP

natebosch:
Clients may be influenced by things like instructions for running
tests in `README.md` and then decided to run `dart test` in a shell
instead of using the `run_tests` MCP tool. If we are providing a MCP
version of a tool it may be tailored to the LLM client use case and so
should be preferred. Add a server level instruction.

### DIFF
--- a/pkgs/dart_mcp_server/CHANGELOG.md
+++ b/pkgs/dart_mcp_server/CHANGELOG.md
@@ -1,4 +1,8 @@
-# Version 0.1.0 (Dart SDK 3.8.0)
+# 0.1.1-wip
+
+* Instruct clients to prefer MCP tools over running tools in the shell.
+
+# 0.1.0 (Dart SDK 3.8.0)
 
 * Handle relative paths under roots without trailing slashes.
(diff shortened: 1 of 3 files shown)


natebosch:
Some clients seems to use MCP more reliably than others, this instruction makes it more consistent.

csells:
Love this. I was running into this today. Thank you!

kenzieschmoll (approved):

pkgs/dart_mcp_server/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-# Version 0.1.0 (Dart SDK 3.8.0)
+# 0.1.1-wip
+
+* Instruct clients to prefer MCP tools over running tools in the shell.
```
jakemac53, line 3: This can actually move into the below version, I plan to do a roll soon so it will make it into the 3.8 SDK.

The versions here are weird since its not actually published but I did want some sort of changelog.

PR #169, opened 2025-06-17
Add JSON schema for test runner arguments

natebosch:
Add strong language to always use the `run_tests` tool.

Add a JSON schema for the CLI arguments to the test runner. The schema
was dumped using https://github.com/dart-lang/core/pull/897

Add a test for the conversion of boolean, string, and list of string
arguments, as well as overriding the reporter argument.


natebosch:
I haven't gotten an eval framework set up, but my manual testing was pretty positive with this. The agent was able to use the test runner more effectively, and I was unsuccessful in attempts to coach usage in other ways like a `getToolInfo` tool.

pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
```diff
@@ -245,3 +265,24 @@ base mixin DashCliSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
...
...\\"false\\\""},"test-randomize-ordering-seed":{"type":"string","description":"Use the specified seed to randomize the execution order of test cases.\\nMust be a 32bit unsigned integer or \\\"random\\\".\\nIf \\\"random\\\", pick a random seed to use.\\nIf not passed, do not randomize test case execution order."},"fail-fast":{"type":"boolean","description":"Stop running tests after the first failure.\\n\\ndefaults to \\\"false\\\""},"reporter":{"type":"string","description":"Set how to print test results.\\ndefaults to \\\"compact\\\"\\nallowed values: compact, expanded, failures-only, github, json, silent"},"file-reporter":{"type":"string","description":"Enable an additional reporter writing test results to a file.\\nShould be in the form <reporter>:<filepath>, Example: \\\"json:reports/tests.json\\\""},"verbose-trace":{"type":"boolean","description":"Emit stack traces with core library frames.\\ndefaults to \\\"false\\\""},"js-trace":{"type":"boolean","description":"Emit raw JavaScript stack traces for browser tests.\\ndefaults to \\\"false\\\""},"color":{"type":"boolean","description":"Use terminal colors.\\n(auto-detected by default)\\ndefaults to \\\"false\\\""}},"required":[]}
```
jakemac53: Why is this a String, how are we intended to update it?

jakemac53: I am also not sure many of these make sense to expose and are just a waste of tokens and will make overall calls less reliable?

For instance `color`, `help`, `file-reporter`, `pause-after-load`, etc?

natebosch: > Why is this a String, how are we intended to update it?

We'd dump it from the test runner binary again and replace the entire String. We can change this away from a String, it will add steps to update, but I don't think that's a problem. Would you prefer a map literal?

> I am also not sure many of these make sense to expose and are just a waste of tokens and will make overall calls less reliable?
> 
> For instance `color`, `help`, `file-reporter`, `pause-after-load`, etc?

I don't think it makes sense to curate the whole list manually, but defining a blocklist is feasible. I'd worry about forgetting to add an argument sometime and leaving a gap where the client needs to use the shell even though we said not to. WDYT about a blocklist defined here? If a blocklist defined by the test runner is better we'd need to update the `package:args` API.

natebosch: I think `file-reporter` has plausible utility for a client, I added a blocklist with the others.

For the overall token usage I'm partially running on a hunch that some sort of tool filtering is likely to become a standard feature for clients. There is a huge number of tools being created and a number of discussions started about managing the volume of tools. I started exploring tool filtering by making a call with less historical context to decide which tools should offered but I didn't figure out how to evaluate the behavior impacts.

jakemac53: Blocklist sounds fine to me

natebosch: I added instructions for updating this. It currently relies on unmerged PRs in `package:args` and `package:test`. I'd be fine seeing these land if someone wants to tackle it before I'm back, otherwise I think it's OK if these sit as drafts just in case they are needed and I'll revisit whether it's worth landing in a few weeks.
https://github.com/dart-lang/test/pull/2508

jakemac53 (approved):

PR #174, opened 2025-06-17
add analytics support to the Dart MCP server

jakemac53:
Based on https://github.com/dart-lang/tools/pull/2112

Adds analytics events for all tool calls as well as for the runtime errors resource.

We could consider a generic resource read event which includes the URI, but we could accidentally create URIs that have some information we don't want to capture, so I chose to instead just customize this and not provide the URI, but a field that describes the type of resource that was read.

I did not add any debouncing because I don't think it is necessary, LLMs do not invoke tools in super rapid succession generally, but let me know if you all disagree and think I should do some debouncing.


pkgs/dart_mcp_server/bin/main.dart
```diff
@@ -40,6 +40,7 @@ void main(List<String> args) async {
             ),
         forceRootsFallback: parsedArgs.flag(forceRootsFallback),
         sdk: Sdk.find(dartSdkPath: dartSdkPath, flutterSdkPath: flutterSdkPath),
+        analytics: null,
```
kenzieschmoll, line 46: Can we actually create the Analytics instance here?

pkgs/dart_mcp_server/lib/src/utils/analytics.dart
```diff
@@ -0,0 +1,72 @@
...
+
+  ReadResourceMetrics({
+    required this.kind,
+    required this.length,
+    required this.elapsedMilliseconds,
+  });
+
+  @override
+  Map<String, Object> toMap() => {
+    'kind': 'runtimeErrors',
```
kenzieschmoll: should this be using the `kind` field instead of the 'runtimeErrors' string?

pkgs/dart_mcp_server/lib/src/utils/analytics.dart
```diff
@@ -0,0 +1,72 @@
...
+/// methods can be easily mocked during testing.
+abstract interface class AnalyticsSupport {
+  Analytics? get analytics;
+
+  ClientImplementation get clientInfo;
+}
+
+enum AnalyticsEvent { callTool, readResource }
+
+final class ReadResourceMetrics extends CustomMetrics {
```
kenzieschmoll, line 19: add dart doc for this class and for CallToolMetrics

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -114,12 +117,32 @@ base mixin DartToolingDaemonSupport
...
+          watch.stop();
+          try {
+            analytics?.send(
+              ua.Event.dartMCPEvent(
+                client: clientInfo.name,
+                clientVersion: clientInfo.version,
+                serverVersion: implementation.version,
+                type: AnalyticsEvent.readResource.name,
+                additionalData: ReadResourceMetrics(
+                  kind: 'runtimeErrors',
```
kenzieschmoll: nit: consider keeping constants in the analytics file or in another analytics_constants.dart file so that that it is easy to locate what strings to query for.

pkgs/dart_mcp_server/lib/src/utils/analytics.dart
```diff
@@ -0,0 +1,72 @@
...
+    required this.tool,
+    required this.success,
+    required this.elapsedMilliseconds,
+  });
+
+  @override
+  Map<String, Object> toMap() => {
+    'tool': tool,
+    'success': success,
+    'elapsedMilliseconds': elapsedMilliseconds,
```
kenzieschmoll: nit: consider storing keys as constants so typos don't creep in. For example, if a typo slipped in, then this could make running queries over this data miss fields in a non-obvious way.

pkgs/dart_mcp_server/bin/main.dart
jakemac53: If we are allowed to that would simplify things, otherwise we need a custom entry point in the SDK which does create one.

I just don't know if it is kosher or not, it feels a bit weird to be putting the current tool as the `dartTool`, when this is actually a binary that can be ran in other ways (from source, or as a git dep).

jakemac53: cc @anderdobo

jakemac53: Another option would be a command line flag to enable analytics.

kenzieschmoll: Can we look up the resolved executable to see if the starting script was from the `dart mcp-server` command?

jakemac53: I don't think so, we will just get the dart native runtime as the resolved executable afaik.

pkgs/dart_mcp_server/lib/src/utils/analytics.dart
jakemac53: yes 🤣

pkgs/dart_mcp_server/bin/main.dart
kenzieschmoll: Hmm. Then if we need to use the Dart CLI as the parent tool, adding a hidden flag that is set from the dart mcp-server command only could be a good workaround. We should get clarity on that first from Ander though.

jakemac53: It sounds like we would have to support showing the message to the user etc in that case, which we have no way of doing as an MCP server, so I think we will need a custom entry point in the SDK.

kenzieschmoll (approved):

Issue #173, opened 2025-06-17
MCPServer class should expose the client info given during initialization


jakemac53:
You can get this today by implementing your own `initialize` function but it is a bit silly to have to, we should just store it by default and provide a getter to access it.

PR #175, opened 2025-06-17
Use shared Implementation type, add clientInfo field to MCPServer

jakemac53:
In the [latest schema](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-03-26/schema.json) the ClientImplementation and ServerImplementation types were merged into a single Implementation type, this updates things to match.

Also closes https://github.com/dart-lang/ai/issues/173 which will make https://github.com/dart-lang/ai/pull/174 slightly cleaner, and matches what we do for ServerConnection as well.


jakemac53:
I also went ahead and bumped the version to release this right away, there aren't any other changes planned for `dart_mcp`.

Issue #177, opened 2025-06-18
Improve the error message when "Connect to DTD" tool fails


DanTup:
If I restart DTD in VS Code, interacting with Copilot Chat will sometimes trigger a call to "Connect to DTD" with a stable URI (because it reuses the last URI it fetched, unaware that it has changed). This results in an exception, which prevents the action from getting any further:

![Image](https://github.com/user-attachments/assets/35d004ca-1755-46e4-aa09-b310f9dfed86)

If I tell it to "try again" then it re-requests the DTD URI from Dart-Code and everything works.

I wonder if improving the error message when "Connect to DTD" fails may help the model understand that it should try getting an updated DTD URI by calling the "Get DTD URI" tool again? (for example if it suggesting "checking the DTD URI", instead of just providing the raw exception text)

jakemac53:
It seems like it did understand it needed to update its DTD Uri in this case, given that it requested it when you clicked try again. So, I am not sure it would necessarily help to change the text.

We could handle some expected error cases though and give a nicer error for those.

jakemac53:
Ok I sent out https://github.com/dart-lang/ai/pull/178, which tries to tackle this from two angles. Improve the error message for web socket errors (suggest getting a new URI), as well as improving the tool description, telling it to get a new DTD Uri before reconnecting.

jakemac53:
> I actually was unable to reproduce my original issue today. No matter what I do, it seems to always fetch the DTD URI again.

This is the struggle with all things LLMs haha. But, I think the changes should help prevent it from happening in any case.

PR #178, opened 2025-06-18
improve tool description for the dtd connection tool and improve error messages

jakemac53:
Closes https://github.com/dart-lang/ai/issues/177

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -4,6 +4,8 @@
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:ffi';
+import 'dart:io';
 
 import 'package:dart_mcp/server.dart';
 import 'package:dds_service_extensions/dds_service_extensions.dart';
@@ -182,10 +184,19 @@ base mixin DartToolingDaemonSupport
       return CallToolResult(
         content: [TextContent(text: 'Connection succeeded')],
       );
+    } on WebSocketException catch (_) {
+      return CallToolResult(
+        isError: true,
+        content: [
+          Content.text(
+            text: 'Connection failed, make sure your DTD Uri is up to date.',
+          ),
+        ],
+      );
     } catch (e) {
       return CallToolResult(
         isError: true,
-        content: [TextContent(text: 'Connection failed: $e')],
+        content: [Content.text(text: 'Connection failed: $e')],
       );
     }
   }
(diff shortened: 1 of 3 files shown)


pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -604,7 +615,8 @@ base mixin DartToolingDaemonSupport
     description:
         'Connects to the Dart Tooling Daemon. You should ask the user for the '
```
DanTup: > You should ask the user

I wonder if we should be less explicit here? The default model I get in Copilot Chat (GPT-4.1) seems to do what we want (use the tool) anyway, but since you added "Do not just make up a random URI" I'm not sure if they're all going to be that smart 🙃

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -4,6 +4,8 @@
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:ffi';
+import 'dart:io';
```
kenzieschmoll: are both these new imports needed?

kenzieschmoll (approved):

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: I think if they have the tool to get the URI, they will use it (even without that text to not make one up).

Maybe we should be a bit less prescriptive here though, given there are going to be tools available in some environments now, and the user doesn't always have to provide it.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: Weird.... not sure how these got added.... auto import gone awry I guess.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
DanTup: Perhaps something like "This should come from a suitable tool or from the user, ..."?

Specifying logic in English like this feels really weird 🙂

jakemac53: Updated the text

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -602,9 +612,11 @@ base mixin DartToolingDaemonSupport
   static final connectTool = Tool(
     name: 'connect_dart_tooling_daemon',
     description:
-        'Connects to the Dart Tooling Daemon. You should ask the user for the '
-        'dart tooling daemon URI, and suggest the "Copy DTD Uri to clipboard" '
-        'command. Do not just make up a random URI to pass.',
+        'Connects to the Dart Tooling Daemon. You should get the uri either '
+        'from available tools or the user, do not just make up a random URI to '
+        'pass. When asking the user for the uri, you should suggest the "Copy '
+        'DTD Uri to clipboard" action. Do . When reconnecting after losing a '
```
DanTup, line 618: I don't think "Do ." was intended to be left here.

jakemac53: whoops!

PR #180, opened 2025-06-18
make the log file test less flaky by retrying deletes and reads

jakemac53:
Seeing a lot of flakes on windows for this test, adding some retry logic for deleting the file and also reading from it should make it more reliable.

### DIFF
--- a/pkgs/dart_mcp_server/test/dart_tooling_mcp_server_test.dart
+++ b/pkgs/dart_mcp_server/test/dart_tooling_mcp_server_test.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:io';
 
 import 'package:test/test.dart';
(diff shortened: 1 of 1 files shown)


pkgs/dart_mcp_server/test/dart_tooling_mcp_server_test.dart
```diff
@@ -23,16 +24,43 @@ void main() {
...
 
     test('logs traffic to a file', () async {
-      expect(
-        await File(logDescriptor.io.path).readAsLines(),
-        containsAll([
-          allOf(startsWith('<<<'), contains('"method":"initialize"')),
-          allOf(startsWith('>>>'), contains('"serverInfo"')),
-          allOf(startsWith('<<<'), contains('"notifications/initialized"')),
-        ]),
+      // It may take a bit for all the lines to show up in the log.
```
kenzieschmoll, line 27: general question, are we writing to file with `flush: true`?

kenzieschmoll (approved):

pkgs/dart_mcp_server/test/dart_tooling_mcp_server_test.dart
jakemac53: We are calling flush on it after writing (we are opening the file and adding to the sink, then calling flush). In my experience on windows flush doesn't do anything though.

jakemac53: Also the API says `NOTE: This is not necessarily the same as the data being flushed by the operating system.`

PR #184, opened 2025-06-23
Add error checks for accessing Request members

gspencergoog:
## Description

This adds error checking and throws `ArgumentError` for `Request` subclasses which have required fields.  This is because in `shared.dart`, in `registerRequestHandler`, it casts the argument's value to a Map<String, Object?>, but it doesn't do any argument checking because creating those request objects don't go through the constructors, they are just cast:

```dart
    return impl((p.value as Map?)?.cast<String, Object?>() as T);
```

So I added error checking to all of the accessors for required members of `Request` objects, so that when they are accessed they will throw.

I ran across this when a client tried to send a malformed initialization request that didn't include capabilities, and the stack trace it sent was less than helpful.  This will at least send clearer errors.

## Tests
 - Added a test to check that these all throw.  Not sure how useful that is, really, but it at least keeps them from being removed accidentally.

### DIFF
--- a/pkgs/dart_mcp/CHANGELOG.md
+++ b/pkgs/dart_mcp/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Add `clientInfo` field to `MCPServer`, assigned during initialization.
 - Move the `done` future from the `ServerConnection` into `MCPBase` so it is
   available to the `MPCServer` class as well.
+- Added error checking to required fields of all `Request` subclasses so that
+  they will throw helpful errors when accessed and not set.
 
 ## 0.2.1
 
--- a/pkgs/dart_mcp/CHANGELOG.md
+++ b/pkgs/dart_mcp/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.2.3-wip
+
+- Added error checking to required fields of all `Request` subclasses so that
+  they will throw helpful errors when accessed and not set.
+
 ## 0.2.2
 
 - Refactor `ClientImplementation` and `ServerImplementation` to the shared
@@ -6,8 +11,6 @@
 - Add `clientInfo` field to `MCPServer`, assigned during initialization.
 - Move the `done` future from the `ServerConnection` into `MCPBase` so it is
   available to the `MPCServer` class as well.
-- Added error checking to required fields of all `Request` subclasses so that
-  they will throw helpful errors when accessed and not set.
 
 ## 0.2.1
 
--- a/pkgs/dart_mcp/lib/src/api/completions.dart
+++ b/pkgs/dart_mcp/lib/src/api/completions.dart
@@ -41,12 +41,11 @@ extension type CompleteRequest.fromMap(Map<String, Object?> _value)
 
   /// The argument's information.
   CompletionArgument get argument {
-    if (_value['argument'] == null) {
+    final argument = _value['argument'] as Map?;
+    if (argument == null) {
       throw ArgumentError('Missing argument field in $CompleteRequest.');
-    } else {
-      return (_value['argument'] as Map).cast<String, Object?>()
-          as CompletionArgument;
     }
+    return argument.cast<String, Object?>() as CompletionArgument;
   }
 }
 
--- a/pkgs/dart_mcp/lib/src/api/completions.dart
+++ b/pkgs/dart_mcp/lib/src/api/completions.dart
@@ -27,11 +27,21 @@ extension type CompleteRequest.fromMap(Map<String, Object?> _value)
   ///
   /// In the case of a [ResourceReference], it must refer to a
   /// [ResourceTemplate].
-  Reference get ref => _value['ref'] as Reference;
+  Reference get ref {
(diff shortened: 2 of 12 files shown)


pkgs/dart_mcp/lib/src/api/completions.dart
```diff
@@ -27,11 +27,21 @@ extension type CompleteRequest.fromMap(Map<String, Object?> _value)
   ///
   /// In the case of a [ResourceReference], it must refer to a
   /// [ResourceTemplate].
-  Reference get ref => _value['ref'] as Reference;
+  Reference get ref {
+    if (_value['ref'] == null) {
+      throw ArgumentError('Missing ref field in $CompleteRequest.');
+    }
+    return _value['ref'] as Reference;
```
jakemac53: nit: use if case or a switch for these to avoid the extra lookup:

```suggestion
    if (_value['ref'] case Reference ref) {
      return ref;
    } else {
      throw ArgumentError('Missing ref field in $CompleteRequest.');
    }
```

or:

```suggestion
  final ref = _value['ref'];
  return switch(ref) {
    case Reference() => ref;
    case null => throw ArgumentError('Missing ref field in $CompleteRequest.');
    _ => throw ArgumentError('Invalid ref field in $CompleteRequest, expected a Reference.');
  };
```

gspencergoog: Done. I did the switch for the ref case, but for the others, I used a final to avoid the extra lookup.

pkgs/dart_mcp/lib/src/api/completions.dart
```diff
@@ -28,19 +28,25 @@ extension type CompleteRequest.fromMap(Map<String, Object?> _value)
   /// In the case of a [ResourceReference], it must refer to a
   /// [ResourceTemplate].
   Reference get ref {
-    if (_value['ref'] == null) {
-      throw ArgumentError('Missing ref field in $CompleteRequest.');
-    }
-    return _value['ref'] as Reference;
+    final ref = _value['ref'];
```
jakemac53: Lets pick one style and stick with it, looks like 3 different styles here.

I don't have a strong opinion about whether we should both try and provide a nice error for null and for other invalid values, but it should be consistent.

gspencergoog: Sure.  Trying to make them all be the final var style, but there are some that are collections that need casting, or enums that need iterating to validate, so they vary a little.

pkgs/dart_mcp/lib/src/api/completions.dart
```diff
@@ -27,11 +27,22 @@ extension type CompleteRequest.fromMap(Map<String, Object?> _value)
...
 
   /// The argument's information.
-  CompletionArgument get argument =>
-      (_value['argument'] as Map).cast<String, Object?>() as CompletionArgument;
+  CompletionArgument get argument {
+    final argument = _value['argument'] as Map?;
+    if (argument == null) {
+      throw ArgumentError('Missing argument field in $CompleteRequest.');
+    }
+    return argument.cast<String, Object?>() as CompletionArgument;
```
jakemac53: I actually think we don't need the deep casting here, you can just do `as CompletionArgument?` above.

gspencergoog: Okay, removed.  I left the other deep casting on the lists (like the roots) because that is necessary.

pkgs/dart_mcp/lib/src/api/completions.dart
```diff
@@ -27,11 +27,22 @@ extension type CompleteRequest.fromMap(Map<String, Object?> _value)
   ///
   /// In the case of a [ResourceReference], it must refer to a
   /// [ResourceTemplate].
-  Reference get ref => _value['ref'] as Reference;
+  Reference get ref {
+    final ref = _value['ref'];
+    if (ref == null) {
+      throw ArgumentError('Missing ref field in $CompleteRequest.');
+    }
+    return ref as Reference;
```
jakemac53:
```suggestion
    final ref = _value['ref'] as Reference?;
    if (ref == null) {
      throw ArgumentError('Missing ref field in $CompleteRequest.');
    }
    return ref;
```

gspencergoog: Fixed.

jakemac53 (approved):

pkgs/dart_mcp/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 - Add `clientInfo` field to `MCPServer`, assigned during initialization.
 - Move the `done` future from the `ServerConnection` into `MCPBase` so it is
   available to the `MPCServer` class as well.
+- Added error checking to required fields of all `Request` subclasses so that
```
jakemac53: Oh actually this needs to move up into the next version `0.2.3-wip` and the pubspec probably needs to be bumped as well.

jakemac53 (approved):

PR #190, opened 2025-06-25
Add enum support to `Schema`

gspencergoog:
## Description

In preparation for adding Elicitation support, this add `enum` support to the Schema class.

Elicitations can request enum responses as part of their `requestedSchema`.

## Tests
 - Added validation tests.

### DIFF
--- a/pkgs/dart_mcp/lib/src/api/tools.dart
+++ b/pkgs/dart_mcp/lib/src/api/tools.dart
@@ -324,13 +324,6 @@ extension type ValidationError.fromMap(Map<String, Object?> _value) {
 /// if you need something more complex you can create your own
 /// `Map<String, Object?>` and cast it to [Schema] (or [ObjectSchema]) directly.
 extension type Schema.fromMap(Map<String, Object?> _value) {
-  factory Schema({JsonType? type, String? title, String? description}) =>
-      Schema.fromMap({
-        'type': JsonType.enumeration.typeName,
-        if (title != null) 'title': title,
-        if (description != null) 'description': description,
-      });
-
   /// A combined schema, see
   /// https://json-schema.org/understanding-json-schema/reference/combining#schema-composition
   factory Schema.combined({


pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -1081,6 +1094,56 @@ extension type const StringSchema.fromMap(Map<String, Object?> _value)
...
+  });
+
+  /// A title for this schema, should be short.
+  String? get title => _value['title'] as String?;
+
+  /// A description of this schema.
+  String? get description => _value['description'] as String?;
+
+  /// The allowed enum values.
+  Set<String> get values => (_value['enum'] as Set).cast<String>();
```
jakemac53: Should we handle null values better here similar to the previous PR?

jakemac53: This is typically going to be a `List`, not a `Set`. We could call `toSet`, but I would probably just make it have an `Iterable` type so that it can be either? We should also validate that the JSON encoder can handle any kind of Iterable.

jakemac53 (commented):
We should also add a custom "constructor" (actually static method) to the Schema class (see the other ones there)

pkgs/dart_mcp/lib/src/api/tools.dart
gspencergoog: Yeah, I went back and forth on that too.  Okay, I'll just make it an `Iterable`.

gspencergoog: I left out the .toSet, but I did add an assert to check for duplicate entries, which is why I wanted to use .toSet in the first place.

gspencergoog: Yes, we should.

jakemac53 (approved):

pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -300,6 +324,13 @@ extension type ValidationError.fromMap(Map<String, Object?> _value) {
 /// if you need something more complex you can create your own
 /// `Map<String, Object?>` and cast it to [Schema] (or [ObjectSchema]) directly.
 extension type Schema.fromMap(Map<String, Object?> _value) {
+  factory Schema({JsonType? type, String? title, String? description}) =>
```
jakemac53: See line 355 for what I meant - I don't think this unnamed constructor makes sense

jakemac53 (changes requested):
See latest comment, one last thing to do actually before landing

pkgs/dart_mcp/lib/src/api/tools.dart
gspencergoog: Did you mean this one?:

```
  /// Alias for [EnumSchema.new].
  static const enumeration = EnumSchema.new;
```

That was already there, which is why I got confused...

jakemac53: Yeah that is what I meant, I guess I just missed that

Issue #172, opened 2025-06-17
Add a test to ensure tool list in README is up to date


kenzieschmoll:
This test should parse the content of the `dart_mcp_server` package `README` file and verify that the list of supported tools documented in the README contains all tools provided by the Dart MCP Server. This test should run on presubmits and will prevent the readme from becoming stale.

Another option would be to generate the README based on the list of provided tools and their descriptions, but this may be more work to implement.

PR #192, opened 2025-06-25
add a CI check that the readme file is up-to-date

devoncarew:
- add a CI check that the readme file is up-to-date
- address https://github.com/dart-lang/ai/issues/172

---

<details>

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>


pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
```diff
@@ -179,6 +179,13 @@ base mixin DashCliSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
     );
   }
 
+  static final List<Tool> allTools = [
```
kenzieschmoll: nit: consider looping through this list in the initialize method so that it is less likely a new tool is added to initialize and not added here. You'd have to store that tools with their associated callback here.

pkgs/dart_mcp_server/tool/update_readme.dart
```diff
@@ -0,0 +1,54 @@
...
+
+void main(List<String> args) {
+  final buf = StringBuffer();
+
+  for (final entry in toolCategories.entries) {
+    final category = entry.key;
+    final tools = entry.value;
+
+    buf.writeln('### $category');
+    buf.writeln('');
```
kenzieschmoll: nit: use cascades

jakemac53:
> Hmm, it looks like `dart pub get` for this package takes `1m 52s`. I expect that's due to the git dep on the dart-lang/sdk repo (to get the language_server_protocol package).

Yeah it is, no good way to avoid it that I am aware of. We could depend on a specific hash and that might get us a cache hit (if we also cache the pub cache on GitHub actions), so it would speed up CI a fair bit.

.github/workflows/dart_mcp_server.yaml
```diff
@@ -55,3 +55,16 @@ jobs:
...
+
+  readme:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      - uses: dart-lang/setup-dart@e51d8e571e22473a2ddebf0ef8a2123f0ab2c02c
+
+      - run: dart pub get
```
jakemac53: You could just merge this into the job above to avoid the extra setup work 🤷‍♂️ . This repo isn't large enough to really need multiple jobs.

devoncarew: I kept this separate since this above job was doing more than we need here (using the flutter sdk instead of the dart sdk, ...).

pkgs/dart_mcp_server/tool/update_readme.dart
devoncarew: (this bit of code has since been removed)

pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
devoncarew: > it is less likely a new tool is added to initialize and not added here.

Yeah, I was a little worried about the list of tools getting out of date. I re-wrote this PR to instead spin up the MCP server and ask it for the list of tools; this should be much less maintenance than keeping separate lists of tools in the src code.

devoncarew (commented):
ptal - updated to query an MCP server instance for the list of tools (instead of maintaining lists in the source code)

pkgs/dart_mcp_server/tool/update_readme.dart
```diff
@@ -4,51 +4,63 @@
...
   final startIndex = original.indexOf(marker) + marker.length;
   final endIndex = original.lastIndexOf(marker);
 
   return '${original.substring(0, startIndex)}\n\n'
-      '$insertion${original.substring(endIndex)}';
+      '$insertion\n${original.substring(endIndex)}';
+}
+
+Future<List<Tool>> _retrieveRegisteredTools() async {
+  final client = MCPClient(
```
jakemac53, line 44: Fwiw, the tools registered can vary based on the client features that are supported.

Probably, this client should claim to support all features so that all the tools show up.

.github/workflows/dart_mcp_server.yaml
jakemac53: Ultimately its going to add a lot of time to the CI run though due to the pub get.

kenzieschmoll:
@devoncarew Is this PR still on your radar? There are a couple of missing tools in the table that your PR adds, so we can either document them manually or wait for this PR to land.

devoncarew:
> @devoncarew Is this PR still on your radar? There are a couple of missing tools in the table that your PR adds, so we can either document them manually or wait for this PR to land.

Yup - I'll circle back to the comments here and update the PR.

pkgs/dart_mcp_server/tool/update_readme.dart
devoncarew: I looked at creating a subclass of an `MCPClient` in order to override `initialize()` so I could create specific client capabilities, but for whatever reason, two of the tools dropped out of the list of supported tools. For now, I kept the implementation as is - not providing an explcit list of capabilities.

.github/workflows/dart_mcp_server.yaml
devoncarew: Gotcha - I merged this into the job above.

devoncarew:
> > What is the status of moving this into the SDK repo? Paused, likely will move, likely won't move? If this is likely not going to move to the SDK repo, we could look to publish package:language_server_protocol. I expect it doesn't have any real deps on the sdk repo and seems fine as a tools package.
> 
> I would say its not likely to happen any time soon, if ever. Working outside the SDK is far more productive and flexible.

In that case, I think looking at publishing `language_server_protocol` (or similar solution) would be worthwhile - taking a git dep on a package from the sdk repo has downsides.

pkgs/dart_mcp_server/tool/update_readme.dart
jakemac53: What likely happened there is we swapped out some tools for their preferred alternatives (resources), due to the client indicating it has resource support.

We can land this as is though and I can eventually take a look at improving it.

DanTup:
> cc @DanTup wrt considering publishing `package:language_server_protocol`; I'm not sure how much time is spent maintaining that package, or, how coupled it is in terms of it needing to be developed in the sdk repo. But something to consider.

I don't think there's much effort in maintaining that package. It changes infrequently besides when LSP updates, and the code is all generated. Given we're exposing some of the LSP functionality over DTD, I think publishing it makes sense.

There are some things in there I don't know if make sense in a public package though, such as:

- the classes for our custom LSP requests (for example `EditableArguments`) - these are useful (eg. DevTools calls these), but I don't know if they belong in the same package
- we use custom `EitherX<>` classes to represent the unions in TypeScript/LSP.. these feel like strange classes to get from an LSP package, but I don't know if there's an alternative (without Dart having native support, or a 1p package for them)
- there are some other classes in here like `extension IncomingMessageExtension` that look server-specific that should be moved out
- it also seems slightly odd that we inject the MS copyright statement in the top of the generated files, making it seem like the source code came from MS, when it was written by use and just derived from their spec (moreso for
`lib\protocol_custom_generated.dart` since we generate that from our own definitions, I guess it's just because we're using the same generation code)

Perhaps we should do a little clean up and publish it as a `0.x`?

(cc @bwilkerson @jwren)

Issue #197, opened 2025-06-26
Errors when calling tools from Gemini in Android Studio


kenzieschmoll:
<img width="512" alt="Image" src="https://github.com/user-attachments/assets/d7da9609-3451-4f56-906a-7e1c4bdd03e1" />

<img width="638" alt="Image" src="https://github.com/user-attachments/assets/fd472a58-8b09-4490-a488-c69c8a7e94ed" />

@jakemac53 is this a bug with our tools or a bug with how GiAS is trying to call our tools? I set up our server in GiAS following these instructions https://developer.android.com/studio/preview/gemini/agent-mode#add-mcp. Right now this is only available on Android Studio Preview

kenzieschmoll:
I filed https://b.corp.google.com/issues/428017715 to track on the Android Studio side.

jakemac53:
We should be providing better errors now which should help - but in any case I am going to close the issue here as it is not an issue with the MCP server.

PR #200, opened 2025-06-26
auto validate tool arguments, improve validation messages

jakemac53:
Marking this as closing https://github.com/dart-lang/ai/issues/197 - it should significantly help the LLM figure out what it did wrong. The error would now be something like "Value `<root>` is not of type `List<dynamic>` at path #root["roots"]".

I made a fair number of changes to the validation errors here, namely only emitting errors for the leaf node where the error actually happened and not the entire path up. I don't think that extra info was helpful.

cc @kenzieschmoll


pkgs/dart_mcp/lib/src/api/tools.dart
```diff
@@ -279,29 +298,38 @@ enum ValidationErrorType {
...
 
   /// Additional details about the error (optional).
   String? get details => _value['details'] as String?;
 
   String toErrorString() {
-    return '${error!.name} in object at '
-        '${path!.map((p) => '[$p]').join('')}'
-        '${details != null ? ' - $details' : ''}';
+    return '${details != null ? '$details' : error.name} at path '
+        '#root${path.map((p) => '["$p"]').join('')}'
```
gspencergoog, line 345: Will the LLM understand what `#root` stands for? Would `[ROOT]` or `$root` be clearer? (or something else?)

pkgs/dart_mcp/test/api/tools_test.dart
```diff
@@ -960,7 +978,14 @@ void main() {
           not: [StringSchema(maxLength: 10), StringSchema(pattern: 'test')],
         );
         expectFailuresMatch(schema, 'test', [
-          ValidationError(ValidationErrorType.notConditionViolated),
+          ValidationError(
```
gspencergoog, line 985: This probably should be de-duped.

pkgs/dart_mcp/lib/src/api/tools.dart
jakemac53: The reason I did this, is when asking Gemini to write tests, it used a syntax sort of like this (actually, it was `#/<path>`).

I think it probably will understand this, but it's hard to say what the best thing to do here is. We could also just emit an empty array in the case of no path and assume the path is enough when there is one.

pkgs/dart_mcp/test/api/tools_test.dart
jakemac53: Actually, each of these is a separate error, we emit one for each condition in the not that did match. They have different detail messages in reality, you just can't see that in this test.

pkgs/dart_mcp/lib/src/api/tools.dart
gspencergoog: Maybe this is something we just need to try for a while and see what works best.

Issue #188, opened 2025-06-25
add docs for how to use with Gemini CLI


devoncarew:
We have docs for how to use this w/ Cursor; we should add docs for how to use with Gemini CLI.

PR #201, opened 2025-06-27
Add documentation for setting up the MCP server in popular clients

kenzieschmoll:
This PR adds docs for the Gemini CLI, Gemini Code Assist, and Copilot.

Fixes https://github.com/dart-lang/ai/issues/188


pkgs/dart_mcp_server/README.md
```diff
@@ -1,11 +1,139 @@
...
+file in your workspace:
+
+```json
+"servers": {
+  "dart": {
+    "type": "stdio",
+    "command": "dart",
+    "args": [
+      "mcp-server",
+      "--experimental-mcp-server", // Can be removed for Dart 3.9.0 or later
```
jakemac53, line 136: Do we know if we should pass `--force-roots-fallback` for this one?

jakemac53 (approved):

pkgs/dart_mcp_server/README.md
```diff
@@ -1,143 +1,166 @@
...
+```
 
-The following button should work for most users:
+For more information, see the official Cursor documentation for
+[installing MCP servers](https://docs.cursor.com/context/model-context-protocol#installing-mcp-servers).
 
-[![Add to Cursor](https://cursor.com/deeplink/mcp-install-dark.svg)](https://cursor.com/install-mcp?name=dart_tooling&config=eyJ0eXBlIjoic3RkaW8iLCJjb21tYW5kIjoiZGFydCBtY3Atc2VydmVyIC0tZXhwZXJpbWVudGFsLW1jcC1zZXJ2ZXIgLS1mb3JjZS1yb290cy1mYWxsYmFjayJ9)
+### Visual Studio Code Copilot
+<!-- TODO: update these docs to mention the automatic configuration provided by the Dart
+VS Code extension. -->
```
kenzieschmoll: @dantup FYI

pkgs/dart_mcp_server/README.md
kenzieschmoll: @dantup did you need to add this flag to get things to work with your local testing?

jakemac53:
See https://code.visualstudio.com/docs/copilot/guides/mcp-developer-guide#_mcp-url-handler also - it looks like VsCode itself has some similar deep linking functionality to cursor.

pkgs/dart_mcp_server/README.md
jakemac53: https://code.visualstudio.com/docs/copilot/guides/mcp-developer-guide#_workspace-roots it claims to provide the workspace root

DanTup: I have not been passing this flag (I didn't actually know about it 🙂)

DanTup:
> See https://code.visualstudio.com/docs/copilot/guides/mcp-developer-guide#_mcp-url-handler also - it looks like VsCode itself has some similar deep linking functionality to cursor.

I think it would be better to just have people use the flag to enable it via Dart-Code, because that way we can control the flags being passed (particularly if they change over time), and it also enables the DTD URI tool (although in theory, we could register that tool regardless of the flag).

pkgs/dart_mcp_server/README.md
jakemac53: It is just there for clients that advertise roots support but never set roots. We will automatically fall back if they don't advertise roots support.

kenzieschmoll:
>I think it would be better to just have people use the flag to enable it via Dart-Code, because that way we can control the flags being passed (particularly if they change over time), and it also enables the DTD URI tool (although in theory, we could register that tool regardless of the flag).

1. Is the Dart-Code flag ready to document here or should we wait until this goes out in a Dart-Code release? We can always land this now and update the docs once that functionality is ready.
2. I think we should register the tool to provide the DTD URI regardless of the flag. This is helpful for any VS code based IDE (like cursor for example) where the Dart Code extension will be installed.

DanTup:
> Is the Dart-Code flag ready to document here or should we wait until this goes out in a Dart-Code release? We can always land this now and update the docs once that functionality is ready.

I'm out today but will update it on Monday, so it'll be included in the release (which will be Monday or Tuesday).

I suggest we name it `dart.mcpServer` as a boolean, so enabling it would be adding `"dart.mcpServer": true,` to your User or Workspace settings (or searching for "Dart MCP Server" in the settings UI). I think it's reasonable to just add it here now, otherwise I can send a PR after it's released.


> I think we should register the tool to provide the DTD URI regardless of the flag. This is helpful for any VS code based IDE (like cursor for example) where the Dart Code extension will be installed.

I'm fine doing this too. Originally I wanted everything behind the flag so it could be more easily tested (and not enabled until it's hard more testing), but it's such a small function I don't think there's much to go wrong (and if it comes to it, we can always push a patch release out that changes it).

pkgs/dart_mcp_server/README.md
```diff
@@ -100,11 +100,20 @@ For more information, see the official Cursor documentation for
 [installing MCP servers](https://docs.cursor.com/context/model-context-protocol#installing-mcp-servers).
 
 ### Visual Studio Code Copilot
-<!-- TODO: update these docs to mention the automatic configuration provided by the Dart
-VS Code extension. -->
 
-To configure the Dart MCP server for a single project, edit the `.vscode/mcp.json`
-file in your workspace:
+To configure the Dart MCP server for a single project, edit your VS Code user settings to
```
DanTup: nit: Adding it to your User Settings will enable it for all workspaces. If you add it to _Workspace Settings_ then it will be enabled only for that workspace. "project" is probably the wrong term to use here because it's always enabled (or not) for an open workspace (which may be a single folder) and is never specifically scoped to a single project (unless of course your workspace just happens to be a single project).

pkgs/dart_mcp_server/README.md
```diff
@@ -100,11 +100,20 @@ For more information, see the official Cursor documentation for
...
+"dart.mcpServer": true
+```
+
+By adding this setting, the Dart VS Code extension will register the Dart MCP Server
+configuration with VS Code so that you don't have to manually configure the server.
+Copilot or other AI agents that support the
+[VS Code MCP API](https://code.visualstudio.com/api/extension-guides/mcp) will automatically
+configure the Dart MCP server on your behalf.
+
+Optionally, you can also manually edit the `.vscode/mcp.json` file in your workspace:
```
DanTup: I would probably change "Optionally" to "Alternatively", so it's clearer that this doesn't do anything extra if you've already done the above.

pkgs/dart_mcp_server/README.md
```diff
@@ -119,7 +128,7 @@ file in your workspace:
 }
 ```
 
-To make the Dart MCP server available in every project you open, edit your
+Or, to make the Dart MCP server available in every project you open, edit your
```
DanTup: Is there any value in this section? I think it's a bit confusing because it's somewhat the same as the top option (but probably not the way we'd want people to do it).

kenzieschmoll: added a note to the first option to make it clear that it is a global setting, and a note to the second option to make it clear that it is a workspace setting. And then I removed this section.

pkgs/dart_mcp_server/README.md
```diff
@@ -1,142 +1,167 @@
...
+
+[![Add to VS Code](_docs/add_to_vs_code.png)](https://vscode.dev/redirect?url=vscode:mcp/install?%7B%22name%22%3A%22dart%22%2C%22command%22%3A%22dart%22%2C%22args%22%3A%5B%22mcp-server%22%2C%22--experimental-mcp%22%5D%7D)
 
-## Development
+Or, you can manually edit the `.vscode/mcp.json` file in your workspace,
+which will only configure the Dart MCP server for the local workspace:
 
-For local development, use the [MCP Inspector](https://modelcontextprotocol.io/docs/tools/inspector).
+```json
+"servers": {
```
DanTup, line 130: I think for Workspace usage, we should still encouarge using the `"dart.mcpServer"` setting, but instead just adding it to Workspace Settings (`.vscode/settings.json`) instead of User settings. That way we have the same benefits of controlling all the flags/conditions.

If we encourage using `mcp.json` and/or the button, users might end up with multiple registrations if they change the User Setting in future (or we change the default).

Issue #195, opened 2025-06-26
Add environment support to client


CalsRanna:
This is the code in `client.dart`:

```dart
Future<ServerConnection> connectStdioServer(
    String command,
    List<String> arguments, {
    Sink<String>? protocolLogSink,
  }) async {
    final process = await Process.start(command, arguments);
    process.stderr
        .transform(utf8.decoder)
        .transform(const LineSplitter())
        .listen((line) {
          stderr.writeln('[StdErr from server $command]: $line');
        });
    final channel = StreamChannel.withCloseGuarantee(
          process.stdout,
          process.stdin,
        )
        .transform(StreamChannelTransformer.fromCodec(utf8))
        .transformStream(const LineSplitter())
        .transformSink(
          StreamSinkTransformer.fromHandlers(
            handleData: (data, sink) {
              sink.add('$data\n');
            },
          ),
        );
    final connection = connectServer(channel, protocolLogSink: protocolLogSink);
    unawaited(connection.done.then((_) => process.kill()));
    return connection;
  }
```

As we can see, the `Process.start` supports`Map<String, String>? environment`, but MCPClient DOES NOT expose it. So why the environment is not supported in Client?

And the temp way to solve this is you define a new function in extension on MCPClient, and copy the code up here and add environment support yourself with importing `async` and `stream channel` packages.

jakemac53:
Sure, this seems like something we can add. I kind of wish we had just taken a `Process` as an argument so we didn't have to duplicate _all_ of these parameters, but it's probably not worth doing a breaking change for that.

jakemac53:
We would accept a PR for this if you want to provide one, otherwise we can probably get to it in a little bit.

jakemac53:
Actually, we are doing a breaking change release anyways so I think I am going to roll this into that, and change `connectStdioServer` to take a stdin sink and stdout stream, which will allow us to drop the `dart:io` dependency as well, making this package runnable on the web.

PR #204, opened 2025-06-27
update MCPClient.connectStdioServer to operate on streams and sinks

jakemac53:
Updated MCPClient.connectStdioServer to take a stdin and stdout stream instead of a binary name and arguments. It no longer spawns processes on its own. This removes the `dart:io` dependency and allows more flexibility for how processes are spawned (can use package:process or even no process at all).

The migration is pretty easy, just launch your own process and pass in stdin/stdout from that.

- Closes https://github.com/dart-lang/ai/issues/195
- Enables the client to run on the web
- Run all tests on the web, with wasm and dart2js, as well as AOT and kernel (these tests are all very fast)


gspencergoog (approved):
Oh, yes, this is much better.

PR #206, opened 2025-06-27
Remove VS Code mcp instructions in favor of Dart-Code setting.

kenzieschmoll:
@dantup this addresses your comments on https://github.com/dart-lang/ai/pull/201#discussion_r2172757237. Please let me know if there are other changes you'd like me to make to the docs.

### DIFF
--- a/pkgs/dart_mcp_server/README.md
+++ b/pkgs/dart_mcp_server/README.md
@@ -117,7 +117,7 @@ By adding this setting, the Dart VS Code extension will register the Dart MCP Se
 configuration with VS Code so that you don't have to manually configure the server.
 Copilot will then automatically configure the Dart MCP server on your behalf. This is
 a global setting. If you'd only like the setting to apply to a specific workspace, add
-the entry to your workspace settings file: `.vscode/settings.json`.
+the entry to your workspace settings (Command Palette > **Preferences: Open Workspace Settings**).
 
 For more information, see the official VS Code documentation for
 [enabling MCP support](https://code.visualstudio.com/docs/copilot/chat/mcp-servers#_enable-mcp-support-in-vs-code).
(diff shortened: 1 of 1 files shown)


pkgs/dart_mcp_server/README.md
```diff
@@ -116,28 +116,8 @@ to your VS Code user settings:
...
-    "command": "dart",
-    "args": [
-      "mcp-server",
-      "--experimental-mcp-server", // Can be removed for Dart 3.9.0 or later
-    ]
-  }
-}
-```
+a global setting. If you'd only like the setting to apply to a specific workspace, add
+the entry to your workspace settings file: `.vscode/settings.json`.
```
DanTup: I would probably just write "your workspace settings" because these are not always in `.vscode/settings.json` (for example in the case where the user has a `.code-workspace` file, they would go in there).

I would generally encourage people to use the **Preferences: Open Workspace Settings** command, because it will always go to the right place.

DanTup (approved):
LGTM, though one minor suggestion.

jakemac53 (approved):

kenzieschmoll:
@jakemac53 added a note that these setup instructions require Dart 3.9.0 or later (current beta). LMK if you think I should tweak the version number to be more precise.

@DanTup similarly, I added a note that the VS Code setting requires Dart-Code v3.114 or later. Can you confirm this is the correct version to list here?

DanTup:
@kenzieschmoll 

> @DanTup similarly, I added a note that the VS Code setting requires Dart-Code v3.114 or later. Can you confirm this is the correct version to list here?

Yep, that's correct - that version will be released tomorrow and includes all the things we've previously discussed. Other changes LGTM too :)

Issue #198, opened 2025-06-26
Get VM service URIs from the `ConnectedAppService` on DTD instead of the `EditorService`


kenzieschmoll:
Right now, we get the list of debug sessions from the DTD `EditorService` in order to get access to the running app's VM service URI. Now that the `ConnectedAppService` exists, we should call `ConnectedAppService.getVmServiceUris` when the `EditorService.getDebugSessions` method is not available (the `EditorService.getDebugSessions` method is only available from VS Code right now). Maintaining both paths will ensure that the Dart Tooling MCP Server is backwards compatible with older IDE plugin / extension versions. After some safe time period (12-18 months?), switch over to using `ConnectedAppService.getVmServiceUris` exclusively.

Tracking issue for creating the `ConnectedAppService`: https://github.com/dart-lang/sdk/issues/60540
Tracking issue for integrating this service with the Dart / Flutter IntelliJ plugin: https://github.com/flutter/flutter-intellij/issues/8261.

jakemac53:
What is the best way for us to detect whether or not the connected app service is registered and also active?

Should we just listen on both services for now, and ignore VM service URIs that we have already seen from one or the other?

kenzieschmoll:
You can call the `DartToolingDaemon.getRegisteredServices` method to get the list of registered services: https://github.com/dart-lang/sdk/blob/main/pkg/dtd/lib/src/dart_tooling_daemon.dart#L120. For backwards compatibility we may want to call this with a try catch, since the first version of Dart this made it into was 3.9.0-163.0.dev.

We can also listen to the stream to detect when the service is registered. However, this will have the same version constraints since https://github.com/dart-lang/sdk/commit/10dd59c03961d3d9fea84c5ea87359bce334ca70#diff-d65aaf2c962c8469a0a1c72dcdaae8d4a262f8c44a094c7f5c757d2fa2fc9d8b also fixed a bug where DTD internal services weren't sent over this stream, and the `getRegisteredServices` method is an internal service.

kenzieschmoll:
[...]

PR #208, opened 2025-06-30
Migrate to the connected app service

jakemac53:
Closes https://github.com/dart-lang/ai/issues/198

This does drop all support for connecting to apps on SDKs that don't have this service, and I bumped the min SDK accordingly.

### DIFF
--- a/pkgs/dart_mcp_server/CHANGELOG.md
+++ b/pkgs/dart_mcp_server/CHANGELOG.md
@@ -42,3 +42,5 @@
 * Add `--log-file` argument to log all protocol traffic to a file.
 * Improve error text for failed DTD connections as well as the tool description.
 * Add support for injecting an `Analytics` instance to track usage.
+* Listen to the new ConnectedApp service instead of the Editor.DebugSessions
+  service, when available.
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -8,7 +8,6 @@ import 'dart:convert';
 import 'package:dart_mcp/server.dart';
 import 'package:dds_service_extensions/dds_service_extensions.dart';
 import 'package:dtd/dtd.dart';
-import 'package:json_rpc_2/json_rpc_2.dart';
 import 'package:meta/meta.dart';
 import 'package:unified_analytics/unified_analytics.dart' as ua;
 import 'package:vm_service/vm_service.dart';
@@ -28,20 +27,20 @@ base mixin DartToolingDaemonSupport
     implements AnalyticsSupport {
   DartToolingDaemon? _dtd;
 
-  /// Whether or not the DTD extension to get the active debug sessions is
-  /// ready to be invoked.
-  bool _getDebugSessionsReady = false;
-
   /// The last reported active location from the editor.
   Map<String, Object?>? _activeLocation;
 
-  /// A Map of [VmService] object [Future]s by their associated
-  /// [DebugSession.id].
+  /// A Map of [VmService] object [Future]s by their VM Service URI.
   ///
-  /// [VmService] objects are automatically removed from the Map when the
-  /// [DebugSession] shuts down.
+  /// [VmService] objects are automatically removed from the Map when they
+  /// are unregistered via DTD.
   @visibleForTesting
   final activeVmServices = <String, Future<VmService>>{};
+
+  /// Whether or not the connected app service is supported.
+  ///
+  /// Once we connect to dtd, this may be toggled to `true`.
+  bool _connectedAppServiceIsSupported = false;
 
   /// Whether to await the disposal of all [VmService] objects in
   /// [activeVmServices] upon server shutdown or loss of DTD connection.
@@ -64,8 +63,8 @@ base mixin DartToolingDaemonSupport
   /// Called when the DTD connection is lost, resets all associated state.
   Future<void> _resetDtd() async {
     _dtd = null;
-    _getDebugSessionsReady = false;
     _activeLocation = null;
+    _connectedAppServiceIsSupported = false;
 
     // TODO: determine whether we need to dispose the [inspectorObjectGroup] on
     // the Flutter Widget Inspector for each VM service instance.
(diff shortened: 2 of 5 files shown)


pkgs/dart_mcp_server/CHANGELOG.md
```diff
@@ -42,3 +42,5 @@
 * Add `--log-file` argument to log all protocol traffic to a file.
 * Improve error text for failed DTD connections as well as the tool description.
 * Add support for injecting an `Analytics` instance to track usage.
+* Listen to the new ConnectedApp service instead of the Editor.DebugSessions
```
kenzieschmoll:
```suggestion
* Listen to the new DTD `ConnectedApp` service instead of the `Editor.DebugSessions`
```

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -28,21 +27,21 @@ base mixin DartToolingDaemonSupport
...
   Map<String, Object?>? _activeLocation;
 
-  /// A Map of [VmService] object [Future]s by their associated
-  /// [DebugSession.id].
+  /// A Map of [VmService] object [Future]s by their VM Service URI.
   ///
-  /// [VmService] objects are automatically removed from the Map when the
-  /// [DebugSession] shuts down.
+  /// [VmService] objects are automatically removed from the Map when they
+  /// are unregistered via DTD.
```
kenzieschmoll:
```suggestion
  /// are unregistered via DTD or when the VM service shuts down.
```

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -214,51 +203,52 @@ base mixin DartToolingDaemonSupport
...
-              'Editor.getDebugSessions unregistered, dtd is no longer ready',
-            );
-            _getDebugSessionsReady = false;
-          }
+
+    _connectedAppServiceIsSupported = false;
+    try {
+      final registeredServices = await dtd.getRegisteredServices();
+      if (registeredServices.dtdServices.contains(
+        'ConnectedApp.getVmServices',
```
kenzieschmoll: nit: instead of raw strings can you the constants exposed by package:dtd? here and elsewhere below where we are using raw strings

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -214,51 +203,52 @@ base mixin DartToolingDaemonSupport
...
+    final dtd = _dtd!;
+    dtd.onEvent('ConnectedApp').listen((e) async {
       log(LoggingLevel.debug, e.toString());
       switch (e.kind) {
-        case 'debugSessionStarted':
-        case 'debugSessionChanged':
+        case 'VmServiceRegistered':
           await updateActiveVmServices();
-        case 'debugSessionStopped':
+        case 'vmServiceUnregistered':
```
kenzieschmoll: this should be VmServiceUnregistered, but if you use the constants from package:dtd that will always be accurate

pkgs/dart_mcp_server/test/test_harness.dart
```diff
@@ -384,8 +357,32 @@ Future<String> _getDTDUri(TestProcess dtdProcess) async {
   return dtdUri;
 }
 
-typedef ServerConnectionPair =
-    ({ServerConnection serverConnection, DartMCPServer? server});
+/// Reads DTD uri from the [dtdProcess] output.
+Future<String> _getDTDSecret(TestProcess dtdProcess) async {
```
kenzieschmoll: If you start dtd with the --machine flag you can parse the output as json instead of reading stdout lines. Example: https://github.com/flutter/devtools/blob/master/packages/devtools_shared/test/helpers/helpers.dart/#L18. Then you could possibly merge this method with the one above to return a DTD info object as a pattern

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: done, for the non-editor based ones (I don't see those being exposed as they aren't core services). Also migrated to `onVmServiceUpdate()`.

pkgs/dart_mcp_server/test/test_harness.dart
jakemac53: nice

kenzieschmoll (approved):

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -217,51 +208,53 @@ base mixin DartToolingDaemonSupport
...
-    dtd.onEvent('Editor').listen((e) async {
+    if (_connectedAppServiceIsSupported) {
+      await _listenForConnectedAppServiceEvents();
+    }
+    await _listenForEditorEvents();
+  }
+
+  Future<void> _listenForConnectedAppServiceEvents() async {
+    final dtd = _dtd!;
+    dtd.onVmServiceUpdate().listen((e) async {
```
kenzieschmoll, line 237: do we need to cancel these listeners in some dispose method?

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -807,10 +796,10 @@ class _AppErrorsListener {
   final StreamController<String> _errorsController;
 
   /// The listener for Flutter.Error vm service extension events.
-  final StreamSubscription<Event> _extensionEventsListener;
+  final StreamSubscription<Event>? _extensionEventsListener;
```
kenzieschmoll, line 799: where are these stream subscriptions set?

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: We could, but it's not necessary. We keep listening as long as the DTD instance itself is still alive. If that does shut down, this stream will be closed.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: Just normally via the constructor - they aren't late, just nullable. They are created as a part of the `forVmService` static method which passes them into the constructor.

PR #210, opened 2025-06-30
disable the screenshots tool by default

jakemac53:
Disabling due to https://github.com/flutter/flutter/issues/170357 - this only works today by default on the desktop environment.

We could fall back on full device screenshots but are holding off on that for now as we have to evaluate some extra potential privacy concerns with exposing the entire screen as opposed to just the flutter drawn pixels.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -61,6 +61,9 @@ base mixin DartToolingDaemonSupport
   /// for full list of available Flutter Widget Inspector service extensions.
   static const _inspectorServiceExtensionPrefix = 'ext.flutter.inspector';
 
+  /// Whether or not to enable the screenshot tool.
+  bool get enableScreenshots;
+
   /// Called when the DTD connection is lost, resets all associated state.
   Future<void> _resetDtd() async {
     _dtd = null;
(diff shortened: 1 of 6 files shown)


pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -61,6 +61,9 @@ base mixin DartToolingDaemonSupport
   /// for full list of available Flutter Widget Inspector service extensions.
   static const _inspectorServiceExtensionPrefix = 'ext.flutter.inspector';
 
+  /// Whether or not to enable the screenshot tool.
+  bool get enableScreenshots;
```
kenzieschmoll, line 65: should this default to false instead of being an abstract getter?

pkgs/dart_mcp_server/lib/src/server.dart
```diff
@@ -52,6 +52,7 @@ final class DartMCPServer extends MCPServer
     @visibleForTesting this.processManager = const LocalProcessManager(),
     @visibleForTesting this.fileSystem = const LocalFileSystem(),
     this.forceRootsFallback = false,
+    this.enableScreenshots = false,
```
kenzieschmoll, line 56: perhaps link to the bug to explain why this is defaulted to false

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: This is overridden always so the default would be meaningless here

Issue #222, opened 2025-07-10
[feat] enable dart_mcp_server to drive UI and test screenshots


csells:
to build a UI interactively from a design, an AI agent needs an evaluation function. in this case, the standard MCP server for driving UIs on the web won't work, since Puppeteer requires the HTML DOM, which Flutter Web doesn't provider. we really need the dart_mcp_server to allow the AI agent to drive a Flutter UI such that it can press buttons, enter text, scroll, etc. With that plus screenshots, the AI agent has everything it would need to allow Flutter devs to build Flutter UIs like the AI agents build and iterate on web-based UIs today.

jakemac53:
See https://github.com/flutter/flutter/issues/170357 for the issue that would unblock screenshots.

jakemac53:
For flutter driver the biggest issue I was running into is getting it to reliably figure out how to select the right thing on the screen. Likely we need some prompting work here, but I don't know enough about flutter driver itself to know the best course of action to suggest in the prompt.

Usually, I find they like to guess at tooltip or text content, and then eventually end up getting the widget tree when that fails. Even after getting the widget tree they sometimes have trouble clicking specific buttons in menu bars.

The other issue I was having is that Gemini 2.5 pro is just too slow based on the number of function calls required to do things, I have a lot better experience with faster but less intelligent models. They might fumble around a bit more but ultimately complete the task much more quickly.

jakemac53:
Yes I agree figuring out the right prompting (via descriptions) is going to improve this a lot. It's figuring out the right prompting that is hard, when I have never used flutter driver myself or have any context into the best way to use it 🤣.

I will try and find somebody with some domain knowledge though to figure that out.

jakemac53:
Fwiw, I played with this a whole bunch last week, trimmed down the set of available tools some and provided some more instruction in the tool itself, and in general things are working quite reliably. I even managed to live demo it without issues.

I have added a prompt to the MCP server as well, which is what I demo'd. You give it a user journey, and it will try to accomplish that user journey using flutter_driver. Once it has done so, it will write a flutter driver test for you based on the necessary actions it performed, and then also run that test to make sure it works. It works like a charm on simple-ish demos and I am excited to get some real world feedback on it once it hits beta!

I am also going to open up discussions on this repo (well, somebody with access is going to), and then I will convert this issue to a discussion.

PR #223, opened 2025-07-14
Add flutter driver tool

jakemac53:
Towards https://github.com/dart-lang/ai/issues/222 (could close it, but probably we should do some more work to validate it works well).

Adds the flutter driver tool, with a curated set of available commands. I decided to expose this as just a single tool, which mirrors exactly the flutter driver extension method.

The schema for this is pretty weird and complex - but in general Gemini seems to do OK with it.

Note that I only tested the tap/get_text commands, I think that is likely sufficient (really we just need to test we are forwarding commands through, we aren't doing anything special).

Also updates the workflow client to handle enums and use updated model names.

### DIFF
--- a/pkgs/dart_mcp_server/test_fixtures/counter_app/lib/driver_main.dart
+++ b/pkgs/dart_mcp_server/test_fixtures/counter_app/lib/driver_main.dart
@@ -1,0 +1,8 @@
+import 'package:flutter_driver/driver_extension.dart';
+
+import 'main.dart' as app;
+
+void main() {
+  enableFlutterDriverExtension();
+  app.main();
+}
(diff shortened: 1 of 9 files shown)


pkgs/dart_mcp_server/test_fixtures/counter_app/lib/driver_main.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:flutter_driver/driver_extension.dart';
```
kenzieschmoll, line 5: copyright?

kenzieschmoll (approved):
lgtm w one comment

Issue #220, opened 2025-07-08
`mcp_dart` example is better


gaaclarke:
I accidentally wrote the flutter engine mcp server against `mcp_dart` not `dart_mcp`, now I'm looking at converting it over.  The example for `mcp_dart` is better than the one `dart_mcp` has: https://github.com/leehack/mcp_dart/blob/main/example/example.md

It is more concise and still manages to shows a real calculation and how to read in parameters.

jakemac53:
I think this primarily is just because of the Transport class which simplifies talking over stdin/stdout without needing to add stream transforms to split/inject newlines?

jakemac53:
https://github.com/dart-lang/ai/pull/228 adds many more examples, but I am going to leave this open because some features are still not demonstrated there, although I do plan to follow up with additional examples. Specifically, we still need examples for:

- sampling
- elicitation
- completions

PR #228, opened 2025-07-15
Add more and better examples

jakemac53:
Towards https://github.com/dart-lang/ai/issues/220.

I also added a utility function for created stdio stream channels to simplify the examples (and real user code), and fixed a bug in the PromptMessage constructor.

Still missing a few examples, but this covers the most common use cases.

cc @gaaclarke


pkgs/dart_mcp/example/prompts_server.dart
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
```
gspencergoog, line 4: Maybe for these examples, it would be good to add a dartdoc comment at the top that describes what each example is meant to show and a pointer to the `README.md`. In general, for examples, these don't have enough comments.

It's also helpful when perusing the code in an IDE to have dartdoc links back to the classes that are relevant (e.g. to `MCPServer` and `PromptsSupport` for this one).

pkgs/dart_mcp/lib/stdio.dart
```diff
@@ -0,0 +1,27 @@
...
+import 'dart:convert';
+
+import 'package:async/async.dart';
+import 'package:stream_channel/stream_channel.dart';
+
+/// Creates a [StreamChannel] for Stdio communication where messages are
+/// separated by newlines.
+///
+/// This expects incoming messages on [input], and writes messages to [output].
+StreamChannel<String> stdioChannel({
```
gspencergoog, line 15: This is a good addition.

pkgs/dart_mcp/example/prompts_server.dart
jakemac53: I had Gemini add some comments and then cleaned them up a bit, also moved some things around to simplify it.

The version checking for instance doesn't need to happen manually (it is handled for you) and also things can be registered in the constructor instead of initialize which is more concise (no need to override and call super).

gspencergoog (approved):

PR #229, opened 2025-07-16
add elicitation example, fix some issues with the elicitation APIs

jakemac53:
More work towards https://github.com/dart-lang/ai/issues/220

### DIFF
--- a/pkgs/dart_mcp/lib/src/api/initialization.dart
+++ b/pkgs/dart_mcp/lib/src/api/initialization.dart
@@ -194,6 +194,7 @@ extension type ServerCapabilities.fromMap(Map<String, Object?> _value) {
     Prompts? prompts,
     Resources? resources,
     Tools? tools,
+    @Deprecated('Do not use, only clients have this capability')
     Elicitation? elicitation,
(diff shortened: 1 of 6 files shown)


pkgs/dart_mcp/lib/src/api/initialization.dart
```diff
@@ -194,6 +194,7 @@ extension type ServerCapabilities.fromMap(Map<String, Object?> _value) {
     Prompts? prompts,
     Resources? resources,
     Tools? tools,
+    @Deprecated('Do not use, only clients have this capability')
```
gspencergoog, line 197: Oops.  I thought the server had that in its schema to indicate that it might send elicitations.

gspencergoog (approved):

pkgs/dart_mcp/lib/src/api/initialization.dart
jakemac53: The capabilities are only there for the client/server indicating they can _handle_ certain requests, not whether or not they will issue them.

I think you could argue it might be useful to know what kind of requests a server would like to make, but that isn't how its structured 🤷‍♂️

Issue #238, opened 2025-07-21
Ping request from non-Dart client leads to error response.


ksokolovskyi:
### Steps to reproduce

1. Use `dart_mcp: 0.3.2`
2. Build any MCP server from the `dart_mcp/example` directory (e.g. `dart_mcp/example/tools_server.dart)
3. Connect the built server to any non-Dart MCP Client
4. Send a `ping` request, and you will receive the error

### Expected results

An empty response is getting received.

### Actual results

Error response is getting received.

```json
{
   "error": "MCP error -32000: type 'CastMap<String, dynamic, String, Object?>' is not a subtype of type 'Null' in   type cast"
}
```

### Screenshots or Video

https://github.com/user-attachments/assets/0d374db6-5c79-4d88-bc12-77d9ef3a6a8b

PR #239, opened 2025-07-21
Fix ping request handling from non-dart clients.

ksokolovskyi:
Fixes https://github.com/dart-lang/ai/issues/238

- Updates `PingRequest` to wrap `Map<String, Object?>` instead of `Null` to allow empty parameters handling from non-Dart clients.

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/pkgs/dart_mcp/lib/src/api/api.dart
+++ b/pkgs/dart_mcp/lib/src/api/api.dart
@@ -225,9 +225,11 @@ extension type RequestId( /*String|int*/ Parameter _) {}
 ///
 /// The receiver must promptly respond, or else may be disconnected.
 ///
-/// The request itself has no parameters and should always be just `null`.
-extension type PingRequest._(Null _) {
+/// The request itself has no parameters.
+extension type PingRequest._(Map<String, Object?> _) implements Request {
   static const methodName = 'ping';
+
+  factory PingRequest() => PingRequest._(const {});
 }
 
 /// An out-of-band notification used to inform the receiver of a progress
--- a/pkgs/dart_mcp/test/client_and_server_test.dart
+++ b/pkgs/dart_mcp/test/client_and_server_test.dart
@@ -127,6 +127,39 @@ void main() {
     );
   });
 
+  // Regression test for https://github.com/dart-lang/ai/issues/238.
+  test('client and server can handle ping with null parameters', () async {
+    final environment = TestEnvironment(TestMCPClient(), TestMCPServer.new);
+    await environment.initializeServer();
+
+    await expectLater(
+      environment.serverConnection.sendRequest(PingRequest.methodName, null),
+      completes,
+    );
+    await expectLater(
+      environment.server.sendRequest(PingRequest.methodName, null),
+      completes,
+    );
+  });
+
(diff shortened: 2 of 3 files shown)


pkgs/dart_mcp/lib/src/api/api.dart
```diff
@@ -225,9 +225,11 @@ extension type RequestId( /*String|int*/ Parameter _) {}
 ///
 /// The receiver must promptly respond, or else may be disconnected.
 ///
-/// The request itself has no parameters and should always be just `null`.
-extension type PingRequest._(Null _) {
+/// The request itself has no parameters.
+extension type PingRequest._(Map<String, Object?> _) implements Request {
   static const methodName = 'ping';
+
+  factory PingRequest() => PingRequest._(const {});
```
jakemac53: It looks like this should just implement `Request` now, as it explicitly has the metadata field with the progress token https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.json#L1467.

```suggestion
extension type PingRequest._(Map<String, Object?> _) implements Request {
  static const methodName = 'ping';

  factory PingRequest({
    MetaWithProgressToken? meta,
  }) => PingRequest._({
    if (meta != null) '_meta': meta,
  });
```

pkgs/dart_mcp/test/client_and_server_test.dart
```diff
@@ -127,6 +127,39 @@ void main() {
     );
   });
 
+  // Regression test for https://github.com/dart-lang/ai/issues/238.
+  test('client and server can handle ping with null parameters', () async {
```
jakemac53: Note that this is equivalent to the "client and server can ping each other" test above, although that is using the `ping` method on the client/server.

pkgs/dart_mcp/test/client_and_server_test.dart
```diff
@@ -127,6 +127,39 @@ void main() {
...
+      completes,
+    );
+    await expectLater(
+      environment.server.sendRequest(PingRequest.methodName, null),
+      completes,
+    );
+  });
+
+  // Regression test for https://github.com/dart-lang/ai/issues/238.
+  test('client and server can handle ping with no parameters', () async {
```
jakemac53:
```suggestion
  test('client and server can handle ping with non-null parameters', () async {
```

pkgs/dart_mcp/test/client_and_server_test.dart
```diff
@@ -127,6 +127,39 @@ void main() {
...
+    final environment = TestEnvironment(TestMCPClient(), TestMCPServer.new);
+    await environment.initializeServer();
+
+    await expectLater(
```
jakemac53: We could add an optional `PingRequest` argument to the shared `ping` method (see lib/src/shared.dart).

I am not really sure what the use case is for it, but the protocol does allow it so we might as well?

Then you can just call that method with an empty PingRequest instead of using `sendRequest` which is more the intended use for people.

ksokolovskyi:
Sorry, I forgot about the version bump 🤕

jakemac53:
Thanks for the PR and quick responses 🚀

PR #241, opened 2025-07-23
Flutter driver updates, add screenshot support back

jakemac53:
- All the flutter driver arguments are actually supposed to be string values so this fixes those schemas
- Change the widget tree to not be the summary by default - summary trees hide nested text widgets which makes it hard for the LLM to find things. This can be explicitly toggled on though.
- Added support for the flutter driver screenshot command, which works on simulators 🥳 
- Rework tests a bit so there is a helper for normal tool calls without retries
- Removed the `get_layer_tree` and `get_render_tree` flutter driver commands for now - I don't see these ever being used.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -476,15 +476,14 @@ base mixin DartToolingDaemonSupport
       callback: (vmService) async {
         final vm = await vmService.getVM();
         final isolateId = vm.isolates!.first.id;
+        final summaryOnly = request.arguments?['summaryOnly'] as bool? ?? false;
         try {
           final result = await vmService.callServiceExtension(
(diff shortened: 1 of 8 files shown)


pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -673,8 +672,8 @@ base mixin DartToolingDaemonSupport
           // supported, but may be in the future.
           enumValues: [
             'get_health',
-            'get_layer_tree',
-            'get_render_tree',
+            // 'get_layer_tree',
```
gspencergoog, line 675: Remove commented out code.

jakemac53: See the comment above, in this case I think it is helpful to have these because they are flutter driver methods we have chosen not to expose for now, but we could expose later. Finding these is actually pretty hard otherwise.

PR #243, opened 2025-07-24
add prompt for using flutter driver to write a test for a user journey

jakemac53:
For clients that have prompts support, this prompt allows a user to provide a user journey and it will instruct the LLM to:

- Try to accomplish the user journey in the live app using flutter driver
- If successful, write a flutter driver test with the same instructions used to accomplish the journey.

TODO: Add support for running flutter driver tests in the MCP server and add that to the prompt.

### DIFF
--- a/pkgs/dart_mcp/lib/src/api/api.dart
+++ b/pkgs/dart_mcp/lib/src/api/api.dart
@@ -503,15 +503,6 @@ extension type ResourceLink.fromMap(Map<String, Object?> _value)
     return type;
   }
 
-  /// The name of the resource.
-  String get name {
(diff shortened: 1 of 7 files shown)


pkgs/dart_mcp/lib/src/api/api.dart
```diff
@@ -503,15 +503,6 @@ extension type ResourceLink.fromMap(Map<String, Object?> _value)
     return type;
   }
 
-  /// The name of the resource.
```
jakemac53, line 506: Note that this was just duplicative, this getter exists in one of the implemented types already

PR #252, opened 2025-07-30
only run in shell when on windows and running a .bat file

jakemac53:
Based on this suggestion from @DanTup https://github.com/dart-lang/tools/pull/2138#issuecomment-3133627598

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/utils/cli_utils.dart
+++ b/pkgs/dart_mcp_server/lib/src/utils/cli_utils.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:io' as io;
 
 import 'package:collection/collection.dart';
 import 'package:dart_mcp/server.dart';
@@ -224,7 +225,10 @@ Future<CallToolResult> runCommandInRoot(
   final result = await processManager.run(
     commandWithPaths,
     workingDirectory: workingDir.path,
-    runInShell: true,
+    runInShell:
+        // Required when running .bat files on windows, but otherwise should
+        // be avoided.
+        io.Platform.isWindows && commandWithPaths.first.endsWith('.bat'),
   );
 
   final output = (result.stdout as String).trim();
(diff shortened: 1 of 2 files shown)


pkgs/dart_mcp_server/lib/src/utils/cli_utils.dart
```diff
@@ -224,7 +225,10 @@ Future<CallToolResult> runCommandInRoot(
   final result = await processManager.run(
     commandWithPaths,
     workingDirectory: workingDir.path,
-    runInShell: true,
+    runInShell:
+        // Required when running .bat files on windows, but otherwise should
+        // be avoided.
+        io.Platform.isWindows && commandWithPaths.first.endsWith('.bat'),
```
DanTup, line 231: Are you doing any escaping of the command/args before here, or does that just happen inside `processManager` when `runInShell` is true? (you probably don't want any escaping when this is false).

You might also want to include `.cmd` for completeness (I didn't in Dart-Code because I know the only Windows shell scripts are `flutter.bat`).

I don't know how this works for non-Windows. I presume it you use a shell script that has a shebang line it'll just spawn the shell anyway (Dart-Code never uses shell execute for non-Windows).

pkgs/dart_mcp_server/lib/src/utils/cli_utils.dart
```diff
@@ -224,7 +225,10 @@ Future<CallToolResult> runCommandInRoot(
   final result = await processManager.run(
     commandWithPaths,
     workingDirectory: workingDir.path,
-    runInShell: true,
+    runInShell:
+        // Required when running .bat files on windows, but otherwise should
+        // be avoided.
```
DanTup: Maybe worth expanding on this comment a little so it's clear this is because of escaping?

pkgs/dart_mcp_server/lib/src/utils/cli_utils.dart
jakemac53: > Are you doing any escaping of the command/args before here, or does that just happen inside `processManager` when `runInShell` is true? (you probably don't want any escaping when this is false).

Nothing is happening before here (or at all by us)

pkgs/dart_mcp_server/lib/src/utils/cli_utils.dart
jakemac53:
```suggestion
        // be avoided due to escaping behavior.
```

DanTup (approved):
LGTM - the test failure doesn't seem related to this, but there doesn't seem to be much in the logs to understand why it timed out.

Issue #251, opened 2025-07-30
Support `--exclude-tool x --exclude-tool y` to exclude tools from the MCP server


DanTup:
VS Code has a built-in tool for running tests:

<img width="656" height="516" alt="Image" src="https://github.com/user-attachments/assets/d68f3318-d376-4bd0-a167-041a7771a71f" />

However because of the descriptions, the LLM seems to prefer using the one from the Dart MCP server:

<img width="509" height="640" alt="Image" src="https://github.com/user-attachments/assets/bf4192a7-e1d3-4fc2-8b32-52c02ba928d7" />

Despite the LLM's response above, the results from the Dart MCP server do not integrate with the test runner, but running them through VS Code does. While I can untick the Dart MCP tool in the MCP settings, I wonder if it would be a better "default" for users of IDEs with test runners for the IDE ones to be used?

I wonder if adding something like "Use this tool in preference to running terminal commands, but not in preference to any built-in test tools" would be worthwhile?

@jakemac53

jakemac53:
This seems valid but I am not sure of the best way to deal with it that won't end up falling back on other more generic test tools in other contexts that aren't better.

One option is to add `--disable-tool <tool-name>` arguments to the MCP server.

jakemac53:
Ultimately the LLM does not know what tools come from MCP at all - it is just given functions to call. So, it's possible that just from the descriptions it can derive what is "built-in" versus not but I am not sure.

It does seem to me like the best option in this case is to just disable the tool - I don't know when it would be useful to use our tool over the VsCode built-in one?

jakemac53:
Yeah I think adding `--disable-tool` is generally a useful thing. It gives a way for people to disable tools they don't want if their client doesn't support that, as well as a nice way of providing defaults when the configuration comes from other tooling and said tooling has some smart defaults it wants to apply based on the context.

PR #253, opened 2025-07-31
add --exclude-tool option to exclude tools by name

jakemac53:
Closes https://github.com/dart-lang/ai/issues/251

cc @DanTup

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/arg_parser.dart
+++ b/pkgs/dart_mcp_server/lib/src/arg_parser.dart
@@ -47,6 +47,10 @@ ArgParser createArgParser({
           help:
               'Path to a file to log all MPC protocol traffic to. File will be '
               'overwritten if it exists.',
+        )
+        ..addMultiOption(
+          excludeToolOption,
+          help: 'The names of tools to exclude from this run of the server.',
         );
 
   if (includeHelp) parser.addFlag(helpFlag, abbr: 'h', help: 'Show usage text');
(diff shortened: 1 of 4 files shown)


bwilkerson (approved):

pkgs/dart_mcp_server/lib/src/arg_parser.dart
```diff
@@ -47,13 +47,18 @@ ArgParser createArgParser({
           help:
               'Path to a file to log all MPC protocol traffic to. File will be '
               'overwritten if it exists.',
+        )
+        ..addMultiOption(
+          excludeToolOption,
+          help: 'The names of tools to exclude from this run of the server.',
         );
```
DanTup, line 54: It might be useful to include a list of the tools, or have a `list-tools` command to make it easier to get the exact identifiers that should be used here?

DanTup (approved):
lgtm :-)

pkgs/dart_mcp_server/lib/src/arg_parser.dart
jakemac53: That one is a bit harder to do accurately because we might start wanting to have the list of tools change over time... for instance not show flutter_driver as a tool until an app with flutter_driver enabled is actually attached.

DanTup: I just meant a full list of _possible_ tools.

I noticed in VS Code that the tool names shown in the UI don't match what the LLM things they are (for example in the VS Code UI it says that the built-in tool is called `runTests` and Dart's is `run_tests`, but if I ask the model what tools it has, they are both called `run_tests`). So mostly I was just after a way to ensure I can get a list of the IDs that are _definitely the right ones_ to pass to this flag.

(grabbing it out of the source code works too, but printing them would make it harder for me to use the wrong values if there are names + ids or something).

Issue #256, opened 2025-08-02
Multiple DTD tools fail with misleading SDK version error


goodbbai:
When using the dart_mcp, several tools that interact with a running application fail with the following error message:

```
A Dart SDK of version 3.9.0-163.0.dev or greater is required to connect to Dart and Flutter applications.
```
  This error is misleading, as the Dart SDK version being used is higher than the required version.

  *Affected Tools*

  The following tools have been confirmed to be affected by this issue:

   * get_widget_tree
   * get_runtime_errors
   * hot_reload
   * get_selected_widget
   * set_widget_selection_mode

  *Root Cause Analysis*

  The root cause of this issue appears to be a bug in the _listenForServices function in pkgs/dart_mcp_server/lib/src/mixins/dtd.dart. The code that checks for the ConnectedApp service is wrapped in a try-catch block that silently ignores any errors:
```
    try {
      final registeredServices = await dtd.getRegisteredServices();
      if (registeredServices.dtdServices.contains(
        '${ConnectedAppServiceConstants.serviceName}.${ConnectedAppServiceConstants.getVmServices}',
      )) {
        _connectedAppServiceIsSupported = true;
      }
    } catch (_) {}
```
  If dtd.getRegisteredServices() throws an exception, it is caught silently, and the _connectedAppServiceIsSupported flag is not set to true. This leads to the misleading error message about the SDK version for all tools that rely on this flag.

  *Steps to Reproduce*

   1. Connect to the Dart Tooling Daemon using the connect_dart_tooling_daemon tool.
   2. Call any of the affected tools listed above.

  *Expected Behavior*

  The tools should execute successfully.

  *Actual Behavior*

  The tools fail with a misleading error message about the SDK version.

  *Environment*

   * Flutter Version: Flutter 3.35.0-0.1.pre
   * Dart Version: Dart 3.9.0 (build 3.9.0-333.2.beta)
   * Operating System: macOS 15.5 24F74 darwin-arm64

lukemmtt:
**Update:** After rebooting, **I no longer get the version-related error message**.

Instead, now I'm getting an error message that says `No active debug session to take a screenshot`.


Environment:
- MacBook Pro M4
- Flutter Version: Flutter 3.35.0-0.1.pre
- Dart Version: Dart 3.9.0 (build 3.9.0-333.2.beta)

Repro steps:

1. Run app, in debug mode, on either macOS or iPhone simulator, from Android Studio or terminal.
[...]

PR #257, opened 2025-08-03
Add diagnostics for DTD service detection failures

lukemmtt:
Adds error logging to help diagnose DTD connection issues reported in #256.

Changes:
- Capture and log exceptions in service detection
- Log registered DTD services
- Track VM service retrieval failures
- Make error messages reflect actual failure reason

This will help identify why the ConnectedApp service detection is failing for some users.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -42,6 +42,12 @@ base mixin DartToolingDaemonSupport
   /// Once we connect to dtd, this may be toggled to `true`.
   bool _connectedAppServiceIsSupported = false;
 
+  /// Stores service detection errors for diagnostic purposes.
+  String? _serviceDetectionError;
+  
+  /// Stores the last error from getVmServices for diagnostic purposes.
+  String? _vmServicesError;
+
   /// Whether to await the disposal of all [VmService] objects in
   /// [activeVmServices] upon server shutdown or loss of DTD connection.
   ///
@@ -68,6 +74,8 @@ base mixin DartToolingDaemonSupport
     _dtd = null;
     _activeLocation = null;
     _connectedAppServiceIsSupported = false;
+    _serviceDetectionError = null;
+    _vmServicesError = null;
 
     // TODO: determine whether we need to dispose the [inspectorObjectGroup] on
     // the Flutter Widget Inspector for each VM service instance.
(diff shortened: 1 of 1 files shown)


pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -997,16 +1031,21 @@ base mixin DartToolingDaemonSupport
...
-        text:
-            'A Dart SDK of version 3.9.0-163.0.dev or greater is required to '
-            'connect to Dart and Flutter applications.',
-      ),
-    ],
-  )..failureReason = CallToolFailureReason.connectedAppServiceNotSupported;
+  CallToolResult get _connectedAppsNotSupported {
+    String errorText;
+    if (_serviceDetectionError != null) {
+      errorText = 'ConnectedApp service not available: $_serviceDetectionError';
```
jakemac53: Note that in this case it is usually going to be because the user isn't on the required SDK version, so it would be good to include that in the text here?

jakemac53:
Looks like there are still some unformatted files (I think we format with the dev SDK so this could be related possibly).

Note also that you don't need to rebase/force push we squash all commits when we merge so the history here doesn't matter (and it makes reviewing better if you don't force push as that destroys history).

lukemmtt:
Thanks for the tips @jakemac53 🙏🏻 will address these promptly and push, thank you!

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: @lukemmtt it would be good to resolve this comment as well if you can :)

lukemmtt: Hey @jakemac53 , no problem, although I'm not certain I understood. Is this what you had in mind?
Commit: https://github.com/dart-lang/ai/pull/257/commits/a91675f89234f69b147b17b12466d5978e8b56e8

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -1045,16 +1081,24 @@ base mixin DartToolingDaemonSupport
...
-            'connect to Dart and Flutter applications.',
-      ),
-    ],
-  )..failureReason = CallToolFailureReason.connectedAppServiceNotSupported;
+  CallToolResult get _connectedAppsNotSupported {
+    String errorText;
+    if (_serviceDetectionError != null) {
+      errorText =
+          'ConnectedApp service not available: $_serviceDetectionError. '
+          'A Dart SDK of version 3.9.0-163.0.dev or greater is required to '
```
gspencergoog: Let's go ahead and change this to Dart 3.9.0-333.2.beta or greater, since you can't get that dev release directly.  Dart 3.9.0-333.2.beta  is included in Flutter 3.35.0-0.0.pre.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -1045,16 +1081,24 @@ base mixin DartToolingDaemonSupport
...
+          'connect to Dart and Flutter applications.';
+    } else {
+      errorText =
+          'A Dart SDK of version 3.9.0-163.0.dev or greater is required to '
```
gspencergoog: And here too.

Issue #259, opened 2025-08-08
`analyze_files` should filter results to the paths supplied.


gspencergoog:
# Problem

If you have a LOT of analyzer errors, but the LLM supplies paths to analyze, the `analyze_files` tool currently returns ALL of the analyzer errors, regardless of the paths supplied.  While this can help the LLM get a bigger picture of the problems, it can also eat up a lot of context really quickly with things that are irrelevant to the file it is working on.

For example:
<img width="1799" height="226" alt="Image" src="https://github.com/user-attachments/assets/7c664835-e34b-432c-ae2b-9430457087bf" />

The LSP only supports analyzing whole projects, so the `analyze_files` tool should filter the results so that it only includes files matching the paths supplied in the "paths" argument.  Without a paths argument, it should return the entire set of results.

PR #260, opened 2025-08-08
Add ability to filter analzyer results on a set of paths

gspencergoog:
# Description

This adds the ability to filter the results of `analyze_files` with a set of paths.

# Tests
- Added tests for added functionality.

# Related Issues
- Fixes https://github.com/dart-lang/ai/issues/259

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
@@ -249,8 +249,42 @@ base mixin DartAnalyzerSupport
     final errorResult = await _ensurePrerequisites(request);
     if (errorResult != null) return errorResult;
 
+    final paths = (request.arguments?[ParameterNames.paths] as List?)
+        ?.cast<String>();
+
+    Iterable<MapEntry<Uri, List<lsp.Diagnostic>>> entries;
+
+    if (paths != null && paths.isNotEmpty) {
+      final roots = await this.roots;
+      final requestedUris = <Uri>{};
+      for (final path in paths) {
+        final pathAsUri = Uri.tryParse(path);
+        if (pathAsUri != null && pathAsUri.scheme.isNotEmpty) {
+          requestedUris.add(pathAsUri);
+        } else {
+          // Treat as a relative or absolute path.
+          for (final root in roots) {
+            final rootUri = Uri.parse(root.uri);
+            final resolvedUri = rootUri.resolve(path);
+            requestedUris.add(resolvedUri);
+          }
+        }
+      }
+
+      entries = diagnostics.entries.where((entry) {
+        final entryPath = entry.key.toFilePath();
+        return requestedUris.any((uri) {
+          final requestedPath = uri.toFilePath();
+          return entryPath == requestedPath ||
+              entryPath.startsWith('$requestedPath${Platform.pathSeparator}');
+        });
+      });
+    } else {
+      entries = diagnostics.entries;
+    }
+
     final messages = <Content>[];
-    for (var entry in diagnostics.entries) {
+    for (var entry in entries) {
       for (var diagnostic in entry.value) {
         final diagnosticJson = diagnostic.toJson();
         diagnosticJson[ParameterNames.uri] = entry.key.toString();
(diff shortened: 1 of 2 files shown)


pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -411,8 +445,19 @@ base mixin DartAnalyzerSupport
   @visibleForTesting
   static final analyzeFilesTool = Tool(
     name: 'analyze_files',
-    description: 'Analyzes the entire project for errors.',
-    inputSchema: Schema.object(),
+    description:
+        'Analyzes individual paths, or the entire project, for errors.',
+    inputSchema: Schema.object(
+      properties: {
+        ParameterNames.paths: Schema.list(
```
jakemac53: See the dart_format tool in `dash_cli.dart` - we handle this a bit differently there and it would be good to be consistent. Basically it accepts a list of roots, and then relative paths under those roots.

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -249,8 +249,42 @@ base mixin DartAnalyzerSupport
...
 
+    final paths = (request.arguments?[ParameterNames.paths] as List?)
+        ?.cast<String>();
+
+    Iterable<MapEntry<Uri, List<lsp.Diagnostic>>> entries;
+
+    if (paths != null && paths.isNotEmpty) {
+      final roots = await this.roots;
+      final requestedUris = <Uri>{};
+      for (final path in paths) {
```
jakemac53: See `validateRootConfig` under the `lib/src/utils/cli_utils.dart` which will validate that a path is under one of the roots.

gspencergoog: Do you mean `_isUnderRoot`?

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
gspencergoog: Ahh, OK, I am doing that now, and using the path arguments from the roots config.

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: `validateRootConfig` uses that function but does some additional stuff as well, and handles actually extracting the `Root` object and the paths for you, given a single "root" configuration object (one entry in `roots`).

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -249,8 +251,47 @@ base mixin DartAnalyzerSupport
     final errorResult = await _ensurePrerequisites(request);
     if (errorResult != null) return errorResult;
 
+    var rootConfigs = (request.arguments?[ParameterNames.roots] as List?)
+        ?.cast<Map<String, Object?>>();
+    final allRoots = await roots;
+
+    if (rootConfigs != null && rootConfigs.isEmpty) {
+      // Empty list of roots means do nothing.
+      return CallToolResult(content: [TextContent(text: 'No errors')]);
```
jakemac53: I think this should either return errors for the entire project or return an error response. This isn't really a valid request?

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
gspencergoog: There is no [validateRootConfig](https://github.com/search?q=repo%3Adart-lang%2Fai%20validateRootConfig&type=code) in the repo

jakemac53: Oh hmmm, I think this is something that only exists in my "summaries" branch lol, hence the confusion 🤣

jakemac53: yeah, https://github.com/dart-lang/ai/commit/2a83a89262b51dbf2a83b9dac41c45068104ee3a.

You could roll that into this PR, its basically just extracting out some shared logic from the code that runs a command in a root given a configuration.

jakemac53 (approved):
Going to go ahead and approve this - feel free to integrate that helper method in or not, up to you. I am out next week and don't want to hold you up :)

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
gspencergoog: Okay, I incorporated it, and added unit tests for it.  I just incorporated the `validateRootConfig` function, not the rest of that PR.

PR #262, opened 2025-08-11
Stop reporting errors for non-zero exits

### LINKED ISSUE #261: Dart commands that return non zero exit codes shouldn't set isError to true.
Recently, I've had a lot of issues with the `run_tests` tool in Gemini CLI where the LLM interprets the "isError" field to mean that the test tool failed to run, rather than that tests have failed. It basically keeps trying to re-run "run_tests" because it thinks there is an error in parsing the test output. It makes the `run_tests` tool useless.

I think that a non-zero exit code for things like `dart test`, `dart format`, etc. shouldn't be interpreted as a "failure" of the tool.  The LLM doesn't get to see the "failureReason", so it can't tell the difference between missing roots or some other failure and a failed test.

I think that the `isError` should only be set if the tool actually failed to run.

The output of "run_tests" would normally be expected to include test failures. We use the error exit code to inform CI and Unix tools that there were test failures, but even then they can't tell the difference between a path error and a test failure.

It's unlikely that the LLM will interpret a tool that doesn't have `isError` set as having all passing tests in the presence of test output showing failures.

As a data point, the `analyze_files` tool doesn't set `isError` when there are analyzer errors, and the LLM can interpret that perfectly well.

gspencergoog:
## Description

This stops reporting command line tools that return non-zero exit codes as "isError" failures of the tool, so that the LLM won't interpret failed tests or formatting of files as a failure to run the tool.

## Related Issues
 -  Fixes #261

## Tests
 - Updated tests.

### DIFF
--- a/pkgs/dart_mcp_server/CHANGELOG.md
+++ b/pkgs/dart_mcp_server/CHANGELOG.md
@@ -13,6 +13,7 @@
   based on their text values.
 * Add an `--exclude-tool` command line flag to exclude tools by name.
 * Add the abillity to limit the output of `analyze_files` to a set of paths.
+* Stop reporting non-zero exit codes from command line tools as tool errors.
 
 # 0.1.0 (Dart SDK 3.9.0)
 
--- a/pkgs/dart_mcp_server/lib/src/utils/analytics.dart
+++ b/pkgs/dart_mcp_server/lib/src/utils/analytics.dart
@@ -130,7 +130,6 @@ enum CallToolFailureReason {
   noRootGiven,
   noRootsSet,
   noSuchCommand,
-  nonZeroExitCode,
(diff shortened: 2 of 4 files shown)


bwilkerson (approved):

gspencergoog:
@bwilkerson Just as a sanity check, do you think this is a good idea? I think it's the correct move, but I'd like some confirmation from another human. :-)  See [comment](https://github.com/dart-lang/ai/issues/261#issuecomment-3176634735) in the related issue.

bwilkerson:
On the surface it seems fine, but I'm not sure I have enough context to answer that question with confidence. I'm still coming  up to speed on MCP. Happy to VC if you'd like to discuss it.

PR #270, opened 2025-08-22
add an example of using progress tokens to the tools example

jakemac53:
Initially created this just for testing Gemini CLI support for progress tokens, but it makes sense to land.

### DIFF
--- a/pkgs/dart_mcp/example/tools_client.dart
+++ b/pkgs/dart_mcp/example/tools_client.dart
@@ -73,6 +73,8 @@ void main() async {
           'parts': ['a', 'b', 'c', 'd'],
           'delay': delayMs,
         },
+        // Note that in the real world you need unique tokens, either a UUID
+        // or auto-incrementing int would suffice.
         meta: MetaWithProgressToken(progressToken: ProgressToken(1)),
       );
       // Make sure to listen before awaiting the response - you could listen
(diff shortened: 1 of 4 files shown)


jakemac53:
We might eventually want to make this a bit more ergonomic too - such as including a factory for creating unique progress tokens.

pkgs/dart_mcp/example/tools_client.dart
```diff
@@ -62,20 +62,36 @@ void main() async {
...
-      final result = await server.callTool(
-        CallToolRequest(
-          name: tool.name,
-          arguments: {
-            'parts': ['a', 'b', 'c', 'd'],
-          },
-        ),
+      const delayMs = 2000;
+      print(
+        'Calling `${tool.name}` tool, with an artificial ${delayMs}ms second '
```
bwilkerson, line 67: nit: Remove "second".

pkgs/dart_mcp/example/tools_client.dart
```diff
@@ -62,20 +62,36 @@ void main() async {
...
+        arguments: {
+          'parts': ['a', 'b', 'c', 'd'],
+          'delay': delayMs,
+        },
+        // Note that in the real world you need unique tokens, either a UUID
+        // or auto-incrementing int would suffice.
+        meta: MetaWithProgressToken(progressToken: ProgressToken(1)),
       );
+      // Make sure to listen before awaiting the response - you could listen
+      // after sending the request but before awaiting the result as well.
```
bwilkerson, line 81: If both scenarios should work then we should probably test both.

bwilkerson (approved):

PR #274, opened 2025-08-28
Add build_runner build tool.

gspencergoog:
# Description

This adds a tool to be able to run the [`build_runner`](https://pub.dev/packages/build_runner) package's build function.  This is a common code generation package, and having it as a tool will allow it to be approved separately from command line execution in many agents, and it gives the LLM a clear way to invoke `build_runner` to build generated code for common packages like `json_serializable` or `freezed`.

---

### DIFF
--- a/pkgs/dart_mcp_server/CHANGELOG.md
+++ b/pkgs/dart_mcp_server/CHANGELOG.md
@@ -1,66 +1,67 @@
 # 0.1.1 (Dart SDK 3.10.0) - WIP
 
-* Change tools that accept multiple roots to not return immediately on the first
+- Change tools that accept multiple roots to not return immediately on the first
   failure.
-* Add failure reason field to analytics events so we can know why tool calls are
+- Add failure reason field to analytics events so we can know why tool calls are
   failing.
-* Add a flutter_driver command for executing flutter driver commands on a device.
-* Allow for multiple package arguments to `pub add` and `pub remove`.
-* Require dart_mcp version 0.3.1.
-* Add support for the flutter_driver screenshot command.
-* Change the widget tree to the full version instead of the summary. The summary
+- Add a flutter_driver command for executing flutter driver commands on a device.
+- Allow for multiple package arguments to `pub add` and `pub remove`.
+- Require dart_mcp version 0.3.1.
+- Add support for the flutter_driver screenshot command.
+- Change the widget tree to the full version instead of the summary. The summary
   tends to hide nested text widgets which makes it difficult to find widgets
   based on their text values.
-* Add an `--exclude-tool` command line flag to exclude tools by name.
-* Add the abillity to limit the output of `analyze_files` to a set of paths.
-* Stop reporting non-zero exit codes from command line tools as tool errors.
+- Add an `--exclude-tool` command line flag to exclude tools by name.
+- Add the abillity to limit the output of `analyze_files` to a set of paths.
+- Stop reporting non-zero exit codes from command line tools as tool errors.
+- Add support for a tool that runs `dart run build_runner build`.
 
 # 0.1.0 (Dart SDK 3.9.0)
 
-* Add documentation/homepage/repository links to pub results.
-* Handle relative paths under roots without trailing slashes.
-* Fix executable paths for dart/flutter on windows.
-* Pass the provided root instead of the resolved root for project type detection.
-* Be more flexible about roots by comparing canonicalized paths.
-* Create the working dir if it doesn't exist.
-* Add the --platform and --empty arguments to the flutter create tool.
-* Invoke dart/flutter in a more robust way.
-* Remove qualifiedNames from the pub dev api search.
-* Flutter/Dart create tool.
-* Limit the tokens returned by the runtime errors tool/resource.
-* Add RootsFallbackSupport mixin.
-* Fix error handling around stream listeners.
-* Add a 'pub-dev-search' mcp tool.
-* Drop pubspec-parse, use yaml instead.
-* Handle failing to listen to vm service streams during startup.
-* Add tool for enabling/disabling the widget selector.
-* Add a tool to get the active cursor location.
-* Add hover tool support.
-* Add a test command and project detection.
-* Add signature_help tool.
-* Add runtime errors resource and tool to clear errors.
-* Require roots for all CLI tools.
-* Require roots to be set for analyzer tools.
-* Add debug logs for when DTD sees Editor.getDebugSessions get registered.
-* Add tool annotations to tools.
-* Implement a tool to resolve workspace symbols based on a query.
-* Add a dart pub tool.
-* Update analyze tool to use LSP, simplify tool.
-* Add tool for getting the selected widget.
-* Handle missing roots capability better.
-* Add `get_widget_tree` tool.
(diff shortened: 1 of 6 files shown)


pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
```diff
@@ -31,8 +31,9 @@ base mixin DashCliSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
       if (supportsRoots && sdk.dartSdkPath != null) {
         registerTool(dartFixTool, _runDartFixTool);
         registerTool(dartFormatTool, _runDartFormatTool);
-        registerTool(runTestsTool, _runTests);
+        registerTool(runTestsTool, _runTestsTool);
         registerTool(createProjectTool, _runCreateProjectTool);
+        registerTool(buildRunnerTool, _runBuildRunnerTool);
```
jakemac53, line 36: Not sure how we want to deal with code changes, but I would generally probably try to only register this tool if build_runner is actually a dev dependency of one of the added roots.

We generally want to limit the number of tools, especially tools that are just proxies for Shell commands (if anything, we may want to remove this entire mixin and all the tools it registers).

But especially tools that are guaranteed to fail (if the package doesn't use codegen), we should not be registering.

gspencergoog: Ok, I can abandon this change if you want, but I've noticed the LLM wanting to run `build_runner` even before it has added it as a dependency, and having the guidance in this tool's description might help it to add the dependency before running the tool.

I also am loathe to give it blanket shell permission for "dart", but I'm fine giving this tool permission to run, which makes having these tools useful to me, even if they are thin wrappers.

Also, the registered root might be above the packages in a project, and might not have a `pubspec.yaml` anyhow, even if some of the packages do, making it hard to check for at tool registration time.

jakemac53: It's definitely a fair point that we would have to recursively search all subdirectories to see if _any_ of them depend on build_runner, which we likely don't want to do, although we could likely share this understanding of "all known packages" across other tools in a useful way.

In general I think it's worth asking whether we should just be exposing all executables from all dependencies... likely through just a single `run` command. This could possibly expose a structured list of the available commands through a `command` parameter, or we could just rely on training data or users guiding the AI via their *.md files as to what commands they want it to run and how.

jakemac53: Note that we also anyways might want a `flutter run` equivalent command so that it can be longer lived (it can return once the app starts up but leave it running in the background, and auto-connect to DTD, etc).

pkgs/dart_mcp_server/CHANGELOG.md
```diff
@@ -1,66 +1,67 @@
 # 0.1.1 (Dart SDK 3.10.0) - WIP
 
-* Change tools that accept multiple roots to not return immediately on the first
```
jwren: Changelog changes should be a separate PR, will this now be standard with other pkgs from us?  Do we have this documented somewhere?  With the IntelliJ plugins, we are standardizing to https://keepachangelog.com/en/1.1.0/ (shout out to @helin24)

pkgs/dart_mcp_server/README.md
```diff
@@ -150,6 +150,7 @@ For more information, see the official VS Code documentation for
 | `dart_format` | Dart format | Runs `dart format .` for the given project roots. |
 | `run_tests` | Run tests | Run Dart or Flutter tests with an agent centric UX. ALWAYS use instead of `dart test` or `flutter test` shell commands. |
 | `create_project` | Create project | Creates a new Dart or Flutter project. |
+| `build_runner` | Run build_runner | Runs `dart run build_runner build` for the given project roots. For this to succeed, the build_runner package must be a dev dependency of the package it is run on. |
```
jwren, line 153: I am not necessarily against this additional tool, but we need to have some set of core-principles to figure out questions that are being posed below ahead of time.

jwren (approved):

pkgs/dart_mcp_server/CHANGELOG.md
gspencergoog: Sorry, I'll revert this.  This is just because I have "fixAll" turned on in VSCode, and my Markdown linter is "fixing" them.

pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
domesticmouse: FWIW, i'm +1 on the idea of adding a `build_runner` tool, because I'm always telling Gemini CLI to rebuild things.

gspencergoog: > In general I think it's worth asking whether we should just be exposing all executables from all dependencies... likely through just a single run command. This could possibly expose a structured list of the available commands through a command parameter, or we could just rely on training data or users guiding the AI via their *.md files as to what commands they want it to run and how.

I think the problem with this is that it's a moving target.  If I add `build_runner` to a package dependency, then the MCP server will have to re-register the "dart_run" tool with the LLM again with an updated list of available commands. Which is fine, but it means something has to be watching to see when the dependencies change and update the tools, which is probably a long-running filesystem monitoring thread in the server.

Adding a filesystem monitoring thread to the server could open up some good functionality, but it could also be expensive, and means that adding a root means doing an initial search of that root so you know what the initial conditions are. That initial search could be expensive, as could the monitoring (if, say, someone is just moving a lot of stuff around, it could cause a lot of side effect actions).

gspencergoog: Also, if we're discovering the executable dependencies and enumerating them, then we should also incorporate the descriptions of each of the executables somehow.  Otherwise how else will the LLM know that there is a `build` subcommand for the `build_runner` executable?

jakemac53: If all we are exposing is a general purpose "run" command, without the explicit structured list of available commands, then that doesn't require us to do any file watching or discovery. It would also be cheaper in terms of the context size.

I think it's fairly reasonable to ask people to add something to their agents MD files telling it which commands they want it to run and when.

But, it begs the question of why expose these commands at all then, when you can just instruct your agent to run `dart run build_runner build` as a shell command. Pretty much all agents support shell commands. The only real advantage I see there is the finer granularity around granting permissions.

We could potentially alter the behavior somewhat - limiting the output of the command to only that which is useful, with progress messages etc. But, this feels to me like logic that would be better served as either some sort of plugin to the dart MCP server or as a separate MCP server, that people can opt into using.

jakemac53:
The more I think about it, the more I am actually not in favor to landing this as a tool in the MCP server, at least in the current state.

I believe it is more likely to cause issues for non-build_runner users than it is to solve problems for build_runner users.

I think it would be totally reasonable to file a request on build_runner to implement its own MCP server though, and then people can just use that. Of course the discoverability won't be as good, but its the correct place for it imo, this should be an opt-in thing.

PR #275, opened 2025-08-29
Create an initial `.gemini/settings.json` file to dogfood the Dart MCP Server

jwren:


gspencergoog:
I wonder if this would be better as a [Gemini CLI extension](https://github.com/google-gemini/gemini-cli/blob/main/docs/extension.md)?  Then someone would only need to unpack it instead of editing their settings file manually. And it could also include a starter prompt file that would be included in the context without disturbing the user's GEMINI.md.

jwren:
> I wonder if this would be better as a [Gemini CLI extension](https://github.com/google-gemini/gemini-cli/blob/main/docs/extension.md)? Then someone would only need to unpack it instead of editing their settings file manually. And it could also include a starter prompt file that would be included in the context without disturbing the user's GEMINI.md.

A next task might have been to add contents into a `GEMINI.md` file for this directory: Project-specific context (./GEMINI.md): A GEMINI.md file located in a project directory will override or supplement the global context. This allows you to set specific rules for a single project, such as a different coding style or specific terminology. The CLI will traverse up the directory tree from your current working directory to find and apply these files.

I am fine not landing, this but then we should choose to include `.gemini/` into the `.gitignore` file to reduce churn.

gspencergoog:
I still think an extension is probably more useful here.  For instance, in my ~/.gemini/settings.json, I set the `contextFileName` to `agents.md`, which would mean that a `GEMINI.md` file in this directory would be ignored for me. But an extension in `<workspace>/.gemini/extensions` would be able to set its own `contextFileName` and thus would still get picked up.

Instead of adding this settings.json file, you would add this instead:

```txt
.gemini/
└── extensions
    └── dart
        ├── dart.md
        ├── gemini-extension.json
        └── README.md
```

Where the gemini-extension.json could have:

```json
{
  "name": "dart",
  "description": "Enables several Dart-related commands and context.",
  "version": "0.1.0",
  "contextFileName": "dart.md",
  "mcpServers": {
    "dart_mcp_server": {
      "command": "dart",
      "arguments": ["mcp-server"],
      "trust": true
    }
  }
}
```

PR #278, opened 2025-09-03
Add pub add format guidance

gspencergoog:
# Description

This adds some short descriptions to the pub tools on how to add packages with other kinds of references, and in other sections.  Gemini doesn't seem to know how to do this otherwise.

Also, converted the pub tool's argument to be an enum of the available tools, and added support for `deps` and `outdated`.

It's not an issue here because we're passing arguments directly, but I wish that pub add would parse "foo:{path:../foo}" (without the space after the colon, instead of "foo:{path: ../foo}" ), because it's a pain to have to escape the space or quote the arg.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/pub.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/pub.dart
@@ -91,9 +91,7 @@ base mixin PubSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
           enumValues: SupportedPubCommand.values
               .map<String>((e) => e.name)
               .toList(),
-          description:
-              'Only ${SupportedPubCommand.listAll} are supported.\n'
-              '${SupportedPubCommand.commandDescriptions}',
+          description: SupportedPubCommand.commandDescriptions,
         ),
         ParameterNames.packageNames: Schema.list(
           title: 'The package names to run the command for.',
@@ -174,7 +172,7 @@ enum SupportedPubCommand {
   }
 
   static String get commandDescriptions {
-    return _getDescriptions(values);
+    return 'Available subcommands:\n${_getDescriptions(values)}';
   }
 
   static String _getDescriptions(Iterable<SupportedPubCommand> commands) {


pkgs/dart_mcp_server/lib/src/mixins/pub.dart
```diff
@@ -87,7 +87,15 @@ base mixin PubSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
     inputSchema: Schema.object(
       properties: {
         ParameterNames.command: Schema.string(
```
jakemac53, line 89: Should this also be an enum type instead of just adding the supported commands to the description?

pkgs/dart_mcp_server/lib/src/mixins/pub.dart
```diff
@@ -87,7 +87,15 @@ base mixin PubSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
...
-          title: 'The pub command to run.',
+          title: '''The pub subcommand to run.
```
jakemac53: See the `add` command below, it has a TODO for supporting dev dependencies etc, does this resolve that?

gspencergoog: Yes, it does, but in reading that, I realized that this text should probably be on the `packageNames` argument instead of here.

pkgs/dart_mcp_server/lib/src/mixins/pub.dart
gspencergoog: Yes, that would probably be more clear to the LLM.

pkgs/dart_mcp_server/lib/src/mixins/pub.dart
```diff
@@ -87,29 +87,18 @@ base mixin PubSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
...
-    - branch and subdir are optional.
-  - Local path reference: "pkg_name:{path: ../pkg_name}"
-  - Dev Dependency: "dev:pkg_name"
-''',
+          title: 'The pub subcommand to run.',
+          enumValues: SupportedPubCommand.values.map<String>((e) => e.name),
           description:
-              'Currently only ${SupportedPubCommand.listAll} are supported.',
+              'Only ${SupportedPubCommand.listAll} are supported.\n'
+              '${SupportedPubCommand.commandDescriptions}',
```
jakemac53: nit: this first line can probably be removed now, as this is communicated through the enum values

```suggestion
          description: SupportedPubCommand.commandDescriptions,
```

gspencergoog: Done.

PR #279, opened 2025-09-04
add a GEMINI.md file

jakemac53:
Created using the `/init` command from Gemini CLI and then asking it to elaborate on a few things

### DIFF
--- a/GEMINI.md
+++ b/GEMINI.md
@@ -1,0 +1,98 @@
+# Gemini Workspace Context
+
+## Project Overview
+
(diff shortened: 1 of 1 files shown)


jakemac53:
You can just ask Gemini to update it periodically, it's just for Gemini itself.

We could even potentially build that into the MD file itself? Add something like "While working if you discover this file to be out of date or missing useful information please update it at will, it is a living document".

jakemac53:
I added a note to the top of the file encouraging the agents to keep it up to date, we can see how that goes.

PR #296, opened 2025-10-07
[Feat] Add `hot_restart` support to `dart_mcp_server`

### LINKED ISSUE #254: Add a hot restart tool
Sometimes hot reload isn't quite enough, especially at early stages of app development. We should expose hot restart as well.

stefanoamorelli:
Adds `hot_restart` tool support to the `dart_mcp_server` Flutter launcher, enabling hot restart functionality for running Flutter applications while maintaining the current session.

This implementation is based on the changes from https://github.com/gemini-cli-extensions/flutter/pull/1 and addresses issue #254.

---

Closes #254

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/flutter_launcher.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/flutter_launcher.dart
@@ -40,6 +40,7 @@ base mixin FlutterLauncherSupport
     registerTool(listDevicesTool, _listDevices);
     registerTool(getAppLogsTool, _getAppLogs);
     registerTool(listRunningAppsTool, _listRunningApps);
+    registerTool(hotRestartTool, _hotRestart);
     return super.initialize(request);
   }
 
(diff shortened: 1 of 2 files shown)


gspencergoog:
Thanks for the submission @stefanoamorelli !

And thank you for your previous submission for the `flutter_launcher`  MCP server, we really appreciate the effort.  That code would have worked well there (except it's going away), but probably isn't the right approach here in this MCP server.

Here, hot restart should work similarly to hot reload, in that it should use the connection to the DTD to send a restart rather than using the process ID as an identifier and interacting with the `flutter run` process.  This MCP server knows more about how to connect to the DTD and talk to it, and since there's already a `hotReload` command, it makes sense to also have `hotRestart` use the same mechanisms.

Because that's significantly different from your approach here, and because it's actually pretty obscure how to request it, I went ahead and built a separate PR to add this part.

So I'm going to go ahead and close this PR in favor of https://github.com/dart-lang/ai/pull/297 which does the same thing, but using the DTD instead.

sethladd:
Thanks for the contribution!

PR #298, opened 2025-10-20
Fix link for adding Dart MCP server to Cursor

failutee:
### Updated the link for adding Dart MCP server to Cursor.

Resolved an issue where clicking the "Add to Cursor" button would only open the SVG image in the browser. The button's hyperlink has been corrected to the proper installation link.

---

<details>

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>

### DIFF
--- a/pkgs/dart_mcp_server/README.md
+++ b/pkgs/dart_mcp_server/README.md
@@ -73,7 +73,7 @@ for configuring the server in Android Studio. -->
 
 ### Cursor
 
-[![Add to Cursor](https://cursor.com/deeplink/mcp-install-dark.svg)](cursor://anysphere.cursor-deeplink/mcp/install?name=dart&config=eyJ0eXBlIjoic3RkaW8iLCJjb21tYW5kIjoiZGFydCBtY3Atc2VydmVyIC0tZXhwZXJpbWVudGFsLW1jcC1zZXJ2ZXIgLS1mb3JjZS1yb290cy1mYWxsYmFjayJ9)
+[![Add to Cursor](https://cursor.com/deeplink/mcp-install-dark.svg)](https://cursor.com/en/install-mcp?name=dart&config=eyJ0eXBlIjoic3RkaW8iLCJjb21tYW5kIjoiZGFydCBtY3Atc2VydmVyIC0tZXhwZXJpbWVudGFsLW1jcC1zZXJ2ZXIgLS1mb3JjZS1yb290cy1mYWxsYmFjayJ9)
 
 The easiest way to configure the Dart MCP server with Cursor is by clicking the "Add to Cursor"
 button above.


pkgs/dart_mcp_server/README.md
```diff
@@ -73,7 +73,7 @@ for configuring the server in Android Studio. -->
 
 ### Cursor
 
-[![Add to Cursor](https://cursor.com/deeplink/mcp-install-dark.svg)](cursor://anysphere.cursor-deeplink/mcp/install?name=dart&config=eyJ0eXBlIjoic3RkaW8iLCJjb21tYW5kIjoiZGFydCBtY3Atc2VydmVyIC0tZXhwZXJpbWVudGFsLW1jcC1zZXJ2ZXIgLS1mb3JjZS1yb290cy1mYWxsYmFjayJ9)
+[![Add to Cursor](https://cursor.com/deeplink/mcp-install-dark.svg)](https://cursor.com/en/install-mcp?name=dart&config=eyJ0eXBlIjoic3RkaW8iLCJjb21tYW5kIjoiZGFydCBtY3Atc2VydmVyIC0tZXhwZXJpbWVudGFsLW1jcC1zZXJ2ZXIgLS1mb3JjZS1yb290cy1mYWxsYmFjayJ9)
```
jakemac53, line 76: See https://cursor.com/docs/context/mcp/install-links - the current link looks correct to me. This is the old style link I think (we had a different PR to migrate to this new style link).

You may need to update cursor.

failutee: The link might be correct, but it uses a protocol that GitHub can’t open automatically. Go to the README and click the button, it will open the SVG file in your browser instead of launching Cursor with the MCP installation. It’s not caused by an incorrect version of Cursor.

jakemac53: Afaik cursor itself has to register a handler for the `cursor:` scheme with the OS. So, if you don't have the correct version of cursor installed which does that, it will fall back on trying to load it in chrome? I cannot validate the link myself though, as I do not have cursor (and am not allowed to install it)

failutee: I just checked this with the latest version of Cursor... it’s probably not that case. GitHub doesn’t recognize this protocol and doesn’t treat it as a valid link (you can verify that by removing the SVG icon). From what I can tell, GitHub only supports the HTTP and HTTPS protocols. That’s why, at the moment, when a user clicks the button, the SVG file just opens in the browser.

PR #303, opened 2025-10-22
Use process.kill instead of process manager killPid

gspencergoog:
## Description

In my last change, I had switched to `processManager.killPid(process.pid)` to kill a process because it was easier to test, but that doesn't appear to actually work.  It works fine in tests, which is why I didn't catch it, but it doesn't actually kill the process in the real app.

Seems like a bug in  `ProcessManager`, but maybe there's something subtle happening here.  In any case, this now kills processes properly.

## Tests
 - Removed the test for what processes were killed by the process manager, since it's no longer doing that. I don't see a good way to test if `process.kill()` was called...

### DIFF
--- a/pkgs/dart_mcp_server/test/tools/flutter_launcher_test.dart
+++ b/pkgs/dart_mcp_server/test/tools/flutter_launcher_test.dart
@@ -318,7 +318,7 @@ void main() {
             'test-device',
           ],
           stderr: 'Something went wrong',
-          exitCode: Future.value(1),
+          exitCode: 1,
         ),
       );
       final serverAndClient = await createServerAndClient(
@@ -543,7 +543,10 @@ void main() {
       );
 
       test.expect(result.structuredContent, {'success': true});
-      test.expect(mockProcessManager.killedPids, [11111, 22222, processPid]);
+      test.expect(mockProcessManager.killedPids, [
+        if (Platform.isLinux) ...[11111, 22222],
+        processPid,
+      ]);
       test.expect(result.isError, test.isNot(true));
       await server.shutdown();
       await client.shutdown();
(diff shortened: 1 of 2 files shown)


jakemac53:
@gspencergoog I looked at the implementation and it just calls `Process.killPid` from dart:io so that seems very odd indeed... https://github.com/dart-lang/tools/blob/main/pkgs/process/lib/src/interface/local_process_manager.dart#L128

pkgs/dart_mcp_server/test/tools/flutter_launcher_test.dart
```diff
@@ -460,7 +460,6 @@ void main() {
             'TimeoutException',
           ]),
         );
-        test.expect(mockProcessManager.killedPids, [processPid]);
```
jakemac53, line 463: You should be able to look for the mocked processes here and then check if they were killed

gspencergoog:
When I do a `kill -HUP` on the flutter process (on Linux) from the command line, it is exiting, and it's taking down one child, but it leaves the `dart develop` command and the app running.  It kills the `flutter` command (the main process) and the `frontend_server_aot.dart.snapshot`.

jakemac53:
`flutter run` also supports sending a `q` character to exit, maybe we should try just sending that?

pkgs/dart_mcp_server/lib/src/mixins/flutter_launcher.dart
```diff
@@ -274,8 +274,33 @@ base mixin FlutterLauncherSupport
         content: [TextContent(text: 'Application with PID $pid not found.')],
       );
     }
-
-    final success = processManager.killPid(pid);
+    // On Unix, killing the flutter process doesn't kill the entire process
+    // group, so we have to look for the child processes.
+    if (Platform.isLinux) {
+      final ps = processManager.runSync([
```
jakemac53, line 280: nit: lets run this in a try/catch just in case it fails for some reason?

Issue #307, opened 2025-11-10
Consider a flag to disable all Flutter related tools for non-flutter workspaces


DanTup:
The list of tools in the MCP server is quite long now. Many of these tools are only applicable to Flutter projects and use up additional context (and could confuse the model) so it would be nice to hide them all when in a Dart project:

<img width="647" height="531" alt="Image" src="https://github.com/user-attachments/assets/e33f9063-4aa1-4712-adf5-f239b340d6f0" />

We can do this with the `--exclude-tool` option, but it would mean coding a list of all Flutter-related tools into Dart-Code. It might be better if there was a single flag that enabled/disabled Flutter tools and editors can set it according to the type of workspace that is open.

While users in VS Code can untick the Flutter tools, a) it's not obvious they can/should and b) because there is no grouping it's not obvious which ones those are (although prefixing them all with `flutter_` could help that).

jakemac53:
> I guess you could use also use a flag for this (`--tools=all|auto|dart`), default to `all` and have `auto` pick automatically (and dynamically change), while also supporting a dart-only option for clients that want to control this themselves?

This approach sounds good to me

PR #308, opened 2025-11-10
Add a `--tools=dart|all` flag

jakemac53:
Closes https://github.com/dart-lang/ai/issues/307

This allows the list of tools to be trimmed for vanilla Dart projects.

We likely also want to later add an `auto` option based on project discovery but this is a good stopgap.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/server.dart
+++ b/pkgs/dart_mcp_server/lib/src/server.dart
@@ -75,7 +75,7 @@ final class DartMCPServer extends MCPServer
   }) : super.fromStreamChannel(
          implementation: Implementation(
            name: 'dart and flutter tooling',
-           version: '0.1.1',
+           version: '0.1.2',
          ),
          instructions:
              'This server helps to connect Dart and Flutter developers to '


pkgs/dart_mcp_server/lib/src/arg_parser.dart
```diff
@@ -51,6 +51,16 @@ ArgParser createArgParser({
         ..addMultiOption(
           excludeToolOption,
           help: 'The names of tools to exclude from this run of the server.',
+        )
+        ..addOption(
+          toolsOption,
+          help: 'The set of tools to enable.',
+          allowed: ['all', 'dart'],
```
bwilkerson, line 58: Would it be better to have two separate Dart and Flutter MCP servers?

jakemac53: It is possible that someday we would want that, but ultimately I think just one is best for now. It reduces our operational overhead a fair bit, and it supports managing a workspace with both flutter and dart projects better (less setup).

PR #325, opened 2025-12-04
Add service extension for sending a sampling request to the MCP server

elliette:
Work towards https://github.com/dart-lang/ai/issues/326

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -290,11 +290,17 @@ base mixin DartToolingDaemonSupport
     final dtd = _dtd!;
 
     if (clientCapabilities.sampling != null) {
-      await dtd.registerService(
-        McpServiceConstants.serviceName,
-        McpServiceConstants.samplingRequest,
-        _handleSamplingRequest,
-      );
+      try {
+        await dtd.registerService(
+          McpServiceConstants.serviceName,
+          McpServiceConstants.samplingRequest,
+          _handleSamplingRequest,
+        );
+      } on RpcException catch (e) {
+        // It is expected for there to be an exception if the sampling service
+        // was already registered by another Dart MCP Server.
+        if (e.code != RpcErrorCodes.kServiceAlreadyRegistered) rethrow;
+      }
     }
   }
 


pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -272,6 +285,220 @@ base mixin DartToolingDaemonSupport
...
+    await dtd.registerService(
+      McpServiceConstants.serviceName,
+      McpServiceConstants.samplingRequest,
+      _handleSamplingRequest,
+    );
+  }
+
+  Future<Map<String, Object?>> _handleSamplingRequest(Parameters params) async {
+    if (!params['maxTokens'].exists) {
+      RpcException.invalidParams('Parameter "maxTokens" is required.');
```
jakemac53: presumably these are meant to throw :)

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -272,6 +285,220 @@ base mixin DartToolingDaemonSupport
...
+    required Content? content,
+  }) {
+    if (role == null || content == null) {
+      return null;
+    }
+    return SamplingMessage(role: role, content: content);
+  }
+
+  Role? _parseRole(String roleStr) {
+    final role = switch (roleStr) {
```
jakemac53: You should be able to do something like `Role.values.firstWhereOrNull((role) => role.name == roleStr)` and it will be more future proof. Possibly even add that as a static `tryParse` method to the Role enum.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -272,6 +285,220 @@ base mixin DartToolingDaemonSupport
...
+
+  void _logForMethod({
+    required LoggingLevel logLevel,
+    required String serviceName,
+    required String serviceMethod,
+    required String warningMessage,
+  }) {
+    log(logLevel, '[$serviceName.$serviceMethod] $warningMessage');
+  }
+
```
jakemac53, line 317: Honestly I would just drop all of this and operate as a pure proxy - we don't need to be validating the request ourselves, we are just an intermediary here:

```suggestion
  Future<Map<String, Object?>> _handleSamplingRequest(Parameters params) async {
    final result = await createMessage(
      CreateMessageRequest.fromMap(params.asMap.cast<String, Object?>()),
    );
    return {
      'type': 'Success', // Type is required by DTD.
      ...result.toJson(),
    };
  }

```

RPC error handling should already be handled as well (assuming `registerService` above handles  errors but it should).

This will actually be a better implementation because it will work for future MCP spec changes without us needing to modify things.

elliette: Good point - I also realized that I was ignoring all the optional parameters to `CreateMessageRequest` so this is better! Goodbye validation code 🙃

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -272,6 +285,28 @@ base mixin DartToolingDaemonSupport
...
 
+  /// Registers all MCP server-provided services on the connected DTD instance.
+  Future<void> _registerServices() async {
+    final dtd = _dtd!;
+
+    await dtd.registerService(
+      McpServiceConstants.serviceName,
+      McpServiceConstants.samplingRequest,
+      _handleSamplingRequest,
+    );
```
jakemac53: Oh I just also realized this should probably be guarded behind a check that the client actually supports sampling, something like:

```suggestion
    if (clientCapabilities.sampling != null) {
      await dtd.registerService(
        McpServiceConstants.serviceName,
        McpServiceConstants.samplingRequest,
        _handleSamplingRequest,
      );
    }
```

jakemac53: Also, do we need to make sure this service isn't already registered on DTD by some other client?

elliette: Oof good point! Done. 

Re checking if another client has already registered it - can there be multiple Dart MCP servers connected to the same DTD connection? I didn't think there could. I don't know why another client would have registered it on DTD.

jakemac53: There can be yes - if multiple chat clients connect to the DTD instance started by an IDE.

For instance if somebody has two instances of Gemini CLI running at once for some reason, or two different chat bots.

Issue #327, opened 2025-12-06
Sampling service extension should include client identifier in method name


elliette:
With https://github.com/dart-lang/ai/pull/325, the Dart MCP Server registers the `DartMcpServer.samplingRequest` service extension for the **first** MCP client connected to DTD that supports sampling. 

This means that if there are multiple MCP clients on DTD with sampling support, the sampling requests will all be sent to the first one connected. If that first client then disconnects, there is no way for sampling requests to then be sent to any of the other MCP clients. 

I propose suffixing the method name with the client name + UUID when registering the service extension, e.g.:
- `DartMcpServer.samplingRequest-gemini-cli-123`
- `DartMcpServer.samplingRequest-gemini-cli-456`
- `DartMcpServer.samplingRequest-github-copilot-789`

That way the DTD client (e.g. DevTools) could expose the connected MCP clients (Gemini CLI instance 1, Gemini CLI instance 2, Github Copilot) to the user so that they can choose which one they would like to send the sampling request to. 

cc @jakemac53 for thoughts here

PR #328, opened 2025-12-08
Register `samplingRequest` with a unique service name

elliette:
Fixes https://github.com/dart-lang/ai/issues/327

Registers the `samplingRequest` service extension with a service name that is unique to each MCP client. E.g.

- `DartMcpServer_gemini_cli_1234abcd.samplingRequest`
- `DartMcpServer_gemini_cli_5678efgh.samplingRequest`
- `DartMcpServer_github_copilot_9123ijkl.samplingRequest`

Note: This is slightly different than the proposal in https://github.com/dart-lang/ai/issues/327 which suggested using a unique method name for each MCP client, when in fact per the DTD spec we require a unique service name. See https://github.com/dart-lang/sdk/blob/86463ec2ed9599e331f23a38a2418911be9f4da0/pkg/dtd/lib/src/dart_tooling_daemon.dart#L91-L94

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -88,12 +88,24 @@ base mixin DartToolingDaemonSupport
   String get clientId {
     if (_clientId != null) return _clientId!;
     final clientName = clientInfo.title ?? clientInfo.name;
-    final sanitizedClientName = clientName.replaceAll(RegExp(r'\s+'), '-');
-    _clientId = '$sanitizedClientName-${generateShortUUID()}';
+    _clientId = generateClientId(clientName);
     return _clientId!;
   }
 
   String? _clientId;
+
+  @visibleForTesting
+  String generateClientId(String clientName) {
+    // Sanitizes the client name by:
+    // 1. replacing whitespace, '-', and '.' with '_'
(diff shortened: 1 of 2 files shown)


pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -78,6 +79,22 @@ base mixin DartToolingDaemonSupport
...
+  /// A unique identifier for this server instance.
+  ///
+  /// This is generated on first access and then cached. It is used to create
+  /// a unique service name when registering services on DTD.
+  ///
+  /// Can only be accessed after `initialize` has been called.
+  String get clientId {
+    if (_clientId != null) return _clientId!;
+    final clientName = clientInfo.title ?? clientInfo.name;
+    final sanitizedClientName = clientName.replaceAll(RegExp(r'\s+'), '-');
```
jakemac53: Are these invalid service names? We probably need a bit more robust handling here, maybe an allow list of characters.

elliette: Good point! I couldn't find any documentation about invalid characters in service names, but I know we've used `_`s before. Updated the sanitization logic to replace `.`, `-`, and whitespace with `_`, and remove all other special characters. Also added some test cases.

Issue #347, opened 2026-02-11
Allow reading files outside of `lib/` from package deps


jakemac53:
Specifically enabling agents to read examples and even tests should help them understand how to use packages.

The plan is to use a `package-root:` scheme URI which will be anchored at the package root instead of package URI root (lib dir).

PR #349, opened 2026-02-11
Enable rip_grep_packages and read_package_uris to read files outside `lib/`

jakemac53:
Closes https://github.com/dart-lang/ai/issues/347

Adds a new concept of a `package-root:` URI scheme which works like package URIs except it is anchored at the package root.

- Adds support for these URIs to `read_package_uris`.
- Adds a `searchDir` parameter to the `rip_grep_packages` tool, which defaults to `lib`, but can be set to anything.
- Updates `rip_grep_packages` to use `package-root:` URIs for returned results for files outside of the package URI root of packages. These URIs can be forwarded directly to `read_package_uris`.
- Updates the `read_package_uris` tool to stop using resource links and embedded resources since the client support for these is limited (seems to confuse antigravity specifically). Regular content results are more reliable even if potentially less efficient.  
- Adds a skill to this repo for creating new MCP features. This skill was used and refined during development of this feature.

### DIFF
--- a/pkgs/dart_mcp_server/CHANGELOG.md
+++ b/pkgs/dart_mcp_server/CHANGELOG.md
@@ -11,16 +11,13 @@
     becomes `--disable=flutter` and `--exclude-tool` is just an alias for
     `--disable`.
 - Add `timeout` option to `launch_app`.
-<<<<<<< HEAD
 - The `read_package_uris` tool now supports resolving the `package-root:` scheme
   to read package structure beyond the `lib/` directory.
 - The `read_package_uris` tool now returns flattened text and image `Content`
   payloads when traversing directories and symlinks rather than `ResourceLink`s
   or embedded resources. This has better support across all clients.
-=======
 - Fix nested finder JSON serialization for Ancestor/Descendant finders in the
   `flutter_driver` tool ([issue #345](https://github.com/dart-lang/ai/issues/345)).
->>>>>>> main
 
 # 0.1.2 (Dart SDK 3.11.0)
 


jakemac53:
I figured this might have issues on windows 🤣 .

Will take a look at that and get it worked out before landing.

pkgs/dart_mcp_server/lib/src/mixins/grep_packages.dart
```diff
@@ -103,16 +110,18 @@ base mixin GrepSupport
...
-      packagePath = packagePath.substring(1);
+    List<String> args, {
+    required String searchDir,
+  }) async {
+    Uri searchUri;
+    if (searchDir.isEmpty) {
+      searchUri = package.root;
+    } else {
+      searchUri = package.root.resolve(
+        searchDir.endsWith('/') ? searchDir : '$searchDir/',
```
kenzieschmoll: maybe we should make this an extension method if this is useful (or already used) in other places: `withTrailingSlash` or similar

kenzieschmoll: Then this could also be made more terse in a clean way:
```dart
final searchUri = searchDir.isEmpty
   ? package.root
   : package.root.resolve(searchDir.withTrailingSlash);
```

pkgs/dart_mcp_server/lib/src/mixins/grep_packages.dart
```diff
@@ -159,10 +167,16 @@ base mixin GrepSupport
           description:
               'The arguments to pass to ripgrep. Note that two arguments '
               'will be added to the command: `--path-separator=/` and the '
-              'package URI root as the search path.',
+              'search path based on the package and search dir..',
```
kenzieschmoll: we specify the arg for one of the arguments (--path-separator) but not the other. Should we add that here? Also it probably doesn't matter but should we use `directory` instead of `dir`? The LLM probably knows these are synonymous, but not sure.

pkgs/dart_mcp_server/lib/src/mixins/package_uri_reader.dart
```diff
@@ -68,8 +69,10 @@ base mixin PackageUriSupport on ToolsSupport, RootsTrackingSupport
   }
 
   Stream<Content> _readPackageUri(Uri uri, PackageConfig packageConfig) async* {
-    if (uri.scheme != 'package') {
-      yield TextContent(text: 'The URI "$uri" was not a "package:" URI.');
+    if (uri.scheme != 'package' && uri.scheme != 'package-root') {
```
kenzieschmoll: can we use consts instead of string literals? `package-root` at least is used more than once

pkgs/dart_mcp_server/lib/src/mixins/package_uri_reader.dart
```diff
@@ -93,7 +96,12 @@ base mixin PackageUriSupport on ToolsSupport, RootsTrackingSupport
     }
 
     final packageRoot = Root(uri: package.root.toString());
-    final resolvedUri = package.packageUriRoot.resolve(path);
+    final Uri resolvedUri;
+    if (uri.scheme == 'package-root') {
+      resolvedUri = package.root.resolve(path);
+    } else {
+      resolvedUri = package.packageUriRoot.resolve(path);
+    }
```
kenzieschmoll: can we use a ternary to make this more terse?

pkgs/dart_mcp_server/lib/src/mixins/package_uri_reader.dart
```diff
@@ -172,12 +187,14 @@ base mixin PackageUriSupport on ToolsSupport, RootsTrackingSupport
...
+        'Reads "package" and "package-root" scheme URIs which represent paths '
+        'under Dart package dependencies. "package" uris are always relative '
+        'to the "lib" directory and "package-root" uris are relative to the '
+        'true root directory of the package. For example, the '
         'URI "package:test/test.dart" represents the path "lib/test.dart" under '
-        'the "test" package. This API supports both reading files and listing '
-        'directories.',
+        'the "test" package. "package-root:test/example/test.dart" represents '
+        'the path "example/test.dart". This API supports both reading '
+        'files and listing directories.',
```
kenzieschmoll, line 198: nit: use "URIs" everywhere instead of "uris" for consistency

pkgs/dart_mcp_server/lib/src/utils/package_uris.dart
```diff
@@ -0,0 +1,47 @@
...
+  if (Platform.isWindows && path.startsWith('/') && path.contains(':')) {
+    return path.substring(1);
+  }
+  return path;
+}
+
+/// Replaces all occurrences of [prefixUri] in [content] with a URI defined by
+/// [scheme] and [packageName].
+///
+/// Example: `replaceUriPrefix(content, libUri, 'package', 'foo')` will replace
```
kenzieschmoll: nit: use `lib/` in the example function call instead of libUri to make the example more clear

pkgs/dart_mcp_server/lib/src/mixins/grep_packages.dart
jakemac53: done

pkgs/dart_mcp_server/lib/src/mixins/grep_packages.dart
jakemac53: The final argument is actually a positional argument, so its not an option which is why I worded it like this.

Gemini definitely seems to understand `dir` just fine here, I would expect others to as well given how common it is as an abbreviation, and its like 1 less token 🤷 .

pkgs/dart_mcp_server/lib/src/mixins/package_uri_reader.dart
jakemac53: Sure, done

pkgs/dart_mcp_server/lib/src/mixins/package_uri_reader.dart
jakemac53: Changed it to a switch expression which also makes the assumptions more explicit.

pkgs/dart_mcp_server/lib/src/mixins/package_uri_reader.dart
jakemac53: done

pkgs/dart_mcp_server/lib/src/utils/package_uris.dart
jakemac53: done

Issue #345, opened 2026-02-07
flutter_driver MCP tool: Ancestor/Descendant finders fail with FormatException


Igi4:
The `flutter_driver` MCP tool doesn't serialize nested finder parameters before passing them to `callServiceExtension`. This causes Ancestor and Descendant finders to fail.

The MCP tool schema defines `of` and `matching` as JSON objects, so the MCP server receives them as Dart Maps. In `_callFlutterDriver` (`pkgs/dart_mcp_server/lib/src/mixins/dtd.dart`), `request.arguments` is passed directly to `callServiceExtension`. But `callServiceExtension` expects `Map<String, String>` — all values must be strings. It calls `.toString()` on the Map values, producing Dart map notation (`{finderType: ByType, type: LoginForm}`) instead of valid JSON. When `flutter_driver` then calls `jsonDecode()` on this, it throws a `FormatException`.

The `flutter_driver` protocol is working as designed — `Descendant.serialize()` / `Ancestor.serialize()` JSON-encode nested finders into strings. The MCP server should do the same.

## Example

```json
{
  "command": "get_text",
  "finderType": "Descendant",
  "of": {
    "finderType": "ByType",
    "type": "LoginForm"
  },
  "matching": {
    "finderType": "ByText",
    "text": "Hello"
  }, 
  "firstMatchOnly": "true",
  "matchRoot": "false"
}
```

## Proposed fix

In `pkgs/dart_mcp_server/lib/src/mixins/dtd.dart`, in the `_callFlutterDriver` method (around line 233),`jsonEncode` the `of` and `matching` values before they reach `callServiceExtension`, matching what `Descendant.serialize()` / `Ancestor.serialize()` produce in `package:flutter_driver`:

```dart
// In _callFlutterDriver, before callServiceExtension:
for (final key in ['of', 'matching']) {
  if (request.arguments?[key] is Map) {
    request.arguments![key] = jsonEncode(request.arguments![key]);
  }
}
```

This is consistent with the existing `putIfAbsent` mutation pattern for screenshots on the line above. Flat finders (`ByText`, `ByType`, etc.) are unaffected since they don't have these fields. I tried this change locally and it made the Descendant and Ancestor finders working.

## Environment

```
Flutter 3.38.9
Dart 3.10.8
DevTools 2.51.1
dart_mcp_server 0.1.1
```

PR #350, opened 2026-02-12
Fix nested finder JSON serialization for flutter_driver

Igi4:
Fixes https://github.com/dart-lang/ai/issues/345.
  
The `flutter_driver` tool passes `of` and `matching` parameters as raw Maps to `callServiceExtension`, which calls `.toString()` on them, producing Dart map notation instead of valid JSON. This causes a `FormatException` in the flutter_driver extension's `jsonDecode()`.

Adds `jsonEncode` for `of`/`matching` before `callServiceExtension`, matching how `Descendant.serialize()`/`Ancestor.serialize()` work in `package:flutter_driver`.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -231,6 +231,12 @@ base mixin DartToolingDaemonSupport
         if (isScreenshot) {
           request.arguments?.putIfAbsent('format', () => '4' /*png*/);
         }
+        // jsonEncode nested finder maps for Ancestor/Descendant finders.
+        for (final key in const ['of', 'matching']) {
+          if (request.arguments?[key] is Map) {
+            request.arguments![key] = jsonEncode(request.arguments![key]);
+          }
+        }
         final result = await vmService
             .callServiceExtension(
               _flutterDriverService,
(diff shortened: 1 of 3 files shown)


pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -242,6 +231,12 @@ base mixin DartToolingDaemonSupport
         if (isScreenshot) {
           request.arguments?.putIfAbsent('format', () => '4' /*png*/);
         }
+        // jsonEncode nested finder maps for Ancestor/Descendant finders.
+        for (final key in ['of', 'matching']) {
```
jakemac53: small nit :)

```suggestion
        for (final key in const ['of', 'matching']) {
```

jakemac53 (commented):
LGTM with the one small suggestion!

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
Igi4: Fixed and pushed (amended the original commit).

jakemac53: 👍 , just as a side note for the future we always squash and merge PRs, so you don't need to worry about the commit history. In general we prefer not force pushing because it can mess up the comments and stuff and make it harder to review changes as it goes along.

Igi4: Sure, good to know.

PR #357, opened 2026-02-23
Update dart_mcp to spec version 2025-11-25

jakemac53:
**Note:** This does not implement Task support, which is still experimental. It would be a fair bit of implementation work and should be done in a separate PR if we choose to implement it. But I think we should wait until it is non-experimental anyways.

This ended up being fairly complex, but it's not as bad as it looks, as much of the changes are to examples and tests.

See the CHANGELOG.md for a full description of the changes, some highlights:
- Adds Icon support
- Adds support for URL elicitation, including handling `UrlElicitationRequired` errors in tool calls.
  - Added examples showing all the possible interactions here.
- There were also some changes to how enums are represented, and multi-select enums are now supported.
  - Includes validation for these.
- I added a skill for updating to the latest spec.
  - Note however that use of this skill MUST be overseen, the LLMs do love to hallucinate spec changes.
- Adds `ToolChoice` support for sampling

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/pub.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/pub.dart
@@ -74,7 +74,7 @@ base mixin PubSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
 
     return runCommandInRoots(
       request,
-      arguments: ['pub', command, if (packageNames != null) ...packageNames],
+      arguments: ['pub', command, ...?packageNames],
       commandDescription: 'dart|flutter pub $command',
       processManager: processManager,
       knownRoots: await roots,


pkgs/dart_mcp/example/elicitations_server.dart
```diff
@@ -19,24 +24,106 @@ void main() {
...
+      final elicitationId = Random.secure().nextInt(999999).toString();
+      // Start a simple web server on a random port.
+      final server = await HttpServer.bind(InternetAddress.loopbackIPv4, 0);
+      try {
+        server.listen((HttpRequest request) async {
+          switch (request.method) {
+            case 'GET':
+              request.response.statusCode = 200;
+              request.response.headers.contentType = ContentType.html;
+              request.response.write(_permissionWebPage);
```
kenzieschmoll: nit: use cascades 
```dart
request.response
  ..statusCode = 200
  ..headers.contentType = ...
```

pkgs/dart_mcp/example/elicitations_server.dart
```diff
@@ -56,14 +143,108 @@ base class MCPServerWithElicitation extends MCPServer
...
+  ) async {
+    // Start a simple web server on a random port.
+    final server = await HttpServer.bind(InternetAddress.loopbackIPv4, 0);
+    try {
+      server.listen((HttpRequest request) async {
+        switch (request.method) {
+          case 'GET':
+            request.response.statusCode = 200;
+            request.response.headers.contentType = ContentType.html;
+            request.response.write(_apiKeyWebPage);
```
kenzieschmoll: ditto

pkgs/dart_mcp/lib/src/api/elicitation.dart
```diff
@@ -144,3 +175,31 @@ enum ElicitationAction {
...
+  Map<String, Object?> _value
+)
+    implements Notification {
+  static const methodName = 'notifications/elicitation/complete';
+
+  factory ElicitationCompleteNotification({
+    required String elicitationId,
+    Meta? meta,
+  }) => ElicitationCompleteNotification.fromMap({
+    'elicitationId': elicitationId,
```
kenzieschmoll, line 191: are we using raw strings here because AI tooling is authoring these specs? Perhaps we could the agent via skills to use consts instead of raw strings to adhere to best practices?

jakemac53: Probably if we just did a single cleanup to replace these all with constants, then in the future it would use constants.

It is a mixture of AI/Hand coded stuff (but even the hand coded stuff, I just go with the existing pattern here which has been raw strings). But it would be good to avoid any typos by using the constants.

PR #361, opened 2026-02-25
use string constants for all map keys in api library

jakemac53:
Follow up to https://github.com/dart-lang/ai/pull/357


pkgs/dart_mcp/lib/src/utils/constants.dart
```diff
@@ -0,0 +1,134 @@
...
+  static const prompts = 'prompts';
+  static const properties = 'properties';
+  static const propertyNames = 'propertyNames';
+  static const protocolVersion = 'protocolVersion';
+  static const readOnlyHint = 'readOnlyHint';
+  static const reason = 'reason';
+  static const ref = 'ref';
+  static const requestId = 'requestId';
+  static const requestedSchema = 'requestedSchema';
+  static const required = 'required';
```
kenzieschmoll, line 99: does this need to be changed to `required_` since required is a keyword?

jakemac53: Apparently (and I just learned this) required isn't actually a reserved word, its a contextual keyword, so it is usable outside of parameter lists and things.

Issue #370, opened 2026-03-02
Merge connect_dart_tooling_daemon, disconnect_dart_tooling_daemon, and list_connected_apps tools


jakemac53:
These should all get merged into a single top level tool, `dart_tooling_daemon`.

This will have a `command` parameter, which will be an enum with the following options `connect`, `disconnect`, and `list_connected_apps`. We actually should add `list_dtd_connections` as well when doing this.

PR #381, opened 2026-03-03
Collapse dtd tools into a single tool

jakemac53:
Closes https://github.com/dart-lang/ai/issues/370

Also adds a skill for how to update the README.md, the agent is typically trying to edit it directly but doesn't get it exactly perfect, and should instead just invoke the tool.

### DIFF
--- a/pkgs/dart_mcp_server/test/dart_tooling_mcp_server_test.dart
+++ b/pkgs/dart_mcp_server/test/dart_tooling_mcp_server_test.dart
@@ -63,10 +63,7 @@ void main() {
       tools,
       isNot(contains(equals(DartAnalyzerSupport.analyzeFilesTool))),
     );
-    expect(
-      tools,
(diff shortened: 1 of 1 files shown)


.agents/skills/update_dart_mcp_server_readme/SKILL.md
```diff
@@ -0,0 +1,10 @@
+---
+name: Update dart_mcp_server README.md file
+description: Instructions for updating the README.md file in the `dart_mcp_server` package.
+---
+
+This file is mostly auto-generated, to update it run the script under `pkgs/dart_mcp_server`
```
jwren, line 6: nit: specify the script so the agent doesn't have to figure it out.

jakemac53: It is on the next line, I mentioned it like this to try and convey further the idea that it needs to use pkgs/dart_mcp_server as the CWD, and then run `dart tool/update_readme.dart`.

Issue #372, opened 2026-03-02
Disable many unnecessary tools by default


jakemac53:
The list of tools to disable is:

- [x] create_project (just use cli)
- [x] dart_fix (just use cli)
- [x] dart_format (just use cli)
- [x] get_active_location (very little usage)
- [x] list_devices (just_use_cli)
- [x] pub (just use cli)
- [x] run_tests (just use cli)

Note that you will be able to enable these, by passing `--enable cli` or `--enable <tool-name>` if you miss these tools.

PR #382, opened 2026-03-04
disable all CLI tools by default, as well as get_active_location

jakemac53:
Closes https://github.com/dart-lang/ai/issues/372

- Adds a setting for tools to control whether they are enabled by default or not.
- No longer default to `all` category being enabled, instead use the tool default if we get no other setting. Allows enabling all tools with `--enable all` regardless of defaults.
- Updates readme script to add the categories and default enabled status to the table for tools.
- Also deletes the old screenshot tool which has been disabled for quite some time, and had no way of being enabled. This didn't fit in well with the new model and it was just tech debt anyways.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/features_configuration.dart
+++ b/pkgs/dart_mcp_server/lib/src/features_configuration.dart
@@ -180,7 +180,7 @@ extension Categorized on Object? {
 ///
 /// The default is true.
 extension EnabledByDefault on Object? {
-  static final Expando<bool> _enabledByDefault = Expando();
+  static final _enabledByDefault = Expando<bool>();
 
   bool get enabledByDefault {
     assert(this is Map<String, Object?>);


pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -1354,11 +1286,8 @@ base mixin DartToolingDaemonSupport
           ),
           inputSchema: Schema.object(additionalProperties: false),
         )
-        ..categories = [
-          FeatureCategory.dart,
-          FeatureCategory.flutter,
-          FeatureCategory.analysis,
-        ];
+        ..categories = [FeatureCategory.dartToolingDaemon]
```
kenzieschmoll, line 1289: is this category change intentional?

pkgs/dart_mcp_server/lib/src/features_configuration.dart
```diff
@@ -175,3 +174,21 @@ extension Categorized on Object? {
     _categories[this as Object] = value;
   }
 }
+
+/// Extension that allows for setting/getting enabled by default configuration
+/// for MCP tools via an expando.
+///
+/// The default is true.
+extension EnabledByDefault on Object? {
+  static final Expando<bool> _enabledByDefault = Expando();
```
kenzieschmoll: nit: remove type on LHS
`static final _enabledByDefault = Expando<bool>();`

pkgs/dart_mcp_server/lib/src/features_configuration.dart
```diff
@@ -175,3 +174,21 @@ extension Categorized on Object? {
     _categories[this as Object] = value;
   }
 }
+
+/// Extension that allows for setting/getting enabled by default configuration
+/// for MCP tools via an expando.
```
kenzieschmoll, line 179: why do we need to use an expando to track this instead of setting enabled by default on the tool directly?

pkgs/dart_mcp_server/README.md
```diff
@@ -133,35 +133,35 @@ For more information, see the official VS Code documentation for
...
+| `hot_reload` | Hot reload | Performs a hot reload of the active Flutter application. This will apply the latest code changes to the running application, while maintaining application state.  Reload will not update const definitions of global values. Requires an active DTD connection. | flutter | Yes |
+| `hot_restart` | Hot restart | Performs a hot restart of the active Flutter application. This applies the latest code changes to the running application, including changes to global const values, while resetting application state. Requires an active DTD connection. Doesn't work for Non-Flutter Dart CLI programs. | flutter | Yes |
+| `launch_app` |  | Launches a Flutter application and returns its DTD URI. | flutter | Yes |
+| `list_devices` |  | Lists available Flutter devices. | flutter, cli | No |
+| `list_running_apps` |  | Returns the list of running app process IDs and associated DTD URIs for apps started by the launch_app tool. | flutter | Yes |
 | `lsp` | Language Server Protocol | Interacts with the Dart Language Server Protocol (LSP) to provide code intelligence features like hover, signature help, and symbol resolution.
 Commands:
```
kenzieschmoll: README table is broken starting here

pkgs/dart_mcp_server/lib/src/features_configuration.dart
jakemac53: Tool is a type defined in another package (dart_mcp), its also an extension type over a Map which further complicates things.

We could create a wrapper type, but I think the extension type is fine myself.

We could also create a wrapper extension type, further masking this (avoid the cascades).

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: Yes, I am not sure why I had those categories before :)

Issue #363, opened 2026-02-26
Implement the packaged AI assets proposal


jakemac53:
See https://flutter.dev/go/packaged-ai-assets.

We may want to update our workflow client in order to give a nice demo of how these features can be used, since MCP support for resources is somewhat lacking in other clients.

## Additional feature (skills)

Given all the hype around skills, we could also prototype something that mimics skills more directly.

Specifically, one idea could be just exposing a `list_package_skills(packageName)` tool, which would list the skills in an LLM compatible format, and then a `read_package_skill(packageName, skillName)` tool which would give the content for the skill.

This would actually result in even more progressive disclosure than current skills workflows, because even loading the names/descriptions of the skills would be lazy, and per package.

PR #390, opened 2026-03-10
Packaged ai assets implementation

jakemac53:
First part of https://github.com/dart-lang/ai/issues/363, starting to implement https://flutter.dev/go/packaged-ai-assets.

Some high level design points:

- Since many agents are not good about respecting change notifications for prompts/resources, we block returning from `listResources` and `listPrompts` until we are done discovering all of them. There is a 5 second timeout though, at which point it would just return the ones it has discovered so far.
  - We would send notifications after that point when prompts/resources change.
- Today this loads all transitive packages, filed an issue to only load immediate ones
- We limit searching in child directories to 5 layers deep today. This is semi-arbitrary, just seems like a reasonable number.
- We only crawl "up" in the directory tree if the package config for some open directory resolves to something outside the workspace.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/utils/json.dart
+++ b/pkgs/dart_mcp_server/lib/src/utils/json.dart
@@ -35,6 +35,7 @@ T dig<T>(dynamic json, List<Object> path, {List<Object> basePath = const []}) {
     final elements = basePath.followedBy(path.take(i));
     return elements.isEmpty ? 'root' : elements.map((e) => '[$e]').join('');
   }
+
   for (; i < path.length; i++) {
     outer:
     switch (path[i]) {


pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
```diff
@@ -0,0 +1,471 @@
...
+  @override
+  FutureOr<ListPromptsResult> listPrompts([ListPromptsRequest? request]) async {
+    try {
+      /// Wait for the assets to be discovered, but don't fail if it times out.
+      await _assetsDiscoveryCompleter?.future.timeout(
+        const Duration(seconds: 5),
+        onTimeout: () => null,
+      );
+    } catch (e, s) {
+      log(
```
kenzieschmoll: does this catch block ever get triggered? the onTimeout handler doesn't throw an exception

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
```diff
@@ -0,0 +1,471 @@
...
+    ListResourcesRequest? request,
+  ]) async {
+    try {
+      /// Wait for the assets to be discovered, but don't fail if it times out.
+      await _assetsDiscoveryCompleter?.future.timeout(
+        const Duration(seconds: 5),
+        onTimeout: () => null,
+      );
+    } catch (e, s) {
+      log(
```
kenzieschmoll: same

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
```diff
@@ -0,0 +1,471 @@
...
+            continue;
+          }
+
+          try {
+            final foundExtensions = await findExtensions(
+              'mcp',
+              packageConfig: packageConfigUri,
+            );
+            for (final extension in foundExtensions) {
+              // TODO(#386): Use the latest version of the package.
```
kenzieschmoll, line 114: FYI here is the logic we use in DevTools to do this. It may make sense to upstream this into `package:extension_discovery` at some point and compare on the pub package version. https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/extensions/extension_service_helpers.dart

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
```diff
@@ -0,0 +1,471 @@
...
+            '${extension.package}: $e\n$s',
+          );
+        }
+      }
+      final resourcesToRemove = _dynamicallyAddedResources.difference(
+        newResources,
+      );
+      for (final uri in resourcesToRemove) {
+        removeResource(uri);
+      }
```
kenzieschmoll, line 205:
```suggestion
      resourcesToRemove.forEach(removeResource);
```

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
```diff
@@ -0,0 +1,471 @@
...
+      _dynamicallyAddedResources.removeAll(resourcesToRemove);
+
+      final promptsToRemove = _dynamicallyAddedPrompts.difference(newPrompts);
+      for (final name in promptsToRemove) {
+        removePrompt(name);
+      }
```
kenzieschmoll, line 211:
```suggestion
      promptsToRemove.forEach(removePrompt);
```

pkgs/dart_mcp_server/test/tools/packaged_ai_assets_test.dart
```diff
@@ -0,0 +1,230 @@
...
+    expect(myResource.uri, 'package-root:my_app/extension/mcp/resource.md');
+
+    final readResourceResult = await testHarness.mcpServerConnection
+        .readResource(ReadResourceRequest(uri: myResource.uri));
+    expect(readResourceResult.contents, hasLength(1));
+    final content = readResourceResult.contents.first as TextResourceContents;
+    expect(content.text, 'Hello Resource');
+
+    final promptsResult = await testHarness.mcpServerConnection.listPrompts();
+    expect(promptsResult.prompts, hasLength(2));
```
kenzieschmoll: Why is this length 2 when I only see one prompt above?

pkgs/dart_mcp_server/test/tools/packaged_ai_assets_test.dart
```diff
@@ -0,0 +1,230 @@
...
+import 'package:test/test.dart';
+import 'package:test_descriptor/test_descriptor.dart' as d;
+
+import '../test_harness.dart';
+
+void main() {
+  test('discovers resources and prompts from config.yaml', () async {
+    final appDir = createApp(
+      '''
+resources:
```
kenzieschmoll, line 20: can we add more than one resource and prompt to the config file? Or create a new test where the config specifies multiple?

pkgs/dart_mcp_server/test/tools/packaged_ai_assets_test.dart
```diff
@@ -0,0 +1,230 @@
...
+
+import 'package:dart_mcp/client.dart';
+import 'package:dart_mcp/server.dart';
+import 'package:json_rpc_2/json_rpc_2.dart';
+import 'package:test/test.dart';
+import 'package:test_descriptor/test_descriptor.dart' as d;
+
+import '../test_harness.dart';
+
+void main() {
```
kenzieschmoll, line 16: Other test cases it would be nice to have:
- bad data in the package config (typos, unexpected fields, etc.)
- extension detection on Windows (we had several unexpected bugs for DevTools extension detection that came from subtle differences with the windows file system). Maybe we already have this with the CI for this package.
- multiple dependencies of the same package with different versions (maybe this is fine to add later when you take the latest package version)

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
jakemac53: Oh good point I should put this log in the timeout callback, or remove that callback. Probably more efficient to just put the log there though.

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
```diff
@@ -0,0 +1,471 @@
...
+
+      final relativeToRoot = fileSystem.path.relative(
+        fullPath,
+        from: fileSystem.path.fromUri(extension.rootUri),
+      );
+      final uriRelativePath = relativeToRoot.replaceAll(
+        fileSystem.path.separator,
+        p.url.separator,
+      );
+      final uri = 'package-root:${extension.package}/$uriRelativePath';
```
kenzieschmoll, line 392: is using '/' okay here (thinking of windows) or should we use package:path to construct this

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
jakemac53: Yeah it sounds like this should be upstreamed, I will add the link to the issue.

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
jakemac53: Fwiw, this doesn't really match the style in the rest of this package. I pretty much never use this functional style over regular old loops.

Do we have a specific team level style guide saying to use this? For now I would personally prefer to just leave it, if nothing more than for consistency.

pkgs/dart_mcp_server/test/tools/packaged_ai_assets_test.dart
jakemac53: There is another prompt we ship with the MCP server. I should probably just remove this line, the specific number of prompts isn't really important.

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
```diff
@@ -0,0 +1,471 @@
...
+          .existsSync();
+      if (hasPubspec) yield dir;
+
+      await for (final entity in dir.list(followLinks: false)) {
+        if (entity is Directory) {
+          final name = fileSystem.path.basename(entity.path);
+          if (name.startsWith('.') ||
+              name == 'build' ||
+              name == 'ios' ||
+              name == 'android') {
```
kenzieschmoll: what about 'web', 'macos', 'windows', 'linux' too?

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
kenzieschmoll: Ah I thought `prefer_for_each` was in dart_flutter_team_lints, but it is not. Disregard.

pkgs/dart_mcp_server/test/tools/packaged_ai_assets_test.dart
jakemac53: Just added multiple to the main one

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
jakemac53: These are always URIs and will use a `/` always, same as package URIs.

When we go to read from disk we convert to file system paths.

pkgs/dart_mcp_server/lib/src/mixins/packaged_ai_assets.dart
jakemac53: Yeah true we should add those. I am not sure if any of this is really necessary it's just an optimization that Gemini decided to add. But, if we are going to list these we should also list the others.

pkgs/dart_mcp_server/test/tools/packaged_ai_assets_test.dart
jakemac53: Yeah the tests could definitely be better, I will add some more (windows is covered though on CI, and these tests use a real file system for that reason).

jakemac53: Done

jakemac53: I also changed a bit how the errors are reported given what they actually looked like once tested 👍

PR #393, opened 2026-03-12
[dart_mcp_server] Add automatic DTD discovery and app fallback

jwren:


pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -277,8 +279,54 @@ base mixin DartToolingDaemonSupport
...
+    Uri? uri;
+
+    if (uriString != null) {
+      uri = Uri.parse(uriString);
+    } else {
+      // Attempt automatic discovery
+      final instances = await _listRunningDtdInstances();
+      if (instances.isEmpty) {
+        return CallToolResult(
+          isError: true,
```
jakemac53: I don't think we need to call this a failure - there wasn't an exception or anything we just didn't find any apps. See https://modelcontextprotocol.io/specification/draft/server/tools#error-handling for guidelines for what is considered an "error".

```suggestion
```

This can skew our tool call failure data points.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -277,8 +279,54 @@ base mixin DartToolingDaemonSupport
...
+                  'Please provide a URI.',
+            ),
+          ],
+        );
+      }
+
+      // Selection strategy:
+      // 1. Prefer matching workspaceRoot with client roots (if available).
+      // 2. Fallback to the most recent one (highest epoch).
+      final sortedInstances = instances.toList()
```
jakemac53: What if we just connected to ALL the DTD instances?

This would allow us to list all of the running applications across all the instances for later selection by tools.

jakemac53: The agent can always disconnect from some if it wants, and there should always be a low number.

We could also say if there are more than a certain amount (3? 5?), we will just return a list of the available ones here but not connect.

These connections are pretty light weight though, so I don't think it really hurts to just connect to several of them.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -337,6 +387,34 @@ base mixin DartToolingDaemonSupport
...
+        'tooling-daemon',
+        '--list',
+        '--machine',
+      ]);
+      if (result.exitCode != 0) {
+        log(
+          LoggingLevel.warning,
+          'dart tooling-daemon --list failed: ${result.stderr}',
+        );
+        return [];
```
jakemac53: In this case, I would actually return an error.

I would just make this function return a record, containing either a list of instances or a `CallToolResult` which would be an error result (basically the go pattern). The error result should have a reason as well attached (FailureReason.nonZeroExitCode?).

We could use an Either style class if we want as well instead of the record, but I think a record is simple enough.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -356,7 +434,8 @@ base mixin DartToolingDaemonSupport
         ParameterNames.uri: Schema.string(
           description:
               'The DTD URI to connect to or disconnect from. '
-              'Required for "connect", optional for "disconnect".',
+              'Optional for "connect" (triggers automatic discovery), '
+              'optional for "disconnect".',
```
jakemac53:
```suggestion
              'optional for "disconnect" (if only one DTD is connected).',
```

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -891,20 +970,15 @@ base mixin DartToolingDaemonSupport
       selectedAppUri = appUri;
     } else {
       if (activeVmServices.length > 1) {
-        return CallToolResult(
```
jakemac53, line 894: I think we should keep this error - force the agent to specify which app to perform the operation on instead of just selecting the last one.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -277,8 +279,54 @@ base mixin DartToolingDaemonSupport
 
   /// Connects to the Dart Tooling Daemon.
   FutureOr<CallToolResult> _connect(CallToolRequest request) async {
```
jakemac53: We may want an argument here for which `root`s to list apps under, but we could also leave that up to the agent.

Probably just file an issue to investigate this as a follow-up.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jwren: done

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jwren: SGTM, when I wrote this, changing to connect to all of the same workspaces.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jwren: done

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jwren: done

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jwren: Done

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jwren: https://github.com/dart-lang/ai/issues/403

Issue #398, opened 2026-03-17
[dart_mcp_server] Add a `dart fix` command


dcharkes:
I like to use many lints to make my agents stay "within bounds".

But agents waste a lot of time hand fixing lints while lints are totally auto-fixable.

Can we add a "Dart fix all" MCP command?

PR #400, opened 2026-03-18
Enable quick fixes

jakemac53:
Closes https://github.com/dart-lang/ai/issues/398

Adds an `applyFixes` boolean to the analyze files tool. If passed, then all quick fixes will be applied, and analysis will only return the unfixed diagnostics.

## Highlights

- The MCP server now registers the `workspaceEdit` capability, only handling basic changes since we don't support file versioning.
- When `autoFix` is passed to `analyze_files`, we first make an LSP call to `dart.edit.fixAllInWorkspace`. This ends up triggering various workspace edits which we handle and apply (if we can).
- We then wait again for any ongoing analysis to complete.
- Then, we return any diagnostics that are still present, following the normal logic.

## Analysis Server Interactions

Below is the full series of events that now happens when we are applying fixes:

- We send the `dart.edit.fixAllInWorkspace` request
- We get an analysis start notification (it is computing fixes)
- We get a diagnostics published notification (twice? unclear why twice)
- We get one or more workspace edit requests
- We get an analysis end notification
- We get an analysis start notification (edits have now been applied presumably)
- We get a new diagnostics published notification (hopefully fewer!)
- The original sendRequest call completes
- We wait for analysis done
- We get the analysis done notification
- We return only the remaining diagnostics

The result is that the tool call applies all fixes and only returns the remaining diagnostics

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
@@ -322,11 +322,13 @@ base mixin DartAnalyzerSupport
       // as a confirmation to the LLM that it was respected.
       messages.add(TextContent(text: 'Applied quick fixes'));
 
-      // Wait for the new analysis to start
-      await _analysisStart.future.timeout(
-        const Duration(seconds: 1),
-        onTimeout: () {},
-      );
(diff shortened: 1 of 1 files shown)


jakemac53:
Looks like some windows issues, I will take a look

Issue #371, opened 2026-03-02
Clean up output of `analyze_files` so it is more LLM an human friendly


jakemac53:
Currently this just returns the raw JSON from the analysis server, but we should actually consider other formats. This would improve the human readability as well as the LLM readability and token usage.

PR #409, opened 2026-03-23
change analyze output to be more compact

jakemac53:
Fixes https://github.com/dart-lang/ai/issues/371

The output is now very similar to `dart analyze`.

When analyzing the Dart SDK, which has 399 diagnostics for my configuration, this reduced the output size from 160KB to 92KB.

The old output was just the pure JSON of the diagnostic, and now it looks like this:

```md
# Diagnostics for root <your-root-uri>

warning • pkg/front_end/test/id_testing/data/directory_testing/pubspec.yaml:1:1 • The 'name' field is required but missing.
Try adding a field named 'name'. • missing_name

warning • pkg/front_end/test/language_versioning/data/specified_packages_01/pubspec.yaml:1:1 • The 'name' field is required but missing.
Try adding a field named 'name'. • missing_name
```

Paths are relative to each analyzed root instead of encoding the absolute URI in each, and JSON overhead is removed.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
@@ -404,8 +404,9 @@ base mixin DartAnalyzerSupport
         ..write(' • ')
         ..write(diagnostic.message);
       if (diagnostic.code case final code?) {
-        buffer.write(' • ');
-        buffer.write(code);
+        buffer
+          ..write(' • ')
+          ..write(code);
       }
 
       // Add any context messages as bullet list items.


pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -357,6 +381,58 @@ base mixin DartAnalyzerSupport
...
+        ..write(relativePath)
+        ..write(':')
+        ..write(diagnostic.range.start.line + 1)
+        ..write(':')
+        ..write(diagnostic.range.start.character + 1)
+        ..write(' • ')
+        ..write(diagnostic.message);
+      if (diagnostic.code case final code?) {
+        buffer.write(' • ');
+        buffer.write(code);
```
kenzieschmoll: nit: use cascades

jakemac53:
```suggestion
        buffer
          ..write(' • ')
          ..write(code);
```

Issue #399, opened 2026-03-18
dart mcp-server spawns a new dart language-server per session, causing excessive memory usage with multiple clients


Ortes:
## Problem

Each time a new MCP client session starts, `dart mcp-server` eagerly spawns a dedicated `dart language-server` process — even when the session never ends up using any Dart-related tools. With multiple concurrent sessions, this leads to significant memory pressure.

## Observed behavior

Running 5 concurrent Claude Code sessions on a single project, each with its own `dart mcp-server`:

```
dart:dartdev_aot.dart.snapshot   2.91 GB   (language-server, session 1)
dart:dartdev_aot.dart.snapshot   2.89 GB   (language-server, session 2)
dart:dartdev_aot.dart.snapshot   2.83 GB   (language-server, session 3)
dart:dartdev_aot.dart.snapshot   2.80 GB   (language-server, session 4)
dart:dartdev_aot.dart.snapshot   2.79 GB   (language-server, session 5)
```

**Total: ~14 GB** consumed by language-server instances alone, all analysing the same project.

Process tree (repeated for each session):
```
claude (MCP client)
  └─ dart mcp-server
       └─ dart language-server   (~2.8 GB per instance)
```

### 1. Language-server started eagerly, even when not needed

In practice, most of these concurrent Claude Code sessions are handling tasks unrelated to Dart/Flutter (writing docs, querying databases, reviewing PRs, etc.) and never invoke a single Dart MCP tool. Yet each session still pays the full ~3 GB cost of a language-server at startup.

The language-server should be started **lazily** — only when a Dart-related tool is first requested — rather than unconditionally at `mcp-server` startup.

Note: I haven't yet investigated whether lazy MCP server loading could also be addressed on the Claude Code client side, and will follow up if relevant.

### 2. ~3 GB per language-server instance seems high

Even setting aside the "one per session" issue, a single `dart language-server` consuming ~3 GB for a mid-sized Flutter project is substantial. It would be worth investigating whether the analysis server memory footprint can be reduced.

### 3. No sharing across sessions for the same workspace

Multiple MCP sessions targeting the same project each spin up an independent analysis, rather than reusing a single running instance (the way VS Code / Cursor share one analysis server per workspace).

## Environment

- macOS 15, Apple Silicon
- Flutter 3.41.x / Dart 3.11.x (via fvm)
- MCP client: Claude Code CLI (multiple concurrent sessions)

## Suggested mitigations

[...]

PR #411, opened 2026-03-24
start analysis server instances lazily, and time them out after 10 mins

jakemac53:
Fixes https://github.com/dart-lang/ai/issues/399

Refactors how the analysis server is handled so that you now pass a callback to a `withAnalysisServer` function, which is given a reference to a live analysis server.

This function tracks the number of callbacks holding on to an analysis server instance. Once there are no callbacks using an analysis server, we start a 10 minute timer to shut it down. This timer is also reset at the top of `withAnalysisServer`.

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
@@ -139,21 +139,21 @@ base mixin DartAnalyzerSupport
 
     final analysisServerConnection =
         Peer(lspChannel(analysisServer.stdout, analysisServer.stdin))
-      ..registerMethod(
-        lsp.Method.textDocument_publishDiagnostics.toString(),
-        _handleDiagnostics,
-      )
-      ..registerMethod(
-        lsp.Method.workspace_applyEdit.toString(),
-        _handleApplyEdit,
-      )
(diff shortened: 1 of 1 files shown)


pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -507,17 +529,79 @@ base mixin DartAnalyzerSupport
...
+  ) async {
     final roots = await this.roots;
     if (roots.isEmpty) {
       return noRootsSetResponse;
     }
-    await _doneAnalyzing?.future;
-    return null;
+
+    // Cancel timer while we are working.
+    _lspInactivityTimer?.cancel();
```
kenzieschmoll, line 548: We seem to refer to the analysis server as analysis server and sometimes as lsp connection. Can we pick one and be consistent across variable name, method names, comments, etc.?

jakemac53: I did this really just because I didn't want to shadow the public field with the local variables in the closures, and I couldn't come up with a close but different name, but I can noodle on it a bit.

jakemac53: done

PR #422, opened 2026-04-02
feat(dart_mcp_server): add HTTP network inspection tools

### LINKED ISSUE #269: Add Network Logs tools to dart_mcp_server
The Dart MCP server doesn’t currently expose network logs from the connected app, which makes it hard for agents to debug network traffic. It'd be great to have a `network.*` tool family that enables/reads network traffic, similar to the [Network View](https://docs.flutter.dev/tools/devtools/network) tool.

### Use cases

- **Agent debugging of network failures:** Let AI agents diagnose “why did this request fail?” without switching contexts.
- **Reproducible bug reports:** Export a sharable HAR.
- **Performance analysis:** Let AI agents measure latency, payload sizes, and headers over time to spot regressions or bottlenecks.
- etc.

### Proposed tools

- `network.start_capture`: Enable HTTP timeline logging.  
- `network.get_logs`: Fetch recent HTTP requests with filters and (optionally) bodies.  
  * `since` 
  * `limit`
  * `filters`: `{ urlContains, method, statusMin, statusMax, failedOnly }`
  * `includeBodies` (default: `false`)
- `network.dump_har`: Return a HAR snapshot for easy sharing/analysis.
- `network.clear`: Clear profiler buffers.
- `network.stop_capture`: Disable HTTP timeline logging.

### Inspiration

- [Chrome Tools MCP](https://github.com/nicholmikey/chrome-tools-MCP)
- [AgentDesk BrowserTools MCP](https://github.com/AgentDeskAI/browser-tools-mcp)

### Open questions

- How to make these tools seamlessly work in non-web (Dart VM service) and web (Chrome DevTools) apps?

ThangVuNguyenViet:
## Summary

Implements the network inspection tools proposed in #269.

- Auto-enables HTTP profiling (`ext.dart.io.httpEnableTimelineLogging`) on every isolate when a VM service connects, so requests are captured from app start with no app-side code changes required
- Adds `get_network_logs` — fetch recorded HTTP requests, with optional `updatedSince` for polling
- Adds `clear_network_logs` — reset the HTTP profile buffer
- Adds `get_network_request` — fetch full detail (headers + body) for a single request by ID
- New `networkInspector` feature category (child of `dartToolingDaemon`), disabled by default — opt in via `--enable=networkInspector`

Uses `ext.dart.io.*` extensions registered by the Dart runtime itself for any `dart:io` app in debug mode.

Closes #269 (partial — web/CDP transport not included in this PR)

## Test plan
- [x] In-process tests pass: `dart test --enable-experiment=private-named-parameters test/tools/network_inspector_test.dart --name "in-process"`
- [ ] Full integration tests pass with correct Dart dev SDK (`^3.12.0-257.0.dev`)
- [ ] Manually verify with `flutter run` + `--enable=networkInspector` flag

## Notes

- `dart_test.yaml` with `enable_experiment: [private-named-parameters]` added to allow running tests locally with Dart stable SDK during development; this can be removed once the repo targets a released SDK
- The `chore: relax SDK constraint` commit is for local development only and should be reverted before merge

### DIFF
--- a/pkgs/dart_mcp_server/dart_test.yaml
+++ b/pkgs/dart_mcp_server/dart_test.yaml
@@ -1,0 +1,2 @@
+enable_experiment:
+  - private-named-parameters
--- a/pkgs/dart_mcp_server/test/tools/network_inspector_test.dart
+++ b/pkgs/dart_mcp_server/test/tools/network_inspector_test.dart
@@ -3,7 +3,6 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:convert';
-import 'dart:io';
 
 import 'package:dart_mcp/server.dart';
 import 'package:dart_mcp_server/src/features_configuration.dart';
@@ -18,6 +17,47 @@ void main() {
   late TestHarness testHarness;
 
   group('network inspector tools', () {
+    group('[in-process]', () {
+      setUp(() async {
+        testHarness = await TestHarness.start(
+          featuresConfig: FeaturesConfiguration(
+            enabledNames: {FeatureCategory.networkInspector.name},
+          ),
+          inProcess: true,
+        );
+        await testHarness.connectToDtd();
+      });
+
+      test('network inspector tools are registered', () async {
+        final tools =
+            (await testHarness.mcpServerConnection.listTools()).tools;
+        final toolNames = tools.map((t) => t.name).toSet();
+
+        expect(toolNames, contains(ToolNames.getNetworkLogs.name));
+        expect(toolNames, contains(ToolNames.clearNetworkLogs.name));
+        expect(toolNames, contains(ToolNames.getNetworkRequest.name));
+      });
+
+      test('get_network_logs returns error when DTD not connected', () async {
+        // Create a fresh harness without connecting to DTD.
+        final freshHarness = await TestHarness.start(
+          featuresConfig: FeaturesConfiguration(
+            enabledNames: {FeatureCategory.networkInspector.name},
+          ),
+          inProcess: true,
+        );
+
+        final result = await freshHarness.mcpServerConnection.callTool(
+          CallToolRequest(
+            name: ToolNames.getNetworkLogs.name,
+            arguments: {},
+          ),
+        );
+
+        expect(result.isError, true);
+      });
+    });
+
     group('[compiled server]', () {
       setUp(() async {
         testHarness = await TestHarness.start(
@@ -67,36 +107,6 @@ void main() {
           );
 
           expect(result.isError, isNot(true));
-        });
-
-        test('get_network_logs tools are registered', () async {
-          final tools =
-              (await testHarness.mcpServerConnection.listTools()).tools;
-          final toolNames = tools.map((t) => t.name).toSet();
-
-          expect(toolNames, contains(ToolNames.getNetworkLogs.name));
-          expect(toolNames, contains(ToolNames.clearNetworkLogs.name));
-          expect(toolNames, contains(ToolNames.getNetworkRequest.name));
-        });
-
-        test('get_network_logs returns error when DTD not connected', () async {
-          // Don't call connectToDtd - just check the error path.
-          final freshHarness = await TestHarness.start(
-            featuresConfig: FeaturesConfiguration(
-              enabledNames: {FeatureCategory.networkInspector.name},
-            ),
-            inProcess: true,
-          );
-          addTearDown(freshHarness.tearDown);
-
-          final result = await freshHarness.mcpServerConnection.callTool(
-            CallToolRequest(
-              name: ToolNames.getNetworkLogs.name,
-              arguments: {},
-            ),
-          );
-
-          expect(result.isError, true);
         });
       });
     });


pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -1229,6 +1394,80 @@ base mixin DartToolingDaemonSupport
     ),
   )..categories = [FeatureCategory.flutterDriver];
 
+  @visibleForTesting
+  static final getNetworkLogsTool = Tool(
+    name: ToolNames.getNetworkLogs.name,
+    description:
+        'Fetches recorded HTTP requests from the running app. '
+        'HTTP profiling is enabled automatically when the app connects, so '
+        'requests are captured from app start. Use updatedSince to poll only '
```
jakemac53, line 1403: nit: logs are not actually captured from app start, they are only captured once we connect.

I think we actually should have an explicit tool call though to start capturing logs, and also to stop capturing them. Otherwise I think we will implicitly have a bit of a memory leak here?

pkgs/dart_mcp_server/test/tools/network_inspector_test.dart
```diff
@@ -0,0 +1,114 @@
...
+import 'package:process/process.dart';
+import 'package:test/test.dart';
+
+import '../test_harness.dart';
+
+void main() {
+  late TestHarness testHarness;
+
+  group('network inspector tools', () {
+    group('[in-process]', () {
```
jakemac53, line 20: nit: I would just make all these tests uses an in process test harness - we really only need a few total tests that use an out of process MCP server, and its much easier to debug when its in process

docs/superpowers/plans/2026-04-02-network-inspector.md
```diff
@@ -0,0 +1,698 @@
+# Network Inspector Implementation Plan
```
jakemac53, line 1: nit: Please remove these docs from the PR :)

pkgs/dart_mcp_server/dart_test.yaml
```diff
@@ -0,0 +1,2 @@
+enable_experiment:
```
jakemac53, line 1: nit: we should not be enabling experiments in any code we land here, even for tests

pkgs/dart_mcp_server/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ description: >-
   models.
 publish_to: none
 environment:
-  sdk: ^3.12.0-257.0.dev
```
jakemac53, line 7: This needs to remain at the old constraint - some of our tools do rely on stuff that only exists after this dev release (listing active DTD URIs)

pkgs/dart_mcp_server/test/tools/network_inspector_test.dart
```diff
@@ -0,0 +1,114 @@
...
+          final result = await testHarness.callToolWithRetry(
+            CallToolRequest(
+              name: ToolNames.clearNetworkLogs.name,
+              arguments: {},
+            ),
+          );
+
+          expect(result.isError, isNot(true));
+        });
+      });
```
jakemac53, line 111: We should add another test fixture that does make network requests, so that we can test that actual network logs can be captured, cleared, etc.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -284,6 +319,136 @@ base mixin DartToolingDaemonSupport
...
+                  'response_size': r.response?.contentLength,
+                  'error': r.response == null && r.endTime != null
+                      ? 'request failed'
+                      : null,
+                },
+              )
+              .toList();
+          return CallToolResult(
+            content: [TextContent(text: jsonEncode(requests))],
+            structuredContent: {'requests': requests},
```
jakemac53, line 359: nit: I would remove the structured content here unless we need it for something. The text content can already easily just be decoded as JSON if some tooling wants to explicitly do that.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -284,6 +319,136 @@ base mixin DartToolingDaemonSupport
...
+        request.arguments?[ParameterNames.updatedSince] as String?;
+    final updatedSince =
+        updatedSinceStr != null ? DateTime.tryParse(updatedSinceStr) : null;
+
+    return _callOnVmService(
+      appUri: appUri,
+      callback: (vmService) async {
+        try {
+          final vm = await vmService.getVM();
+          final isolateId = vm.isolates!.first.id!;
```
jakemac53, line 335: Above we enable network logging for _all_ isolates, but here we only will return logs for the first one.

Supporting only the root isolate for now is probably fine, but we should also only enable capturing logs for the first isolate then.

We should also file a bug and add a TODO for supporting additional isolates.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -284,6 +319,136 @@ base mixin DartToolingDaemonSupport
...
+
+  /// Clears the HTTP profile buffer for the running app.
+  Future<CallToolResult> _clearNetworkLogs(CallToolRequest request) async {
+    final appUri = request.arguments?[ParameterNames.appUri] as String?;
+    return _callOnVmService(
+      appUri: appUri,
+      callback: (vmService) async {
+        try {
+          final vm = await vmService.getVM();
+          final isolateId = vm.isolates!.first.id!;
```
jakemac53, line 379: Same here, we should either consistently support all the isolates or only the first one.

jakemac53 (changes requested):
Awesome, this is a great start! 🚀 🥳 

One higher level comment is that we are moving to a pattern of "meta tools" generally in order to save on tool counts (and tokens). To be consistent with that please migrate these all to a single "network" tool, with a "command" argument. See the DTD tool as an example.

PR #438, opened 2026-04-12
fix: graceful shutdown of orphaned test apps on windows

### LINKED ISSUE #407: Fix multi app DTD test on windows
See https://github.com/dart-lang/ai/pull/406

The test works fine, but is failing to clean up for some reason. It gets an error when trying to delete the temp directories, due to some process still holding on there.

Sourav-Sonkar:
Fixes #407

This PR resolves the multi-app DTD test teardown failure on Windows.

**What this PR changes:**
- Modified the test dummy applications in `dtd_multiple_connection_test.dart` and `infinite_wait.dart` to gracefully shut down by listening to `stdin` and calling `exit(0)` when the pipe is closed.

**Why:**
On Windows, when `AppDebugSession` terminates the parent `dart run` test process via `process.kill()`, the child `dart` process running the infinite `while(true)` loop tends to be left orphaned. This orphaned process maintained a lock on the `d.sandbox` directory, preventing the test framework from deleting the temp directory during teardown and throwing a "file in use" or "access denied" error. By having the test fixture scripts monitor `stdin` and exit on its closure, the child process dies gracefully as soon as the test stops the debug session or the test forcefully crashes.

---

<details>

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>

### DIFF
--- a/pkgs/dart_mcp_server/test/test_harness.dart
+++ b/pkgs/dart_mcp_server/test/test_harness.dart
@@ -277,7 +277,7 @@ final class AppDebugSession {
       [
         'run',
         '--no${isFlutter ? '' : '-serve'}-devtools',
-        '--enable-vm-service=0',
+        if (!isFlutter) '--enable-vm-service=0',
         if (isFlutter) ...['-d', 'flutter-tester'],
         appPath,
         ...args,
--- a/pkgs/dart_mcp_server/test/test_harness.dart
+++ b/pkgs/dart_mcp_server/test/test_harness.dart
@@ -28,7 +28,7 @@ import 'package:unified_analytics/unified_analytics.dart';
 
 Future<T> callWithRetry<T>(
   FutureOr<T> Function() fn, {
-  int maxTries = 10,
+  int maxTries = 5,
   bool Function(T)? retryUntil,
 }) async {
   var tryCount = 0;
@@ -231,7 +231,7 @@ class TestHarness {
   /// Some methods will fail if the DTD connection is not yet ready.
   Future<CallToolResult> callToolWithRetry(
     CallToolRequest request, {
-    int maxTries = 10,
+    int maxTries = 5,
     bool expectError = false,
     bool Function(CallToolResult)? retryUntil,
   }) => callWithRetry(
@@ -275,9 +275,9 @@ final class AppDebugSession {
     final process = await TestProcess.start(
       isFlutter ? sdk.flutterExecutablePath : sdk.dartExecutablePath,
       [
-        if (isFlutter) 'run',
-        if (isFlutter) '--no-devtools',
-        if (!isFlutter) '--enable-vm-service=0',
+        'run',
+        '--no${isFlutter ? '' : '-serve'}-devtools',
+        '--enable-vm-service=0',
         if (isFlutter) ...['-d', 'flutter-tester'],
         appPath,
         ...args,
@@ -336,12 +336,9 @@ final class AppDebugSession {
       }
       try {
         await process.shouldExit(0).timeout(const Duration(seconds: 10));
-      } catch (_) {
+      } on TimeoutException catch (_) {
         await process.kill();
-        // Ignore exit code since we had to forcefully kill
-        try {
-          await process.shouldExit().timeout(const Duration(seconds: 5));
-        } catch (_) {}
+        await process.shouldExit();
       }
     } else {
       // Send q which triggers graceful shutdown in our test scripts
@@ -355,11 +352,9 @@ final class AppDebugSession {
       }
       try {
         await process.shouldExit().timeout(const Duration(seconds: 10));
-      } catch (_) {
+      } on TimeoutException catch (_) {
         await process.kill();
-        try {
-          await process.shouldExit().timeout(const Duration(seconds: 5));
-        } catch (_) {}
+        await process.shouldExit();
       }
     }
   }
@@ -485,12 +480,14 @@ class FakeEditorExtension {
     await dtd.close();
     try {
       await dtdProcess.stdin.close();
-      await dtdProcess.shouldExit().timeout(const Duration(seconds: 5));
     } catch (_) {
+      // stdin already closed
+    }
+    try {
+      await dtdProcess.shouldExit().timeout(const Duration(seconds: 5));
+    } on TimeoutException catch (_) {
       await dtdProcess.kill();
-      try {
-        await dtdProcess.shouldExit().timeout(const Duration(seconds: 5));
-      } catch (_) {}
+      await dtdProcess.shouldExit();
     }
   }
 }
@@ -579,6 +576,8 @@ Future<ServerConnectionPair> _initializeMCPServer(
     connection = client.connectServer(clientChannel);
   } else {
     final process = await Process.start(sdk.dartExecutablePath, [
+      'pub', // Using `pub` gives us incremental compilation
+      'run',
       'bin/main.dart',
       ...cliArgs,
       for (var enabled in featuresConfig.enabledNames) '--enable=$enabled',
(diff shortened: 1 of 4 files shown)


pkgs/dart_mcp_server/test/tools/dtd_multiple_connection_test.dart
jakemac53: This is the line that needs to be removed

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,29 @@
+name: Test on Windows
```
jakemac53: Instead of adding a new file here, just remove the skip below (I left a comment on it). This job wouldn't actually run the test (it would be skipped), and the existing job already does run for windows.

Sourav-Sonkar: Removed

pkgs/dart_mcp_server/test/tools/dtd_multiple_connection_test.dart
Sourav-Sonkar: Removed

pkgs/dart_mcp_server/test/tools/dtd_multiple_connection_test.dart
```diff
@@ -275,9 +277,7 @@ void main() async {
           contains('error from app 2'),
         );
       },
-      skip: Platform.isWindows
-          ? 'https://github.com/dart-lang/ai/issues/407'
-          : false,
+      skip: false,
```
jakemac53:
```suggestion
```

Sourav-Sonkar:
@jakemac53  i will check the issue and fix them

Sourav-Sonkar:
@jakemac53 can you please run the workflows

pkgs/dart_mcp_server/test/test_harness.dart
```diff
@@ -322,10 +322,29 @@ final class AppDebugSession {
   static Future<void> kill(TestProcess process, bool isFlutter) async {
     if (isFlutter) {
       process.stdin.writeln('q');
-      await process.shouldExit(0);
```
jakemac53, line 325: Are these try/catch still necessary? It seems like these will just paper over real issues

Sourav-Sonkar: Changed to on StateError catch (_) with a lint ignore. The catch is legit — kill() gets called twice during teardown (test + harness), so stdin can already be closed on the second pass. Added a comment explaining this. Both Ubuntu and Windows CI passing

Sourav-Sonkar: @jakemac53  can please check

pkgs/dart_mcp_server/test/test_harness.dart
```diff
@@ -28,7 +28,7 @@ import 'package:unified_analytics/unified_analytics.dart';
 
 Future<T> callWithRetry<T>(
   FutureOr<T> Function() fn, {
-  int maxTries = 5,
+  int maxTries = 10,
```
jakemac53: Do we need to bump this to 10 or is this just left over from debugging?

pkgs/dart_mcp_server/test/test_harness.dart
```diff
@@ -275,8 +275,8 @@ final class AppDebugSession {
     final process = await TestProcess.start(
       isFlutter ? sdk.flutterExecutablePath : sdk.dartExecutablePath,
       [
-        'run',
-        '--no${isFlutter ? '' : '-serve'}-devtools',
+        if (isFlutter) 'run',
+        if (isFlutter) '--no-devtools',
```
jakemac53: I think these changes should be reverted

pkgs/dart_mcp_server/test/test_harness.dart
```diff
@@ -319,13 +319,48 @@ final class AppDebugSession {
...
-      process.stdin.writeln('q');
-      await process.shouldExit(0);
+      try {
+        process.stdin.writeln('q');
+        // ignore: avoid_catching_errors
+      } on StateError catch (_) {
+        // stdin already closed from a prior kill() call.
+      }
+      try {
+        await process.shouldExit(0).timeout(const Duration(seconds: 10));
```
jakemac53, line 338: These try/catch blocks here will no longer assert a non-zero exit code like they previously did (since we are just catching the test failures)

pkgs/dart_mcp_server/test/test_harness.dart
```diff
@@ -447,9 +482,16 @@ class FakeEditorExtension {
 
   Future<void> shutdown() async {
     await _debugSessions.toList().map(removeDebugSession).wait;
-    await dtdProcess.kill();
-    await dtdProcess.shouldExit();
     await dtd.close();
+    try {
```
jakemac53, line 481: same for these ones, we will no longer be checking the exit code or that it exited at all.

pkgs/dart_mcp_server/test/test_harness.dart
```diff
@@ -537,16 +579,27 @@ Future<ServerConnectionPair> _initializeMCPServer(
     connection = client.connectServer(clientChannel);
   } else {
     final process = await Process.start(sdk.dartExecutablePath, [
-      'pub', // Using `pub` gives us incremental compilation
```
jakemac53, line 540: Do we need to remove this? The incremental compilation speeds up tests a lot (it is shared across test suites).

pkgs/dart_mcp_server/test/test_harness.dart
```diff
@@ -537,16 +579,27 @@ Future<ServerConnectionPair> _initializeMCPServer(
...
-      'run',
       'bin/main.dart',
       ...cliArgs,
       for (var enabled in featuresConfig.enabledNames) '--enable=$enabled',
       for (var disabled in featuresConfig.disabledNames) '--disable=$disabled',
     ]);
     addTearDown(() async {
-      process.kill();
-      await process.exitCode;
+      try {
```
jakemac53, line 587: A same here, we won't actually wait for exit codes any more.

If `stdin.close()` can fail lets only try/catch that, but not the waiting for the exit code.

pkgs/dart_mcp_server/test/test_harness.dart
Sourav-Sonkar: This was just from debugging slow CI runs. Reverted back to 5.

pkgs/dart_mcp_server/test/test_harness.dart
Sourav-Sonkar: Done. Restored dart run and the original flag structure. Also fixed it so --enable-vm-service=0 only goes to Dart apps since Flutter doesn't recognize that flag.

pkgs/dart_mcp_server/test/test_harness.dart
Sourav-Sonkar: You're absolutely right. Changed these to only catch TimeoutException so we still assert the exit codes properly. The graceful shutdown flow is now: try to exit cleanly with timeout, if timeout then kill and wait for actual exit.

pkgs/dart_mcp_server/test/test_harness.dart
Sourav-Sonkar: No we don't need to remove it - you're right about the performance benefit. Restored pub run for the MCP server.

pkgs/dart_mcp_server/test/test_harness.dart
Sourav-Sonkar: Good point. Split the exception handling so stdin.close() failures are caught separately, but we still wait for and assert the exit code.

pkgs/dart_mcp_server/test/test_harness.dart
Sourav-Sonkar: Fixed. Now we only catch exceptions from stdin.close() itself, and the exit code wait is separate so it still asserts proper shutdown.

jakemac53 (approved):

PR #473, opened 2026-05-14
add a tool to make arbitrary vm service method calls

### LINKED ISSUE #472: Exposing the drift dev_tool extension via the MCP
Drift is a flutter package that provides a dart API to connect to a SQLite database.
It has a dev_tool extension: https://drift.simonbinder.eu/tools/devtools/

What would it take to make it accessible via the MCP server so that I could ask copilot in vscode questions about the database running in my simulator/emulator and it would be able to run some sql queries on it?

jakemac53:
Adds a new tool `call_vm_service_method`, which takes an app uri, method name, isolate ID, and arguments, and returns the raw vm service response.

While low level, you can easily write various skills on top of this API to do things such as calling custom VM service methods (specific to a package or app), setting breakpoints, evaluating expressions, etc.

Ultimately it is a very powerful building block for arbitrary agent interactions, and anybody can write skills to guide agents through complex use cases.

Closes https://github.com/dart-lang/ai/issues/472

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -204,6 +204,7 @@ base mixin DartToolingDaemonSupport
     registerTool(hotReloadTool, hotReload);
     registerTool(widgetInspectorTool, _widgetInspector);
     registerTool(flutterDriverTool, _callFlutterDriver);
+    registerTool(callVmServiceMethodTool, _callVmServiceMethod);
 
     return super.initialize(request);
   }
@@ -217,6 +218,7 @@ base mixin DartToolingDaemonSupport
     hotReloadTool,
     widgetInspectorTool,
     flutterDriverTool,
+    callVmServiceMethodTool,
   ];
 
   @override
@@ -284,7 +286,46 @@ base mixin DartToolingDaemonSupport
     );
   }
 
+  Future<CallToolResult> _callVmServiceMethod(CallToolRequest request) async {
+    // These are already validated to match the schema definition.
+    final method = request.arguments?[ParameterNames.method] as String;
+    var isolateId = request.arguments?[ParameterNames.isolateId] as String?;
+    final args =
+        request.arguments?[ParameterNames.arguments] as Map<String, Object?>?;
+    final appUri = request.arguments?[ParameterNames.appUri] as String?;
+
+    return _callOnVmService(
+      appUri: appUri,
+      callback: (vmService) async {
+        isolateId ??= (await vmService.getVM()).isolates!.first.id;
+        try {
+          final result = await vmService.callMethod(
+            method,
+            isolateId: isolateId,
+            args: args,
+          );
+          return CallToolResult(
+            content: [TextContent(text: jsonEncode(result.json))],
+          );
+        } on RPCError catch (e) {
+          return CallToolResult(
+            isError: true,
+            content: [
+              TextContent(text: 'RPCError when invoking method $method: $e'),
+            ],
+          )..failureReason = CallToolFailureReason.rpcError;
+        } catch (e) {
+          return CallToolResult(
+            isError: true,
+            content: [TextContent(text: 'Error invoking method $method: $e')],
+          )..failureReason = CallToolFailureReason.unhandledError;
+        }
+      },
+    );
+  }
+
   /// Connects to the Dart Tooling Daemon.
+  ///
   /// Connects to a single DTD at [uri].
   Future<CallToolResult> _connectToDtdSingle(Uri uri) async {
     if (_dtds.any((dtd) => dtd.uri == uri)) {
@@ -1364,6 +1405,38 @@ base mixin DartToolingDaemonSupport
         )
         ..categories = [FeatureCategory.dartToolingDaemon]
         ..enabledByDefault = false;
+
+  @visibleForTesting
+  static final callVmServiceMethodTool = Tool(
+    name: ToolNames.callVmServiceMethod.name,
+    description:
+        'Invoke VM service methods on a connected app. See the Public RPCs '
+        'section of '
+        'https://raw.githubusercontent.com/dart-lang/sdk/refs/heads/main/runtime/vm/service/service.md',
+    annotations: ToolAnnotations(title: 'Invoke VM Service Method'),
+    inputSchema: Schema.object(
+      properties: {
+        ParameterNames.method: Schema.string(
+          description: 'The name of the method to invoke.',
+        ),
+        ParameterNames.isolateId: Schema.string(
+          description:
+              'The isolate ID to target (optional), defaults to the main '
+              'isolate.',
+        ),
+        ParameterNames.arguments: Schema.object(
+          description: 'Arguments for the method (optional).',
+          additionalProperties: true,
+        ),
+        ParameterNames.appUri: Schema.string(
+          description:
+              'The app URI to target. Required if multiple apps are connected.',
+        ),
+      },
+      required: [ParameterNames.method],
+      additionalProperties: false,
+    ),
+  )..categories = [FeatureCategory.dartToolingDaemon];
 
   static final _connectedAppsNotSupported = CallToolResult(
     isError: true,
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -1421,7 +1421,8 @@ base mixin DartToolingDaemonSupport
         ParameterNames.isolateId: Schema.string(
           description:
               'The isolate ID to target, you can find isolate IDs using the '
-              '`getVM` method.',
+              '`getVM` method. Only methods that require an isolate ID should '
+              'be given one.',
         ),
         ParameterNames.arguments: Schema.object(
           description: 'Arguments for the method (optional).',
--- a/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
@@ -1411,6 +1411,8 @@ base mixin DartToolingDaemonSupport
     description:
         'Invoke VM service methods on a connected app. See the Public RPCs '
         'section of '
+        // TODO: Make vm service self describing
+        // https://github.com/dart-lang/sdk/issues/63415
         'https://raw.githubusercontent.com/dart-lang/sdk/refs/heads/main/runtime/vm/service/service.md',
     annotations: ToolAnnotations(title: 'Invoke VM Service Method'),
     inputSchema: Schema.object(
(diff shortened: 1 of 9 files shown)


jakemac53:
Just as an anecdote, Gemini CLI was able to successfully set breakpoints using just this tool and a simple prompt.

jakemac53:
cc @simolus3

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -1365,6 +1406,38 @@ base mixin DartToolingDaemonSupport
         ..categories = [FeatureCategory.dartToolingDaemon]
         ..enabledByDefault = false;
 
+  @visibleForTesting
+  static final callVmServiceMethodTool = Tool(
+    name: ToolNames.callVmServiceMethod.name,
+    description:
+        'Invoke VM service methods on a connected app. See the Public RPCs '
+        'section of '
+        'https://raw.githubusercontent.com/dart-lang/sdk/refs/heads/main/runtime/vm/service/service.md',
```
kenzieschmoll, line 1416: @bkonyi is this spec guaranteed to be up to date?

@jakemac53 one thing to consider is that this may not be compatible with the version of Dart a user has. It should mostly work because the VM Service API is relatively stable, but I could see breaking changes there leading to tool call failures.

jakemac53: That is fair... I don't know if we maintain a changelog or something, hopefully Ben can enlighten us.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -284,7 +286,46 @@ base mixin DartToolingDaemonSupport
     );
   }
 
+  Future<CallToolResult> _callVmServiceMethod(CallToolRequest request) async {
+    // These are already validated to match the schema definition.
+    final method = request.arguments?[ParameterNames.method] as String;
+    var isolateId = request.arguments?[ParameterNames.isolateId] as String?;
```
kenzieschmoll: does the agent have to provide this? if so, how does it get access to it?

jakemac53: This is optional but the agents can provide it.

In the example I walked through though, it actually ended up calling `getVM` and actually listing out the isolates and then passing one in. Essentially just exactly what we would do in code.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -284,7 +286,46 @@ base mixin DartToolingDaemonSupport
...
+    final args =
+        request.arguments?[ParameterNames.arguments] as Map<String, Object?>?;
+    final appUri = request.arguments?[ParameterNames.appUri] as String?;
+
+    return _callOnVmService(
+      appUri: appUri,
+      callback: (vmService) async {
+        isolateId ??= (await vmService.getVM()).isolates?.firstOrNull?.id;
```
kenzieschmoll: Note: we are assuming the first isolate is the main isolate here. Are we okay with that or should we add some heuristics around the isolate name?

jakemac53: Maybe @bkonyi should weigh in on this as well. This is what we do throughout this file so we should update everything if we want to change it.

pkgs/dart_mcp_server/test_fixtures/dart_cli_app/bin/infinite_wait.dart
```diff
@@ -2,9 +2,25 @@
...
+    );
+  });
+
+  registerExtension('ext.test.failure', (method, parameters) async {
+    return ServiceExtensionResponse.error(
+      ServiceExtensionResponse.extensionError,
+      jsonEncode({'message': 'Something went wrong'}),
+    );
+  });
+  print('ready');
```
kenzieschmoll, line 22: should this print be removed?

jakemac53: This helps the tests to know when the extension methods are registered

jakemac53:
I noticed that the defaulting of the `isolateId` here actually makes some RPC calls fail (the ones like `getVm` which don't take an Isolate ID) - so I think I will remove that, at least for now.

We could evaluate replacing it with a special `main` placeholder which will do the defaulting.

Surprisingly, agents are quite good at calling `getVm` to get the list of isolate IDs themselves.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: I removed this entirely, during evals I found agents are able to call `getVM` themselves, and providing this for some methods causes invalid params errors.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
bkonyi: Yes, the spec is updated whenever we make changes to the public API surface. We don't document implementation changes or modifications to private APIs.

We do our best to keep the change log at the end of the spec document updated. For breaking changes, we won't break the API surface without a major revision bump, but deprecated RPCs and fields may return valid empty results.

bkonyi: Do we always want to be checking the latest spec from main? If we start making breaking changes with a major revision bump, users on the latest stable using this tool could be broken.

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
```diff
@@ -1365,6 +1405,38 @@ base mixin DartToolingDaemonSupport
...
+        'https://raw.githubusercontent.com/dart-lang/sdk/refs/heads/main/runtime/vm/service/service.md',
+    annotations: ToolAnnotations(title: 'Invoke VM Service Method'),
+    inputSchema: Schema.object(
+      properties: {
+        ParameterNames.method: Schema.string(
+          description: 'The name of the method to invoke.',
+        ),
+        ParameterNames.isolateId: Schema.string(
+          description:
+              'The isolate ID to target, you can find isolate IDs using the '
```
bkonyi, line 1425: Is it worth mentioning that not all RPCs that do not require an isolate ID should not be provided one?

pkgs/dart_mcp_server/test/tools/dtd_test.dart
```diff
@@ -1277,6 +1277,74 @@ void main() {
...
+            name: ToolNames.callVmServiceMethod.name,
+            arguments: {
+              ParameterNames.method: 'ext.test.echo',
+              ParameterNames.arguments: {'message': 'hello'},
+              ParameterNames.isolateId: isolateId,
+            },
+          ),
+        );
+
+        expect(result.isError, isNot(true));
```
bkonyi: `isFalse` or `false`?

pkgs/dart_mcp_server/test/tools/dtd_test.dart
```diff
@@ -1277,6 +1277,74 @@ void main() {
...
+          CallToolRequest(
+            name: ToolNames.callVmServiceMethod.name,
+            arguments: {
+              ParameterNames.method: 'ext.test.failure',
+              ParameterNames.arguments: {'message': 'hello'},
+              ParameterNames.isolateId: isolateId,
+            },
+          ),
+          expectError: true,
+          retryUntil: (r) =>
```
bkonyi: Nit: consider using a `switch` here with object destructuring:

```dart
(r) => switch (r) {
      CallToolResult(isError: true, content: [TextContent(text: var msg)]) =>
        msg.contains('Something went wrong'),
      _ => false,
    };
```

pkgs/dart_mcp_server/test/tools/dtd_test.dart
```diff
@@ -1418,3 +1486,18 @@ Future<void> _deleteWithRetry(Directory dir) async {
...
 }
+
+Future<String> _getIsolateId(TestHarness testHarness) async {
+  final getVmResponse = await testHarness.callTool(
+    CallToolRequest(
+      name: ToolNames.callVmServiceMethod.name,
+      arguments: {ParameterNames.method: 'getVM'},
+    ),
+  );
+  final vmObject =
```
bkonyi: Nit: also consider using object destructuring here:

```dart
if (getVmResponse case CallToolResult(content: [TextContent(text: var text)])) {
  if (jsonDecode(text) case {'isolates': [{'id': String id}, ...]}) {
    return id;
  }
}
throw StateError('Failed to extract isolate ID from VM response');
```

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: sure, done

pkgs/dart_mcp_server/test/tools/dtd_test.dart
jakemac53: this isn't even necessary actually its validated by the function call above

pkgs/dart_mcp_server/test/tools/dtd_test.dart
jakemac53: Sure, done

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: I don't know what else we can do - do we have the current stable version posted anywhere? I don't know of a way to point at it.

We could also potentially ship the spec with the SDK, possibly in a different format (a json schema?). Or even make the VM service be self-describing and allow you to directly fetch a JSON schema from it. This would probably be the ideal approach.

If we did that, I think you would want progressive disclosure as well, only show the method names and their descriptions initially, and then allow you to separately query for the JSON schemas of the methods.

jakemac53: This would ideally also allow publishing a JSON schema for registered VM service methods, either from other clients or the app itself.

jakemac53: @kenzieschmoll given that I don't think there is really a great solution  here are you fine with just landing this as is? The VM service doesn't really tend to change often (especially not in breaking ways) so I doubt it will really cause an issue.

kenzieschmoll: I like the idea of making the VM service self describing and allowing you to fetch a JSON schema from it. Can you file an issue so we don't lose track of this? That would make this MCP tool more stable and guaranteed to be compatible with the user's SDK version.

Fine to land as is though, but maybe link an issue in a TODO to polish.

kenzieschmoll (approved):

pkgs/dart_mcp_server/lib/src/mixins/dtd.dart
jakemac53: https://github.com/dart-lang/sdk/issues/63415

PR #488, opened 2026-05-26
add back executable section, necessary for dart run

jakemac53:
Apparently this is required by `dart run <package>@`.

cc @dcharkes is this intentional? It isn't required for `dart run <package>`. It just assumes a binary with the package name afaik.


dcharkes:
> cc @dcharkes is this intentional? It isn't required for `dart run <package>`. It just assumes a binary with the package name afaik.

Yes. From the `dart install` design doc:

> Making a package installable has more requirements on a package then making it runnable. The executables section communicates from the package author that the files in the bin/ directory satisfy these requirements.

(The requirements are for example not having access to your own source files. Being AOT compiled and not having access to a Dart SDK necessarily. Or the Dart SDK that is on your system being a different version than the AOT compiled + aotruntime installed version.)

And `dart run ...@` is built on `dart install` (for now).

bwilkerson (approved):

PR #495, opened 2026-06-03
migrate dart_mcp to package:checks

jakemac53:
Migrates dart_mcp to `package:checks` as a test case.

This is a pure AI migration in the initial commit.

## Notable changes

- Exposed a new `api.dart` public library which contains no server or client specific APIs.
- Made all extension types implement `Object`, this helps with writing extensions on these extension types, otherwise they all appear nullable and checks extensions don't really work.

### DIFF
--- a/pkgs/dart_mcp/test/api/initialization_test.dart
+++ b/pkgs/dart_mcp/test/api/initialization_test.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:checks/checks.dart';
 import 'package:dart_mcp/server.dart';
 import 'package:test/test.dart';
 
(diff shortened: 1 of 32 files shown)


pkgs/dart_mcp/test/api/initialization_test.dart
```diff
@@ -26,6 +27,6 @@ void main() {
     );
 
     final map = result as Map<String, Object?>;
-    expect(map['instructions'], equals('foo'));
+    check(map)['instructions'].equals('foo');
```
jakemac53, line 30: Should `containsKey` return a `Subject<V>`? It feels like `check(map).containsKey('instructions').equals('foo')` might be better here?

cc @natebosch

jakemac53: Or I guess the `[]` operator actually does exactly this, so nvm

PR #516, opened 2026-07-08
(Draft) Cursor plugin

johnpryan:


keertip:
Moved to flutter/agent-plugins.

PR #521, opened 2026-07-13
suggest installing dart/flutter skill repos

### LINKED ISSUE #504: Suggest installing flutter/dart skills in new repos
When we don't have any saved configuration, we should present a dialog with "suggested skill sources" on `skills get`, and include both global and local options. This should list the dart/flutter repos, and it should do the equivalent of a `skills add <repo>` for any selected repos, possibly with an additional dialog asking whether you want to install them globally or locally. The list should also only show the flutter skills repo if your project depends on flutter.

We should also save to the manifest whether or not you have been shown this dialog, (and for which repos), so we don't show it again for that package.

Additionally we may want a "never ask again" or "remember my choice for future packages" option, which would save to the global config your selections and apply that to future packages without a dialog.

Blocked on https://github.com/dart-lang/ai/issues/470

jakemac53:
Closes https://github.com/dart-lang/ai/issues/504

- Once per project/machine (as recorded in the skill manifest), suggest installing dart/flutter skill repos.
- Allows a global opt out for this behavior.
- Can install either globally or locally (if global, this dialog will be skipped for future projects).


pkgs/skills/lib/src/models/skill_manifest.dart
```diff
@@ -114,7 +119,17 @@ class SkillManifest {
       return MapEntry(ideKey, pkgs);
     });
 
-    return SkillManifest(version: version, installations: installations);
+    final suggestedRepos =
+        (json['suggestedRepos'] as List<dynamic>?)
+            ?.map((e) => e as String)
+            .toSet() ??
+        const {};
```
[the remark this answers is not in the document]
jakemac53: I don't like just skipping non-string entries, but I did change this to use the more explicit `.cast<String>()`.

pkgs/skills/lib/src/commands/get_skills.dart
```diff
@@ -235,6 +259,101 @@ Future<bool> getSkills({
...
+  // Only prompt when we can have dialog support, and the user didn't already
+  // specify specific skills or sources.
+  if (globalConfig.neverPromptForSuggestedSkills ||
+      dialogSupport == null ||
+      sourceUris.isNotEmpty ||
+      skillNames.isNotEmpty) {
+    return (globalConfig: globalConfig, manifest: manifest);
+  }
+
+  const flutterSkillsRepo = 'https://github.com/flutter/skills.git';
```
kenzieschmoll: this now needs to be flutter/agent-plugins

pkgs/skills/lib/src/commands/get_skills.dart
```diff
@@ -235,6 +259,101 @@ Future<bool> getSkills({
...
+  final hasFlutter = resolvedPackages.any((p) => p.name == 'flutter');
+  if (hasFlutter && shouldSuggest(flutterSkillsRepo)) {
+    suggestedRepos.add(flutterSkillsRepo);
+  }
+
+  if (suggestedRepos.isNotEmpty) {
+    final options = [...suggestedRepos, 'Never ask again on this machine'];
+    final selectedIndices = await dialogSupport.showMultiSelectDialog(
+      options,
+      title: 'Would you like to install the official Dart or Flutter skills?',
```
kenzieschmoll: The dialog title always says "Would you like to install the official Dart or Flutter skills?" even if only Dart skills are being suggested (for example, in a non-Flutter project). It could dynamically adjust to just say "Dart skills" if flutterSkillsRepo isn't in the suggestedRepos list.

pkgs/skills/test/fake_dialog_support.dart
```diff
@@ -52,6 +57,12 @@ class FakeDialogSupport implements DialogSupport {
     String? title,
     Set<int> initialSelected = const {},
   }) async {
+    if (skipSuggestedRepos &&
+        title ==
+            'Would you like to install the official Dart or Flutter skills?') {
```
kenzieschmoll: nit: if you apply the comment above to change the string this will be fragile. Let's put this string in a const that can be accessed from this fake dialog support file instead of doing an equality check on the raw string

pkgs/skills/CHANGELOG.md
```diff
@@ -30,6 +30,8 @@
   `npx skills` for `dart run skills@`.
   - The `get` command now supports `--git` arguments for git repos to update
     skills from.
+- feat: Oncer per machine/project, suggest installing the dart and flutter
```
kenzieschmoll: typo: Once

kenzieschmoll (approved):

Issue #162, opened 2025-06-06
Add support for Streamable HTTP to dart_mcp client


ghchinoy:
Provide support for the latest transport of the MCP project, [Streamable HTTP](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http) so that dart/flutter clients can utilize HTTP-based MCP Servers and tools.

michael-mcroskey:
+1 from [FlutterFlow](https://github.com/FlutterFlow) — we're interested in Streamable HTTP support for dart_mcp as we're soon rolling out MCP support

jakemac53:
Hi @ksokolovskyi thanks for offering to help here!

My original thought was that we don't need a `Transport` API at all - I don't know what it provides over just a raw StreamChannel. Instead, I was planning to just provide helper functions for each transport (see `lib/stdio.dart` for an example), which just give you the stream channel.

But, I am definitely open to any arguments for adding an explicit `Transport` class if you think we need one.

jakemac53:
@ksokolovskyi checking if you have any thoughts or have you started looking into this at all?

nikeokoronkwo:
Would love to see this feature.
Would authentication be supported as well for the client and server?

ksokolovskyi:
> I am a bit confused, it seems to me that your example demonstrates it would be enough? Ultimately, you were able to create a StreamChannel and give that to the server. So, basically you would just be extracting out the logic to create that stream given an /mcp post request into a helper, something like this: 

Unfortunately, my example works only with batched requests. Below, I share the non-batched requests and the responses that will be received

<details closed>
<summary>Non-batched requests</summary>

```sh
curl -X POST http://localhost:3333/mcp \
  -H "Content-Type: application/json" \
  -H "mcp-session-id: test-session-123" \
  -d '{
      "jsonrpc": "2.0",
      "id": 1,
      "method": "initialize",
      "params": {
        "protocolVersion": "2025-03-26",
        "capabilities": {
          "roots": {
            "listChanged": true
          },
          "sampling": {}
        },
        "clientInfo": {
          "name": "ExampleClient",
          "version": "1.0.0"
        }
      }
    }' && 
curl -X POST http://localhost:3333/mcp \
  -H "Content-Type: application/json" \
  -H "mcp-session-id: test-session-123" \
  -d '{
      "jsonrpc": "2.0",
      "id": 2,
      "method": "notifications/initialized"
    }' && 
[...]

PR #524, opened 2026-07-14
Separate server feature registration from legacy initialization

Yusufihsangorgel:
Separates feature registration from the legacy MCP initialize handshake.

- Add `MCPServer.initialize(MCPServerInitialization)` as the reusable feature and capability lifecycle hook. The context carries the real protocol version, client information, and client capabilities for either lifecycle.
- Move protocol negotiation and the legacy wire response to `MCPServer.initializeLegacy(InitializeRequest)`.
- Move the built-in `dart_mcp` support mixins and examples to the reusable hook, including roots and elicitation setup shared by both lifecycle modes.
- Keep `dart_mcp_server` on the published 0.5 line by removing its local `dart_mcp` override; its 0.6 migration remains a separate concern.
- Document the lifecycle boundary and start the breaking `dart_mcp` 0.6.0 development version.

The legacy channel path still performs the same negotiation and calls the reusable hook with the real request context. A future request-scoped dispatcher can invoke feature setup without fabricating a legacy initialize request, then complete `initialized` after the full registration chain and transport setup have finished.

Tests:

- `dart_mcp`: analyze; 736/736 Chrome/VM tests across dart2wasm, dart2js, kernel, and exe.
- `dart_mcp_server` against published `dart_mcp 0.5.2`: analyze; 18/18 selected API, configuration, and server tests.
- `mcp_examples`: analyze and format.
- `dart doc --dry-run`: 0 errors; 128 baseline-identical re-export warnings.

This is only the lifecycle seam discussed in #162. It does not add or claim Streamable HTTP support.

### DIFF
--- a/pkgs/dart_mcp/lib/src/server/elicitation_request_support.dart
+++ b/pkgs/dart_mcp/lib/src/server/elicitation_request_support.dart
@@ -5,12 +5,12 @@ part of 'server.dart';
 base mixin ElicitationRequestSupport on LoggingSupport {
   /// Whether or not the connected client supports elicitation.
   ///
-  /// Only safe to call after calling [initialize] on `super` since this is
-  /// based on the client capabilities.
+  /// Only safe to call after calling [initializeLegacy] on `super` since this
+  /// is based on the client capabilities.
   bool get supportsElicitation => clientCapabilities.elicitation != null;
 
   @override
-  FutureOr<InitializeResult> initialize(InitializeRequest request) {
+  FutureOr<InitializeResult> initializeLegacy(InitializeRequest request) {
     initialized.then((_) {
       if (!supportsElicitation) {
         log(
@@ -20,7 +20,7 @@ base mixin ElicitationRequestSupport on LoggingSupport {
         );
       }
     });
-    return super.initialize(request);
+    return super.initializeLegacy(request);
   }
 
   /// Sends an `elicitation/create` request to the client.
--- a/pkgs/dart_mcp/lib/src/server/elicitation_request_support.dart
+++ b/pkgs/dart_mcp/lib/src/server/elicitation_request_support.dart
@@ -5,12 +5,14 @@ part of 'server.dart';
 base mixin ElicitationRequestSupport on LoggingSupport {
   /// Whether or not the connected client supports elicitation.
   ///
-  /// Only safe to call after calling [initializeLegacy] on `super` since this
+  /// Only safe to call after calling [initialize] on `super` since this
   /// is based on the client capabilities.
   bool get supportsElicitation => clientCapabilities.elicitation != null;
 
   @override
-  FutureOr<InitializeResult> initializeLegacy(InitializeRequest request) {
+  FutureOr<ServerCapabilities> initialize(
+    MCPServerInitialization initialization,
+  ) {
     initialized.then((_) {
       if (!supportsElicitation) {
         log(
@@ -20,7 +22,7 @@ base mixin ElicitationRequestSupport on LoggingSupport {
         );
       }
     });
-    return super.initializeLegacy(request);
+    return super.initialize(initialization);
   }
 
   /// Sends an `elicitation/create` request to the client.
--- a/pkgs/dart_mcp/lib/src/server/roots_tracking_support.dart
+++ b/pkgs/dart_mcp/lib/src/server/roots_tracking_support.dart
@@ -31,20 +31,20 @@ base mixin RootsTrackingSupport on LoggingSupport {
 
   /// Whether or not the connected client supports [listRoots].
   ///
-  /// Only safe to call after calling [initialize] on `super` since this is
-  /// based on the client capabilities.
+  /// Only safe to call after calling [initializeLegacy] on `super` since this
+  /// is based on the client capabilities.
   bool get supportsRoots => clientCapabilities.roots != null;
 
   /// Whether or not the connected client supports reporting changes to the
   /// list of roots.
   ///
-  /// Only safe to call after calling [initialize] on `super` since this is
-  /// based on the client capabilities.
+  /// Only safe to call after calling [initializeLegacy] on `super` since this
+  /// is based on the client capabilities.
   bool get supportsRootsChanged =>
       clientCapabilities.roots?.listChanged == true;
 
   @override
-  FutureOr<InitializeResult> initialize(InitializeRequest request) {
+  FutureOr<InitializeResult> initializeLegacy(InitializeRequest request) {
     initialized.then((_) async {
       if (!supportsRoots) {
         log(
@@ -61,7 +61,7 @@ base mixin RootsTrackingSupport on LoggingSupport {
         await updateRoots();
       }
     });
-    return super.initialize(request);
+    return super.initializeLegacy(request);
   }
 
   /// Updates the list of [roots] by calling [listRoots].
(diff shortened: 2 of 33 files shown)


pkgs/dart_mcp/lib/src/server/elicitation_request_support.dart
```diff
@@ -5,12 +5,12 @@ part of 'server.dart';
 base mixin ElicitationRequestSupport on LoggingSupport {
   /// Whether or not the connected client supports elicitation.
   ///
-  /// Only safe to call after calling [initialize] on `super` since this is
-  /// based on the client capabilities.
+  /// Only safe to call after calling [initializeLegacy] on `super` since this
```
jakemac53: I think this should still get set up in `initialize` and not `initializeLegacy`, which should get called once for each request essentially in the new model, or once per connection in the legacy model (I think we will ultimately want an entirely fresh server instance per request when operating in the new model).

pkgs/dart_mcp/lib/src/server/elicitation_request_support.dart
```diff
@@ -5,12 +5,12 @@ part of 'server.dart';
...
+  /// is based on the client capabilities.
   bool get supportsElicitation => clientCapabilities.elicitation != null;
 
   @override
-  FutureOr<InitializeResult> initialize(InitializeRequest request) {
+  FutureOr<InitializeResult> initializeLegacy(InitializeRequest request) {
     initialized.then((_) {
       if (!supportsElicitation) {
```
jakemac53, line 17: I think this should probably move into `initialize` and out of `initializeLegacy`, it will apply to both modes.

pkgs/dart_mcp/lib/src/server/roots_tracking_support.dart
```diff
@@ -31,20 +31,20 @@ base mixin RootsTrackingSupport on LoggingSupport {
 
   /// Whether or not the connected client supports [listRoots].
   ///
-  /// Only safe to call after calling [initialize] on `super` since this is
-  /// based on the client capabilities.
+  /// Only safe to call after calling [initializeLegacy] on `super` since this
```
jakemac53: Same comment, I think this should happen in initialize still

pkgs/dart_mcp/lib/src/server/roots_tracking_support.dart
```diff
@@ -31,20 +31,20 @@ base mixin RootsTrackingSupport on LoggingSupport {
...
+  /// is based on the client capabilities.
   bool get supportsRoots => clientCapabilities.roots != null;
 
   /// Whether or not the connected client supports reporting changes to the
   /// list of roots.
   ///
-  /// Only safe to call after calling [initialize] on `super` since this is
-  /// based on the client capabilities.
+  /// Only safe to call after calling [initializeLegacy] on `super` since this
```
jakemac53: here as well

pkgs/dart_mcp/lib/src/server/roots_tracking_support.dart
```diff
@@ -31,20 +31,20 @@ base mixin RootsTrackingSupport on LoggingSupport {
...
+  /// is based on the client capabilities.
   bool get supportsRootsChanged =>
       clientCapabilities.roots?.listChanged == true;
 
   @override
-  FutureOr<InitializeResult> initialize(InitializeRequest request) {
+  FutureOr<InitializeResult> initializeLegacy(InitializeRequest request) {
     initialized.then((_) async {
```
jakemac53, line 50: And this logic should also move into `initialize` and out of `initializeLegacy`.

pkgs/dart_mcp/lib/src/server/server.dart
```diff
@@ -24,12 +24,13 @@ part 'tools_support.dart';
 /// Actual functionality beyond server initialization is done by mixing in
 /// additional support mixins such as [ToolsSupport], [ResourcesSupport] etc.
 abstract base class MCPServer extends MCPBase {
-  /// Completes when this server has finished initialization and gotten the
-  /// final ack from the client.
+  /// Completes when this legacy server connection has finished initialization
```
jakemac53: I think we can/should reasonably complete this future in both modes - in legacy it will happen after the ack from the client. And in the "new" mode it would complete after server registration is done (after `initialize` completes).

This gives you a chance to run some code after all mixins have had a chance to do their initialize logic. And it will also make it easier to write code that works on all versions.

pkgs/dart_mcp/lib/src/server/server.dart
```diff
@@ -42,17 +43,17 @@ abstract base class MCPServer extends MCPBase {
 
   /// The negotiated protocol version.
   ///
-  /// Only assigned after `initialize` has been called.
+  /// Only assigned after [initializeLegacy] has been called.
```
jakemac53: Same for this and below, we should be able to assign them in `initialize`.

pkgs/dart_mcp_server/pubspec.yaml
jakemac53: I would just remove this section, and actually revert the `dart_mcp_server` changes for now. The reason for that is I don't want releases of our actual mcp server to be blocked while we are working on `dart_mcp`, as I think it might be a bit before we finalize the `0.6.0` release (we will probably also want to fully migrate to the new spec in that release).

jakemac53 (commented):
This looks pretty good to me, with some comments throughout about overriding `initialize` instead of `initializeLegacy` in most places.

I also think we should revert the dart_mcp_server changes for right now, until we are ready to do a public release of dart_mcp, so that we don't get blocked on being able to do releases of the server.

We could maintain a separate branch of the repo where dart_mcp_server is migrated over to the new dart_mcp, if we want to do that in order to validate the changes.

Yusufihsangorgel:
Thanks, this clarified the intended boundary. I’ve moved the roots/elicitation setup and all shared client state into `initialize`. The hook now takes an `MCPServerInitialization` context so `protocolVersion`, `clientInfo`, and `clientCapabilities` are real in both lifecycle modes.

`initializeLegacy` now only handles negotiation and the legacy response. `initialized` remains the shared completion point: after the client acknowledgement in legacy mode, and after the full initialization and transport setup in a request-scoped mode.

I also reverted the `dart_mcp_server` migration and removed its local `dart_mcp` override, so the server continues to build against published 0.5.2. The package matrix and selected server compatibility tests are green.

pkgs/dart_mcp/test/server/tools_support_test.dart
jakemac53: This test should probably go through the actual full legacy handshake (`initializeLegacy` etc?)

jakemac53:
Just the one last comment - looking good to me though 🚀

pkgs/dart_mcp/test/server/tools_support_test.dart
Yusufihsangorgel: Thanks, Jake — updated this to exercise the full wire handshake (`initialize` → `initializeLegacy` → shared `initialize`), send the `notifications/initialized` acknowledgement, and list tools through the connection.

jakemac53 (approved):

jakemac53:
thanks @Yusufihsangorgel

Yusufihsangorgel:
> thanks @Yusufihsangorgel

Thanks for the review and guidance.

PR #528, opened 2026-07-17
add request-scoped message dispatch for MCP servers

Yusufihsangorgel:
Part of https://github.com/dart-lang/ai/issues/162.

Adds `handleRequestScopedMessage` and `MCPServerFactory`: each decoded
JSON-RPC message is served on a fresh `MCPServer` instance (following
https://github.com/dart-lang/ai/pull/524#discussion_r3582076312), which gets
`initialize(MCPServerInitialization)` with the per-request context and a
completed `initialized` before the message is delivered. The shape follows
the sketch in
https://github.com/dart-lang/ai/issues/162#issuecomment-3198530648: a server
factory plus a handle function, no new transport abstraction.

Each message round-trips through an in-memory string channel, so the exchange
runs through the exact same `Peer` validation and dispatch path as a wire
connection; a message-typed channel could drop the extra encode/decode later.
Notifications emitted while handling are surfaced through a callback so a
transport can decide how to deliver them, and server-to-client requests fail
with an error inside their handler instead of deadlocking the exchange; real
routing for those comes with the transport work.

Also makes `MCPServerInitialization.clientInfo` (and `MCPServer.clientInfo`)
nullable, since
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/3002 made
the per-request client info optional, and records the server implementation
on responses under the reserved `io.modelcontextprotocol/serverInfo` result
metadata key which the same spec change added. `RootsTrackingSupport` now
tolerates the connection closing while `listRoots` is in flight, which the
per-request teardown made easy to hit.

Tests:

- `dart_mcp`: analyze clean; format clean (dev SDK); 828/828 tests across
  chrome/vm and dart2wasm/dart2js/kernel/exe. 22 new request-scoped tests
  cover fresh-instance isolation, per-request capabilities, absent client
  info, server info stamping (including malformed metadata), notification
  passthrough, server-to-client request failure, malformed, legacy, and
  response-shaped message rejection, early shutdown, and initialization
  failure, plus a legacy handshake regression test for the nullable client
  info.
- `dart_mcp_examples` (path-overrides the local `dart_mcp`): analyze clean.

This PR does not add Streamable HTTP or any transport; wire formats, envelope
validation, and `server/discover` land separately.

### DIFF
--- a/pkgs/dart_mcp/lib/src/server/request_scoped.dart
+++ b/pkgs/dart_mcp/lib/src/server/request_scoped.dart
@@ -79,14 +79,12 @@ Future<Map<String, Object?>?> handleRequestScopedMessage(
       'A request or notification must not carry a result or error',
     );
   }
-  // Read the raw value so a non-string method fails with this argument error
-  // rather than a cast error from [JsonRpc2Object.method].
-  final method = message[Keys.method];
-  if (method is! String) {
+  final method = object.method;
+  if (method == null) {
     throw ArgumentError.value(
       message,
       'message',
-      'A dispatched message must have a string method',
+      'A dispatched message must have a method',
     );
   }
   if (object.kind == JsonRpc2Kind.request && object.id == null) {
--- a/pkgs/dart_mcp/lib/src/server/request_scoped.dart
+++ b/pkgs/dart_mcp/lib/src/server/request_scoped.dart
@@ -79,7 +79,9 @@ Future<Map<String, Object?>?> handleRequestScopedMessage(
       'A request or notification must not carry a result or error',
     );
   }
-  final method = object.method;
+  // Read the raw value so a non-string method fails with this argument error
+  // rather than a cast error from [JsonRpc2Object.method].
+  final method = message[Keys.method];
   if (method is! String) {
     throw ArgumentError.value(
       message,
--- a/pkgs/dart_mcp/lib/src/server/request_scoped.dart
+++ b/pkgs/dart_mcp/lib/src/server/request_scoped.dart
@@ -1,0 +1,221 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of 'server.dart';
+
+/// Creates the [MCPServer] which communicates over [channel].
+///
+/// The returned server must be constructed on [channel], typically by passing
+/// it to [MCPServer.fromStreamChannel].
+typedef MCPServerFactory = MCPServer Function(StreamChannel<String> channel);
+
+/// Handles one decoded JSON-RPC [message] on a fresh server instance.
+///
+/// In a request-scoped lifecycle there is no connection to initialize: every
+/// request carries its own client context, see
+/// https://modelcontextprotocol.io/specification/draft/basic. This function
+/// creates a server with [serverFactory], runs [MCPServer.initialize] with
+/// the [initialization] built by the transport, completes
+/// [MCPServer.initialized], and then delivers [message] to the server as if
+/// it had arrived over a connection. The server handles this one message and
+/// is shut down afterward; it is never reused, and no state carries over
+/// between messages. Decoding the wire format, extracting the per-request
+/// context, and anything HTTP-specific stay in the transport. Protocol
+/// metadata carried in [message]'s own `_meta` is not read here;
+/// [initialization] is the sole source of the per-request context.
+///
+/// [message] is a request if it has a non-null `id` member and a notification
+/// otherwise. Returns the decoded JSON-RPC response for a request, and `null`
+/// for a notification, which per JSON-RPC gets no response. A successful
+/// result records the server's [MCPServer.implementation] under the reserved
+/// `io.modelcontextprotocol/serverInfo` result metadata key; a result which
+/// already carries a server info entry, and every error response, is returned
+/// unchanged. If the server closes before responding to a request, an
+/// internal-error response is returned instead. The server may still be
+/// processing a notification when the returned future completes.
+///
+/// The returned future completes once the server responds or the exchange
+/// closes; it does not time out on its own. A handler that never returns
+/// leaves it pending and the server alive. To bound execution, retain the
+/// server your factory creates and call [MCPServer.shutdown] on it; the
+/// exchange then completes with an internal-error response.
+///
+/// The [ServerCapabilities] returned by [MCPServer.initialize] are
+/// intentionally not surfaced: in this lifecycle clients discover
+/// capabilities with `server/discover` rather than per message.
+///
+/// Notifications the server emits during the exchange, including any it emits
+/// while initializing, are passed to [onNotification] with their JSON-RPC
+/// envelope so a transport can decide how to deliver them. Errors thrown by
+/// [onNotification] are reported as uncaught errors and do not fail the
+/// exchange.
+///
+/// Requests from the server back to the client, such as `roots/list`, cannot
+/// be answered within a single-message exchange: they fail with an
+/// [RpcException] inside their handler, or with a [StateError] if the exchange
+/// has already been torn down.
+///
+/// Throws an [ArgumentError] if [message] is not a JSON-RPC request or
+/// notification (no string `method`, a `null` id, or a `result` or `error`
+/// member), or if its method is the legacy `initialize` request or
+/// `initialized` notification; classifying a message as legacy or
+/// request-scoped is the transport's job. Errors thrown by [serverFactory] or
+/// by [MCPServer.initialize] propagate to the caller; a server that was
+/// created is shut down first.
+// TODO: Support server-to-client requests once a transport can route them.
+// https://github.com/dart-lang/ai/issues/162
+Future<Map<String, Object?>?> handleRequestScopedMessage(
+  Map<String, Object?> message,
+  MCPServerInitialization initialization,
+  MCPServerFactory serverFactory, {
+  void Function(Map<String, Object?> notification)? onNotification,
+}) async {
+  final method = message[Keys.method];
+  if (method is! String) {
+    throw ArgumentError.value(
+      message,
+      'message',
+      'A dispatched message must have a string method',
+    );
+  }
+  if (message.containsKey(Keys.id) && message[Keys.id] == null) {
+    throw ArgumentError.value(
+      message,
+      'message',
+      'A request id must not be null',
+    );
+  }
+  if (message.containsKey(Keys.result) || message.containsKey(Keys.error)) {
+    throw ArgumentError.value(
+      message,
+      'message',
+      'A request or notification must not carry a result or error',
+    );
+  }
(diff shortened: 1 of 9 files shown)


pkgs/dart_mcp/lib/src/server/request_scoped.dart
```diff
@@ -0,0 +1,221 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of 'server.dart';
+
+/// Creates the [MCPServer] which communicates over [channel].
+///
+/// The returned server must be constructed on [channel], typically by passing
+/// it to [MCPServer.fromStreamChannel].
+typedef MCPServerFactory = MCPServer Function(StreamChannel<String> channel);
```
jakemac53, line 11: We should consider a breaking change here where we operate on a `StreamChannel<Map<String, Object?>>` instead of a `StreamChannel<String>`. We will also want to update the `stdioChannel` helper method and any other helpers we have.

Currently we end up encoding and decoding JSON within the same process which can be fairly expensive.

pkgs/dart_mcp/lib/src/server/request_scoped.dart
```diff
@@ -0,0 +1,221 @@
...
+/// created is shut down first.
+// TODO: Support server-to-client requests once a transport can route them.
+// https://github.com/dart-lang/ai/issues/162
+Future<Map<String, Object?>?> handleRequestScopedMessage(
+  Map<String, Object?> message,
+  MCPServerInitialization initialization,
+  MCPServerFactory serverFactory, {
+  void Function(Map<String, Object?> notification)? onNotification,
+}) async {
+  final method = message[Keys.method];
```
jakemac53: We have some similar logic below, I think it would be worth sharing this logic in some way.

One option could be a `JsonRpc2Object` extension type, with some subtypes `JsonRpc2Request`, `JsonRpc2Response` and `JsonRpc2Notification`.

We could also maybe have an enum or something `JsonRpc2Kind`, and a getter `kind` on `JsonRpc2Object` which would encapsulate this logic, and let you switch over the enum, casting to the subtypes etc.

We could also make this a real sealed class hierarchy of wrapper classes instead of extension types if we wanted to have some better safety guarantees.

jakemac53: nit: here and below, cast these to `Object?` to avoid dynamic and force explicit type checks/casts.

pkgs/dart_mcp/test/server/request_scoped_test.dart
```diff
@@ -0,0 +1,516 @@
...
+    test('records server info on the response', () async {
+      final harness = _DispatcherHarness();
+      final response = await harness.dispatch(
+        _callTool('probe'),
+        _initialization(),
+      );
+
+      final meta = _result(response)['_meta'] as Map?;
+      final serverInfo = Implementation.fromMap(
+        (meta!['io.modelcontextprotocol/serverInfo'] as Map)
```
jakemac53: nit: use the key constant here

Same for other tests

pkgs/dart_mcp/test/server/request_scoped_test.dart
```diff
@@ -0,0 +1,516 @@
...
+    });
+
+    test('records server info on the response', () async {
+      final harness = _DispatcherHarness();
+      final response = await harness.dispatch(
+        _callTool('probe'),
+        _initialization(),
+      );
+
+      final meta = _result(response)['_meta'] as Map?;
```
jakemac53: nit: might as well cast directly to `Map` here and avoid the `!` on the next line

pkgs/dart_mcp/test/server/request_scoped_test.dart
```diff
@@ -0,0 +1,516 @@
...
+      final serverInfo = Implementation.fromMap(
+        (meta!['io.modelcontextprotocol/serverInfo'] as Map)
+            .cast<String, Object?>(),
```
jakemac53: nit: just cast to `Map<String, Object?>` above, and ensure the maps are created with that type in the harness if they aren't already. Real JSON maps always have that type.

Same for the other tests

pkgs/dart_mcp/test/server/request_scoped_test.dart
```diff
@@ -0,0 +1,516 @@
...
+        ),
+      );
+      await harness.dispatch(_callTool('probe'), _initialization());
+
+      expect(harness.servers, hasLength(2));
+      final first = harness.servers[0];
+      final second = harness.servers[1];
+      expect(first, isNot(same(second)));
+      expect(first.initializedWith?.clientCapabilities.roots, isNotNull);
+      expect(second.initializedWith?.clientCapabilities.roots, isNull);
```
jakemac53: The servers themselves should be recording the client capabilities during initialization (even per request), it would be better to directly read that to ensure it is getting set, something like:

```suggestion
      expect(first.clientCapabilities.roots, isNotNull);
      expect(second.clientCapabilities.roots, isNull);
```

jakemac53:
I am generally wondering how we are going to want to deal with roots in the new version, since we already know we can't rely on roots/list being sensible and need a fallback, but the fallback requires some kind of stateful session.

I don't think that matters for this PR, just something I was thinking about while reviewing it.

Yusufihsangorgel:
Addressed the review comments in 87d0746:

- Added a `JsonRpc2Object` extension type with a `JsonRpc2Kind` getter, moved the outbound-frame classification onto it, and rewrote the argument validation in terms of it - the kind logic now mirrors json_rpc_2's own routing (result/error first). Kept it package-internal for now; happy to export it if you want it public.
- The raw-map indexing goes through typed casts everywhere now - no dynamic left.
- Tests use the `Keys` constants, cast directly to `Map<String, Object?>`, and read `server.clientCapabilities` directly (the recording field is gone).

pkgs/dart_mcp/lib/src/server/request_scoped.dart
Yusufihsangorgel: Agreed - it would also drop the encode/decode round-trip this PR leaves as a TODO. It touches `MCPBase`, `stdioChannel`, and the client helpers, so I'd take it as the immediate next PR unless you'd rather have it here.

Yusufihsangorgel:
I think the RC splits this in two. The ask-the-client path is stateless now: server-to-client requests (roots/list included) became MRTR embeddings (SEP-2322), the server hands the roots/list back inside the result with an opaque `requestState` and the client retries, so that round trip needs no shared state across instances. It's the fallback that has no home, like you said: `requestState` only spans the one retry, and with protocol sessions gone there is no key left to hang a cross-call roots cache on. My read of the deprecation notes (SEP-2577) is that this is intentional: roots become per-call input, tool parameters or config, and generic cross-call state moves to server-minted handles in tool args, SEP-2567 style. fwiw the TS and Python SDKs already landed there: TS hard-fails a push `listRoots()` in the new model and Python folds it into the `InputRequiredResult` retry; neither keeps a server-side roots cache.

jakemac53:
The problem with no stateful roots is that we actually want them to be persistent because that allows us to keep a live LSP server open to those roots (faster analysis, less duplicate work per request).

We can do this somewhat separately - keep that on the side and just deactivate inactive roots (we already do this anyways, so its really just moving code around).

Maybe that would anyways be better - we already require roots to be given per request that needs them, and we can just drop the add/remove root functions entirely, and treat any root given by any request as an `addRoot` call essentially. And then just dispose of roots as we see fit based on timeouts/LRU etc.

pkgs/dart_mcp/lib/src/server/request_scoped.dart
jakemac53: Sure that is fine with me

pkgs/dart_mcp/lib/src/utils/json_rpc_2_object.dart
```diff
@@ -0,0 +1,40 @@
...
+  /// notifications.
+  JsonRpc2Kind get kind =>
+      _value.containsKey(Keys.result) || _value.containsKey(Keys.error)
+          ? JsonRpc2Kind.response
+          : _value.containsKey(Keys.id)
+          ? JsonRpc2Kind.request
+          : JsonRpc2Kind.notification;
+
+  /// The method of this message, if it is a request or notification.
+  Object? get method => _value[Keys.method];
```
jakemac53: This should be typed as `String?` I think, and then it would be good to make the cast explicit instead of implicit.

pkgs/dart_mcp/lib/src/utils/json_rpc_2_object.dart
```diff
@@ -0,0 +1,40 @@
...
+
+  /// The id of this message, if it is a request or response.
```
jakemac53, line 35: Probably worth a comment that this should be a String or int.

pkgs/dart_mcp/lib/src/utils/json_rpc_2_object.dart
Yusufihsangorgel: Done - typed it `String?` with the explicit cast. The dispatcher's validation still reads the raw value first, so a non-string `method` keeps failing with the `ArgumentError` rather than a cast error.

Yusufihsangorgel:
Agreed, and treating per-request roots as the input is what makes dropping add/remove safe rather than just tidier: the warm-LSP pool becomes a cache derived from those roots, not state you have to keep consistent. Timeout/LRU eviction can then only cost a cold re-analysis the next time a request names that root, never a wrong answer. It also makes the pool root-keyed rather than session-keyed, so two clients on the same workspace share one warm analyzer instead of each paying for their own. What you lose is the full root set up front, since a root only warms once a request touches it, but lazy warm-up seems fine for analysis and it turns the fallback into a cache miss instead of a consistency problem.

pkgs/dart_mcp/lib/src/utils/json_rpc_2_object.dart
Yusufihsangorgel: Done, added a note that it's a `String` or an `int`.

pkgs/dart_mcp/lib/src/server/request_scoped.dart
```diff
@@ -79,7 +79,9 @@ Future<Map<String, Object?>?> handleRequestScopedMessage(
       'A request or notification must not carry a result or error',
     );
   }
-  final method = object.method;
+  // Read the raw value so a non-string method fails with this argument error
```
jakemac53: I would just move the nicer error into the `method` getter, since this just throws anyways.

jakemac53 (commented):
just the one last comment

pkgs/dart_mcp/lib/src/server/request_scoped.dart
Yusufihsangorgel: Done - the getter throws the `ArgumentError` for a non-string method now, and the dispatcher only checks for a missing one.

jakemac53 (approved):

PR #531, opened 2026-07-18
operate on decoded messages instead of JSON strings

Yusufihsangorgel:
Follow-up to [this thread](https://github.com/dart-lang/ai/pull/528#discussion_r3604241409) in #528: `MCPBase`, `MCPClient.connectServer`, `MCPServerFactory`, and `stdioChannel` now operate on `StreamChannel<Map<String, Object?>>` instead of `StreamChannel<String>`, so JSON is encoded and decoded only at the transport edge rather than once more within the process.

- `MCPBase` hands the channel to `Peer.withoutJson`. The channel type admits only JSON objects, so json_rpc_2 never receives a batch and never emits the `List` frames its batch support would answer one with.
- `stdioChannel` still speaks newline-delimited JSON on the wire and now returns the decoded channel. The string-to-map adaptation is exposed separately as `jsonRpcChannel` for other `StreamChannel<String>` transports.
- Behavior deltas worth calling out:
  - Frames which are not valid JSON, or which decode to something other than a JSON object, are answered at the transport edge by `jsonRpcChannel` instead of by json_rpc_2 inside the peer. The two existing invalid-JSON tests moved to `stdio_test.dart` accordingly, and neither those frames nor their error responses appear in `protocolLogSink` anymore (which now logs each message re-encoded as JSON, falling back to `toString` when a message cannot be encoded). If wire-level visibility of rejected frames turns out to matter for debugging, an optional log sink on `jsonRpcChannel` would be a small follow-up.
  - JSON-RPC batch frames now get a single invalid request error response with a null id and a message pointing at the 2025-06-18 removal. Batching was added in protocol version 2025-03-26 and removed in 2025-06-18; this package never sent batches itself.
  - `handleRequestScopedMessage` now stamps server info onto a copy of the response. With decoded channels the response embeds the very maps the handler returned — which may be retained by the handler or even unmodifiable, like `EmptyResult()`'s — where previously it saw private `jsonDecode` copies.
- The URL elicitation retry path now restores a typed view of the error data itself. `RpcException.serialize` copies the data into an untyped map, and the JSON round trip previously re-laundered it to `Map<String, dynamic>` before user code saw it.

Does not add any transport or change protocol behavior beyond the frame type: HTTP, `server/discover`, and roots are untouched. Since 0.6.0-wip has never been published, current `dart_mcp` users only see this when they opt into 0.6.0; `dart_mcp_server` pins `^0.5.0` and migrates with its own 0.6.0 dependency bump.

## Tests

- New `stdio_test.dart` covers `jsonRpcChannel` (decode/encode round trip, parse error and non-object frames with null ids, batch rejection, and a server and a client each surviving invalid frames) plus `stdioChannel` at the byte level.
- New request-scoped tests pin that server info is stamped onto unmodifiable results (the built-in `ping` path) without modifying the map the handler returned, and skipped for metadata with non-string keys.
- A new test pins that a server survives a map which is not a valid message, now representable at the public channel seam.
- `dart test -p chrome,vm -c dart2wasm,dart2js,kernel,exe` passes (218 tests in each of the four configurations).
- `dart analyze --fatal-infos` and `dart format` (dev SDK) are clean here, and in `mcp_examples` resolved against this branch.

Part of #162.

### DIFF
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -335,8 +335,10 @@ base class ServerConnection extends MCPBase {
       if (_elicitationUrlSupport?.autoHandleUrlElicitationRequired == true &&
           e.code == McpErrorCodes.urlElicitationRequired &&
           data is Map) {
-        // json_rpc_2 copies the error data into an untyped map when it
-        // serializes the exception, so restore the typed view.
+        // `RpcException.serialize` spreads the error data into an untyped map
+        // literal (adding a `request` key), so the map arriving here is not a
+        // `Map<String, Object?>` and a representation type check against
+        // `ElicitRequest` would fail. Restore the typed view instead.
         final elicitRequest = data.cast<String, Object?>() as ElicitRequest;
         final elicitationComplete =
             elicitRequest.onElicitationComplete(this).firstOrNull;
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -331,12 +331,13 @@ base class ServerConnection extends MCPBase {
       // If we are set up to try and auto handle url elicitation and we get
       // an error that the url elicitation is required, we will try and handle
       // it and then retry the request a single time.
+      final data = e.data;
       if (_elicitationUrlSupport?.autoHandleUrlElicitationRequired == true &&
-          e.code == McpErrorCodes.urlElicitationRequired) {
+          e.code == McpErrorCodes.urlElicitationRequired &&
+          data is Map) {
         // json_rpc_2 copies the error data into an untyped map when it
         // serializes the exception, so restore the typed view.
-        final elicitRequest =
-            (e.data as Map).cast<String, Object?>() as ElicitRequest;
+        final elicitRequest = data.cast<String, Object?>() as ElicitRequest;
         final elicitationComplete =
             elicitRequest.onElicitationComplete(this).firstOrNull;
         final elicitResult = await _elicitationUrlSupport!.handleElicitation(
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -74,8 +74,7 @@ base class MCPClient {
   /// Returns a connection for an MCP server using a [channel], which is already
   /// established.
   ///
-  /// Each [String] sent over [channel] represents an entire JSON request or
-  /// response.
+  /// Each map sent over [channel] is a single decoded JSON-RPC message.
   ///
   /// If [protocolLogSink] is provided, all messages sent on [channel] will be
   /// forwarded to that [Sink] as well, with `<<<` preceding incoming messages
@@ -85,7 +84,7 @@ base class MCPClient {
   /// To perform cleanup when this connection is closed, use the
   /// [ServerConnection.done] future.
   ServerConnection connectServer(
-    StreamChannel<String> channel, {
+    StreamChannel<Map<String, Object?>> channel, {
     Sink<String>? protocolLogSink,
   }) {
     // For type promotion in this function.
@@ -334,7 +333,10 @@ base class ServerConnection extends MCPBase {
       // it and then retry the request a single time.
       if (_elicitationUrlSupport?.autoHandleUrlElicitationRequired == true &&
           e.code == McpErrorCodes.urlElicitationRequired) {
-        final elicitRequest = e.data as ElicitRequest;
+        // json_rpc_2 copies the error data into an untyped map when it
+        // serializes the exception, so restore the typed view.
+        final elicitRequest =
+            (e.data as Map).cast<String, Object?>() as ElicitRequest;
         final elicitationComplete =
             elicitRequest.onElicitationComplete(this).firstOrNull;
         final elicitResult = await _elicitationUrlSupport!.handleElicitation(
--- a/pkgs/dart_mcp/lib/src/server/request_scoped.dart
+++ b/pkgs/dart_mcp/lib/src/server/request_scoped.dart
@@ -211,8 +211,11 @@ Map<String, Object?> _withServerInfo(
   if (existingMeta != null && existingMeta.containsKey(Keys.serverInfoMeta)) {
     return response;
   }
-  // The response embeds the very maps the handler returned, which may be
-  // retained by the handler or even unmodifiable, so stamp copies of them.
+  // With decoded channels there is no decode step between the server and this
+  // dispatcher, so `result` can be the very map a handler returned. Handlers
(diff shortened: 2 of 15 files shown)


pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -334,7 +333,10 @@ base class ServerConnection extends MCPBase {
       // it and then retry the request a single time.
       if (_elicitationUrlSupport?.autoHandleUrlElicitationRequired == true &&
           e.code == McpErrorCodes.urlElicitationRequired) {
```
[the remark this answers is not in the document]
Yusufihsangorgel: Done in afe28a1. Note `e.data` can't actually be null here (`RpcException.serialize` wraps null data in a map carrying the original request), but any other non-map data passes through as is, so the guard is worthwhile; added a test.

pkgs/dart_mcp/lib/src/shared.dart
```diff
@@ -191,15 +198,15 @@ base class MCPBase {
...
           ),
         )
         .transformSink(
           StreamSinkTransformer.fromHandlers(
             handleData: (data, sink) {
-              protocolLogSink.add('>>> ($name) $data\n');
+              protocolLogSink.add('>>> ($name) ${jsonEncode(data)}\n');
               sink.add(data);
             },
           ),
```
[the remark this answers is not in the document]
Yusufihsangorgel: Done in add6d6d, with a `toString` fallback and a test. Kept it as a local helper rather than duplicating the try/catch in both handlers.

pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -332,9 +331,13 @@ base class ServerConnection extends MCPBase {
       // If we are set up to try and auto handle url elicitation and we get
       // an error that the url elicitation is required, we will try and handle
       // it and then retry the request a single time.
+      final data = e.data;
       if (_elicitationUrlSupport?.autoHandleUrlElicitationRequired == true &&
-          e.code == McpErrorCodes.urlElicitationRequired) {
-        final elicitRequest = e.data as ElicitRequest;
+          e.code == McpErrorCodes.urlElicitationRequired &&
+          data is Map) {
```
jakemac53, line 337: nit: can this be a `data is ElicitRequest`? Are we getting maps with non-String key types here?

That would be easier to read, and we wouldn't need any explicit casts any more, relying on type promotion instead

pkgs/dart_mcp/lib/src/server/request_scoped.dart
```diff
@@ -200,26 +194,35 @@ Map<String, Object?> _errorResponse(Object? id, String message) => {
...
+  if (result is! Map<String, Object?>) return response;
+  final existingMeta = result[Keys.meta];
+  if (existingMeta is! Map<String, Object?>?) return response;
+  if (existingMeta != null && existingMeta.containsKey(Keys.serverInfoMeta)) {
+    return response;
+  }
+  // The response embeds the very maps the handler returned, which may be
+  // retained by the handler or even unmodifiable, so stamp copies of them.
+  return {
+    ...response,
```
jakemac53, line 220: This will copy the entire map, instead we should just modify it in place for efficiency, unless we require a copy.

I would go ahead and use this time to make that more clear in the function definition though, by making it return `void` and then add a comment that it modifies `response` by injecting the server info metadata.

pkgs/dart_mcp/lib/stdio.dart
```diff
@@ -6,22 +6,79 @@ import 'dart:async';
...
+/// helper does not provide framing.
+///
+/// Frames which are not valid JSON, or which decode to something other than a
+/// JSON object, are answered on [channel] with a JSON-RPC error response and
+/// never surface on the returned channel. This includes JSON-RPC batch
+/// frames, which MCP removed in protocol version 2025-06-18.
+StreamChannel<Map<String, Object?>> jsonRpcChannel(
+  StreamChannel<String> channel,
+) {
+  void answer(int code, String message, Object? request) {
```
jakemac53: nit: rename this `answerWithError` or similar, as its just for errors.

pkgs/dart_mcp/lib/stdio.dart
```diff
@@ -6,22 +6,79 @@ import 'dart:async';
...
+    channel.sink.add(
+      jsonEncode(RpcException(code, message).serialize(request)),
     );
+  }
+
+  final stream =
+      channel.stream
+          .map<Object?>(jsonDecode)
+          .handleError((Object error) {
+            final formatException = error as FormatException;
```
jakemac53: I would prefer a type check here with a fallback, just in case

jakemac53 (commented):
Just some minor comments, looks good

pkgs/dart_mcp/lib/src/client/client.dart
Yusufihsangorgel: Yes, unfortunately: `RpcException.serialize` spreads the error data into an untyped map literal (adding a `request` key), so over the in-memory transport `data` arrives here as a `Map<dynamic, dynamic>`. The keys are all Strings at runtime, but `is ElicitRequest` checks the reified type and is false, and the auto-handle test fails with the promotion version. Over stdio it would pass only because `jsonDecode` re-types the keys. Kept the cast and expanded the comment in fc34821 to say where the untyped map comes from.

pkgs/dart_mcp/lib/src/server/request_scoped.dart
Yusufihsangorgel: We do need the copy here: with decoded channels there is no decode step between the per-request server and this dispatcher, so `result` can be the very map a handler returned. Handlers may retain that map, and for the built-in ping handler it is `EmptyResult()`, which is backed by a const map, so modifying in place throws (there is a test covering that case). Expanded the comment in fc34821 to spell out the aliasing, which is also why it still returns the map rather than `void`.

pkgs/dart_mcp/lib/stdio.dart
Yusufihsangorgel: Done in fc34821.

pkgs/dart_mcp/lib/stdio.dart
Yusufihsangorgel: Done in fc34821, with `Error.throwWithStackTrace` to keep the original stack trace (`only_throw_errors` also flags a bare `throw` of an `Object`).

pkgs/dart_mcp/lib/src/client/client.dart
jakemac53: Sounds good, thanks

pkgs/dart_mcp/lib/src/server/request_scoped.dart
jakemac53: Ah yeah that makes sense, SGTM

jakemac53 (approved):

PR #540, opened 2026-07-22
update the github repo link to the new location

jakemac53:


mit-mit (approved):

pkgs/skills/lib/src/core/version.dart
```diff
@@ -2,4 +2,4 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-const String version = '1.0.0-beta.3';
```
mit-mit, line 5: Should we have a CI job that catches when we forget to keep this in sync with the pubspec?

jakemac53: We do, which is why I remembered to add it :)

PR #544, opened 2026-07-23
use the dart/workspace/analysis/complete request where possible

### LINKED ISSUE #499: Analyze tool is stuck?
Image

The MCP tool started way before I thought about comparing against the terminal version. I'll fully write this issue, and the tool will not complete (on a basically empty project).

How can I help identify the cause of the problem?

### LINKED ISSUE #536: Migrate to new dart/workspace/analysis/complete request
See https://github.com/dart-lang/sdk/issues/63772 for context, should resolve https://github.com/dart-lang/ai/issues/499 hopefully

jakemac53:
Closes https://github.com/dart-lang/ai/issues/536
Closes https://github.com/dart-lang/ai/issues/499

Also preps for a release, we have a decent chunk of functionality ready now.

cc @DanTup

### DIFF
--- a/pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
+++ b/pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
@@ -47,12 +47,16 @@ base mixin DartAnalyzerSupport
   /// The current diagnostics for a given file.
   Map<Uri, List<lsp.Diagnostic>> diagnostics = {};
 
-  /// If currently analyzing or waiting on first analysis, a completer which
-  /// will be completed once analysis is over.
-  Completer<void>? _doneAnalyzing = Completer();
+  /// LEGACY MODE ONLY: If currently analyzing or waiting on first analysis, a
+  /// completer which will be completed once analysis is over based on legacy
+  /// analysis events.
+  ///
+  /// Always use [_waitForAnalysisToComplete] instead reading this directly.
+  Completer<void>? _legacyDoneAnalyzing = Completer();
 
-  /// Completes the next time we get an analysis start event.
-  Completer<void> _analysisStart = Completer();
+  /// LEGACY MODE ONLY: Completes the next time we get a legacy analysis start
+  /// event.
+  Completer<void> _legacyAnalysisStart = Completer();
 
   /// The future for the current LSP server initialization.
   Future<String?>? _lspInitialization;
@@ -625,9 +629,9 @@ base mixin DartAnalyzerSupport
   /// `dart/workspace/analysis/complete` request (if supported) or by relying
   /// on `$/analyzerStatus` push notifications.
   ///
-  /// If [waitForSecondaryAnalysisInLegacyMode] is `true` and we are operating in legacy
-  /// mode, we will wait up to one second for a new analysis to start before
-  /// returning, unless we are already analyzing.
+  /// If [waitForSecondaryAnalysisInLegacyMode] is `true` and we are operating
+  /// in legacy mode, we will wait up to one second for a new analysis to start
+  /// before returning, unless we are already analyzing.
   Future<void> _waitForAnalysisToComplete(
     Peer analysisServer, {
     bool waitForSecondaryAnalysisInLegacyMode = false,
@@ -635,15 +639,16 @@ base mixin DartAnalyzerSupport
     if (_supportsWorkspaceAnalysisComplete) {
       await analysisServer.sendRequest('dart/workspace/analysis/complete');
     } else {
-      if (waitForSecondaryAnalysisInLegacyMode && _doneAnalyzing == null) {
+      if (waitForSecondaryAnalysisInLegacyMode &&
+          _legacyDoneAnalyzing == null) {
         // Wait a bit for the new analysis to start if not currently
         // analyzing.
-        await _analysisStart.future.timeout(
+        await _legacyAnalysisStart.future.timeout(
           const Duration(seconds: 1),
           onTimeout: () {},
         );
       }
-      await _doneAnalyzing?.future;
+      await _legacyDoneAnalyzing?.future;
     }
   }
 
@@ -672,7 +677,7 @@ base mixin DartAnalyzerSupport
 
     // Reset analysis related state.
     diagnostics.clear();
-    _doneAnalyzing = Completer<void>();
+    _legacyDoneAnalyzing = Completer<void>();
     _currentWorkspaceFolders.clear();
 
     server?.kill();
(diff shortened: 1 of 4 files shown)


pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -235,6 +238,12 @@ base mixin DartAnalyzerSupport
     }
 
     if (initializeResult != null) {
+      final experimental = initializeResult.capabilities.experimental;
+      if (experimental is Map &&
+          experimental.containsKey('workspaceAnalysisComplete')) {
+        _supportsWorkspaceAnalysisComplete = true;
+      }
```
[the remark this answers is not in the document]
jakemac53: This is really not necessary in this case

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -350,15 +359,10 @@ base mixin DartAnalyzerSupport
...
-          await _analysisStart.future.timeout(
-            const Duration(seconds: 1),
-            onTimeout: () {},
-          );
-        }
-        await _doneAnalyzing?.future;
+        await _waitForAnalysisToComplete(
+          analysisServer,
+          debounceDelay: const Duration(seconds: 1),
+        );
```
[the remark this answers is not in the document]
jakemac53: I think its better how it is

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -617,6 +621,24 @@ base mixin DartAnalyzerSupport
...
+      // Unconditionally wait for the debounce delay if provided.
+      await Future<void>.delayed(debounceDelay);
+    }
+
+    if (_supportsWorkspaceAnalysisComplete) {
+      await analysisServer.sendRequest('dart/workspace/analysis/complete');
+    } else {
+      await _doneAnalyzing?.future;
+    }
+  }
```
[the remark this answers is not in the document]
jakemac53: The problem here is the analyzer may need a bit to detect file changes and start analysis. The one second delay is a hack to be sure but its as good as we can do.

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -350,15 +359,10 @@ base mixin DartAnalyzerSupport
         // as a confirmation to the LLM that it was respected.
         messages.add(TextContent(text: 'Applied quick fixes'));
 
-        if (_doneAnalyzing == null) {
-          // Wait a bit for the new analysis to start if not currently
-          // analyzing.
-          await _analysisStart.future.timeout(
-            const Duration(seconds: 1),
-            onTimeout: () {},
-          );
```
DanTup, line 359: Is this timeout still required for if the SDK is older and doesn't support the new request? (I don't know how likely it is that someone gets this version of the MCP server with an older SDK, but if they do, this may be a change in behaviour?).

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -617,6 +621,24 @@ base mixin DartAnalyzerSupport
...
+  }) async {
+    if (debounceDelay != null) {
+      // Unconditionally wait for the debounce delay if provided.
+      await Future<void>.delayed(debounceDelay);
+    }
+
+    if (_supportsWorkspaceAnalysisComplete) {
+      await analysisServer.sendRequest('dart/workspace/analysis/complete');
+    } else {
+      await _doneAnalyzing?.future;
```
DanTup: Is this the only code that uses `_doneAnalyzing`? I'd like to remove the legacy status notifications from the server soon, which means this will never complete. It may be worth adding a comment to `_doneAnalyzing` to make that clearer.

(If you do still need it for something else, we should switch it to the standard LSP progress notifications - like I did in https://dart-review.googlesource.com/c/sdk/+/526621 for the benchmarks)

DanTup (approved):

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: Technically we have an unconditional 1 second delay now where previously we might not have done that or it might have been less.

I guess this is actually what the Gemini comment was getting at as well, and I could probably avoid the extra delay here.

DanTup: I think your response to Gemini was reasonable - if you're modifying on disk, there is a chance that you call the server before the watcher event was processed. I think the 1s delay isn't a bad idea.

But I realise now my comment may have been wrong. I thought the timeout was on waiting for analysis (eg. if the server never produced any analyzing status, you'd continue anyway after 1s), however I realise now it was on the start event and not the end event (so probably if the server didn't emit any status events, it wouldn't have helped).

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
jakemac53: This path is only for the legacy mode now anyways, so I think just adding comments to it should be sufficient. On newer analyzers, we should send the new request instead.

I also renamed the variables to `legacy*` to drive the point home.

DanTup: sgtm :)

(to answer your edited question, support was added at least 6 years ago - that's when we marked the legacy status notification as deprecated! 😄)

pkgs/dart_mcp_server/lib/src/mixins/analyzer.dart
```diff
@@ -41,15 +41,22 @@ base mixin DartAnalyzerSupport
...
   Map<Uri, List<lsp.Diagnostic>> diagnostics = {};
 
-  /// If currently analyzing or waiting on first analysis, a completer which
-  /// will be completed once analysis is over.
-  Completer<void>? _doneAnalyzing = Completer();
+  /// LEGACY MODE ONLY: If currently analyzing or waiting on first analysis, a
+  /// completer which will be completed once analysis is over based on legacy
+  /// analysis events.
+  ///
+  /// Always use [_waitForAnalysisToComplete] instead reading this directly.
```
bwilkerson: nit: "instead reading" --> "instead of reading"

bwilkerson (approved):

DanTup:
@jakemac53 FYI, I don't think it will affect you (because I think you delay the server shutdown) but I found an issue while working on migrating `dart analyze` to use this request.

When the server shuts down, we destroy analysis contexts, and this triggers clearing out all existing diagnostics (this makes sense in the case where you change the workspace roots, because we want to drop diagnostics from the removed roots).

This means if you:

- start collecting all diagnostics
- tell the server to shut down
- wait for the server process to exit
- examine the diagnostics

You'll end up with 0 diagnostics, because during shutdown we wiped out those that were send during analysis. If you're not shutting the server down while still collecting the diagnostics, you wouldn't be affected by it.

I have a fix at https://dart-review.googlesource.com/c/sdk/+/528020 that suppresses diagnostics flushes during shutdown.

PR #564, opened 2026-07-23
fix(skills): Do not save repositories in skills add when git clone fails (fixes #561)

### LINKED ISSUE #561: Passing a bad URL to `skills add` leaves you in a bad state
From the bug bash, thanks @sigmundch !

```
🟥 friction point: recovering from user errors is not ideal. Here is an example:
skill add boo where boo is an unresolved package
At first this produces a reasonable error:

Cloning boo...
Warning: Failed to clone boo: fatal: repository 'boo' does not exist

No skills found in the given source boo.


However, this still adds it to the list of skills, which results in odd behavior in other commands. For example:
skills get now fails with the same error
skills list shows boo with an empty list underneath
skills remove can't be used to fix the problem - when selecting boo it notices that there are no skills to remove and aborts.
ditto skills prune also doesn't remove anything.
It's hard to get out of this broken state
At this point my CLI appears busted
deleting .agent/skills under the package doesn't fix the problem either.
deleting .config/dart_skills/skills_config.json did the trick!
```

jakemac53:
Fixes #561

### Description
In `skills add`, repositories were previously saved to `GlobalConfig` or local `SkillManifest` files before verification/installation of skills took place. If git clone failed (e.g. invalid repository URL or non-existent remote repository), the invalid repository was still recorded in configuration files, leaving the CLI in a broken state where future commands would continually attempt to sync the non-existent repository.

This change performs a pre-sync verification check in `AddCommand.run()` prior to updating configuration/manifest files. Any repository that fails to clone or sync is excluded and an error is logged. If all provided repositories fail, `AddCommand` bails out early without saving any entries.

### DIFF
--- a/pkgs/skills/lib/src/commands/add_command.dart
+++ b/pkgs/skills/lib/src/commands/add_command.dart
@@ -7,6 +7,7 @@ import 'package:skills/src/models/skill_manifest.dart';
 import '../core/dialog_support.dart';
 import '../core/git_repos.dart';
 import '../core/git_runner.dart';
+import '../core/git_sync.dart';
 import '../models/global_config.dart';
 
 import 'options.dart';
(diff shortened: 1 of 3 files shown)


pkgs/skills/lib/src/commands/add_command.dart
```diff
@@ -78,13 +79,33 @@ class AddCommand extends SkillsCommand {
     );
     if (agents.isEmpty) return;
 
+    final validRepos = <GitRepo>[];
+    if (await gitRunner.isAvailable) {
```
sigmundch: What could make the gitRunner to be unavailable? Alternatively, what if we don't add anything to validRepos when it is unavailable? 

(basically I'm primarily wondering if the else case here is still problematic and we'd rather issue an error in that case instead.

pkgs/skills/test/commands/add_command_test.dart
```diff
@@ -148,5 +148,37 @@ void main() {
...
+        expect(
+          manifest
+              .sourceUrisForAgent('cursor')
+              .containsKey(
+                'https://github.com/bad/nonexistent_repo_for_test.git',
+              ),
+          isFalse,
+        );
+      },
+    );
```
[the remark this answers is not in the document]
jakemac53: The whole point is that it fails anyways, so I don't think this is a problem :)

sigmundch (approved):
Thank you! One question/comment below, but I'm OK landing as it is as well.

pkgs/skills/lib/src/commands/add_command.dart
jakemac53: This really just means `git` is not installed.

jakemac53: But yeah good point, it should just error I think

sigmundch (approved):
thanks!

PR #565, opened 2026-07-24
add Result.resultType from protocol 2026-07-28

Yusufihsangorgel:
First slice of the 2026-07-28 core-protocol support tracked in #162, split out from the HTTP transport work to keep it small and reviewable on its own.

Adds the `resultType` result field as an additive, non-breaking API:

- `Result.resultType` reads the field. A null value means the server did not send it; callers treat null as `complete`, the backward-compatible convention the 2026-07-28 schema requires for results from servers on earlier protocol versions. See https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2726, which reframes an absent `resultType` as backward compatibility.
- The getter is a nullable `String`, exactly matching `CreateMessageResult.stopReason` (both `String?`), the open union this package already has. The schema types `resultType` as `"complete" | "input_required" | string`, so a server may send a value this package does not know. A closed `enum` cannot represent that set; it would have to throw on an unknown value or fold it into a fallback constant, and both lose the string the server actually sent. See https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2773, the SEP-2322 fix that made `ResultType` an open union so servers can extend it.
- No `latestSupported` bump and no server-side emission; those are later slices. This is the read side, so a client parses `resultType` from a 2026-07-28 server correctly without waiting for the rest.

## Tests

- `resultType` is null when the field is absent, and passes both a known value and an unknown string through unchanged.
- `dart test -p chrome,vm -c dart2wasm,dart2js,kernel,exe` passes (219 tests in each of the four configurations).
- `dart analyze --fatal-infos` is clean here and in `mcp_examples` resolved against this branch. The diff only adds lines, so existing formatting is untouched.

Part of #162.

### DIFF
--- a/pkgs/dart_mcp/lib/src/api/api.dart
+++ b/pkgs/dart_mcp/lib/src/api/api.dart
@@ -180,6 +180,14 @@ extension type Notification(Map<String, Object?> _value) {
 /// Base interface for all responses to requests.
 extension type Result._(Map<String, Object?> _value) {
   Meta? get meta => _value[Keys.meta] as Meta?;
+
+  /// The type of this result, which determines how to parse it.
+  ///
+  /// Known types are "complete" and "input_required", but a server may send
+  /// any other string. A null value means the server did not send the field;
+  /// treat it as "complete", as the schema requires for results from servers
+  /// on protocol versions before 2026-07-28.
+  String? get resultType => _value[Keys.resultType] as String?;
 }
 
 /// A response that indicates success but carries no data.
--- a/pkgs/dart_mcp/lib/src/api/api.dart
+++ b/pkgs/dart_mcp/lib/src/api/api.dart
@@ -184,10 +184,10 @@ extension type Result._(Map<String, Object?> _value) {
   /// The type of this result, which determines how to parse it.
   ///
   /// Known types are "complete" and "input_required", but a server may send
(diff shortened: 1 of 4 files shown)


pkgs/dart_mcp/lib/src/api/api.dart
```diff
@@ -180,6 +180,14 @@ extension type Notification(Map<String, Object?> _value) {
...
 extension type Result._(Map<String, Object?> _value) {
   Meta? get meta => _value[Keys.meta] as Meta?;
+
+  /// The type of this result, which determines how to parse it.
+  ///
+  /// Known types are "complete" and "input_required", but a server may send
+  /// any other string. A null value means the server did not send the field;
+  /// treat it as "complete", as the schema requires for results from servers
+  /// on protocol versions before 2026-07-28.
+  String? get resultType => _value[Keys.resultType] as String?;
```
jakemac53: Should we just default this to complete?

```suggestion
  String get resultType => _value[Keys.resultType] as String? ?? 'complete';
```

Yusufihsangorgel: Done. Also updated the doc comment and the absent-case test, which both assumed null.

jakemac53 (approved):

PR #566, opened 2026-07-24
Support hyphens in package name prefixes when matching and creating skills

### LINKED ISSUE #541: Support matching package names using hyphens instead of underscores
The skills specification dictates hyphens, but package names have underscores.

This package requires an exact package name match as a prefix, which is somewhat nice in terms of disambiguating parts of the skill name, but violates the spec.

We should probably support either - and possibly push people towards the hyphen approach.

jakemac53:
Fixes #541.

- Support matching package names using hyphens instead of underscores in package name prefixes ( as well as ).
- Update `skills create` to generate hyphenated package prefixes ().
- Update documentation and unit tests.

### DIFF
--- a/pkgs/skills/CHANGELOG.md
+++ b/pkgs/skills/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add a --version flag.
 - JSON encode descriptions in the `create` command.
+- Support hyphens instead of underscores in package name prefixes when matching and creating skills.
 
 ## 1.0.0-beta.4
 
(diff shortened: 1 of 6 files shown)


reidbaker (approved):
I tried to think through a case where a valid dart package when running replace _ with - would cause you to have an invalid skill name  (start or end with a hyphen, double hyphen) but best I can tell the combination of rules wont let that happen.

jakemac53:
> I tried to think through a case where a valid dart package when running replace _ with - would cause you to have an invalid skill name (start or end with a hyphen, double hyphen) but best I can tell the combination of rules wont let that happen.

Yeah they can only match ` [a-z0-9_]` so the trivial solution here should be fine

PR #568, opened 2026-07-24
feat(skills): clean up unused git repo sources in skills prune and skills get

### LINKED ISSUE #562: No way to remove a git repo skill source once added
From @sigmundch (thanks!)

```
🟨 minor friction point: similar to the above, using skill add dart-lang/test makes note that the
repo exists and has no skills (yay!) however, skills list still shows it and skills remove can't make
the entry disappear from skills list. Unlike the above, skills get does work and once you add the 
default skills from dart-lang/skills, this entry gets removed from the config.
```

jakemac53:
Resolves #562

### Summary of Changes

- **Prune Utility (`lib/src/core/pruner.dart`)**:
  - Extracted package dependency skill pruning and git repository source cleanup into a reusable `pruneSkills` function.
  - Checks both local (`skills_config.json`) and global (`global_config.json`) configurations for git repo sources with 0 installed skills.
  - Presents candidate unreferenced package dependencies and empty local git repo sources to the user in a unified multi-select dialog (`showMultiSelectDialog`).
  - Presents candidate empty global git repo sources in a multi-select dialog.
  - Early-bails if dialog support is unavailable and `--all` flag is false.
  - Supports `--all` (`allFlag: true`) to skip prompts and automatically prune all candidate packages and empty git sources.

- **Commands**:
  - **`skills prune`**: Updated `PruneCommand` to support `-a`/`--all` flag and delegate to `pruneSkills`.
  - **`skills get`**: Invokes `pruneSkills` with `quietIfNothingToPrune: true` at the completion of `getSkills` so unused skills and zero-skill git sources are cleaned up automatically after getting skills.

- **Tests & Docs**:
  - Added unit tests for package and local/global git repo pruning in `test/commands/prune_command_test.dart` using declarative `test_descriptor` `.validate()` assertions.
  - Updated `README.md` to document the new `prune` source cleanup behavior and highlight `--git <repo>` options.
  - Updated `CHANGELOG.md`.

### DIFF
--- a/pkgs/skills/lib/src/commands/get_skills.dart
+++ b/pkgs/skills/lib/src/commands/get_skills.dart
@@ -13,6 +13,7 @@ import 'package:skills/src/commands/skills_command.dart';
 import 'package:skills/src/core/advisory_checker.dart';
 import 'package:skills/src/core/git_runner.dart';
 import 'package:skills/src/core/package_resolver.dart';
+import 'package:skills/src/core/pruner.dart';
 import 'package:skills/src/core/pub_runner.dart';
 import 'package:skills/src/core/git_scanner.dart';
 import 'package:skills/src/core/git_sync.dart';
@@ -257,6 +258,14 @@ Future<bool> getSkills({
 
   await globalConfig.save(globalConfigFile);
   await manifest.save(manifestFile(rootPath));
+
+  await pruneSkills(
+    workspace: workspace,
+    logger: logger,
+    dialogSupport: dialogSupport,
+    targetAgents: agents,
+    quietIfNothingToPrune: true,
+  );
 
   return true;
 }
--- a/pkgs/skills/lib/src/commands/get_skills.dart
+++ b/pkgs/skills/lib/src/commands/get_skills.dart
@@ -265,6 +265,7 @@ Future<bool> getSkills({
     dialogSupport: dialogSupport,
     targetAgents: agents,
     quietIfNothingToPrune: true,
+    allFlag: allFlag,
   );
 
   return true;
--- a/pkgs/skills/lib/src/commands/get_skills.dart
+++ b/pkgs/skills/lib/src/commands/get_skills.dart
@@ -148,7 +148,7 @@ Future<bool> getSkills({
 
   final skills = [...dartSkills, ...gitData.gitSkills];
 
-  Map<Agent, Set<String>>? selectedSkillNamesByIde;
+  Map<Agent, Set<String>>? selectedSkillNamesByAgent;
 
   // Log about any unrecognized skills
   if (skillNames.isNotEmpty) {
@@ -216,9 +216,9 @@ Future<bool> getSkills({
   // Use `skillNames` or the `--all` flag as the selection if provided,
   // otherwise prompt for which skills to install or log the available skills.
   if (skillNames.isNotEmpty) {
-    selectedSkillNamesByIde = {};
+    selectedSkillNamesByAgent = {};
     for (final agent in agents) {
-      selectedSkillNamesByIde[agent] = skillNames;
+      selectedSkillNamesByAgent[agent] = skillNames;
     }
   } else if (!allFlag) {
     final promptResult = await _promptForSkillsToInstall(
@@ -234,16 +234,16 @@ Future<bool> getSkills({
     if (!promptResult.continueInstall) {
       return false;
     }
-    selectedSkillNamesByIde = promptResult.selectedSkillNamesByIde;
+    selectedSkillNamesByAgent = promptResult.selectedSkillNamesByAgent;
   }
 
   final installer = SkillInstaller(dialogSupport);
   for (final agent in agents) {
-    final result = await installer.installSkillsForIde(
+    final result = await installer.installSkillsForAgent(
       agent: agent,
       rootPath: rootPath,
       skills: skills,
-      selectedSkills: selectedSkillNamesByIde?[agent],
+      selectedSkills: selectedSkillNamesByAgent?[agent],
       previousManifest: manifest,
       globalConfig: globalConfig,
       sourceUris: sourceUris, // this still acts as source filter
@@ -668,7 +668,7 @@ Future<_PromptResult> _promptForSkillsToInstall({
   required DialogSupport? dialogSupport,
   required Logger logger,
 }) async {
-  final selectedSkillNamesByIde = <Agent, Set<String>>{
+  final selectedSkillNamesByAgent = <Agent, Set<String>>{
     for (final agent in agents) agent: {},
   };
   var hasAnyChangesToPrint = false;
@@ -752,19 +752,19 @@ Future<_PromptResult> _promptForSkillsToInstall({
         for (final index in selectedIndices) {
           final opt = dialogOptions[index];
           for (final adapter in opt.adapters) {
-            selectedSkillNamesByIde[adapter.agent]!.add(opt.skill.skillName);
+            selectedSkillNamesByAgent[adapter.agent]!.add(opt.skill.skillName);
           }
         }
       } else {
         logger.info('Installation aborted by user.');
-        return (continueInstall: false, selectedSkillNamesByIde: null);
+        return (continueInstall: false, selectedSkillNamesByAgent: null);
       }
     }
   }
 
   if (!hasAnyChangesToPrint) {
     logger.info('All skills are up to date.');
-    return (continueInstall: false, selectedSkillNamesByIde: null);
+    return (continueInstall: false, selectedSkillNamesByAgent: null);
   }
 
   if (dialogSupport == null) {
@@ -772,12 +772,12 @@ Future<_PromptResult> _promptForSkillsToInstall({
       'Rerun with `--skill <name>`, or `--all` to '
       'install, update, or remove the given skills.',
     );
-    return (continueInstall: false, selectedSkillNamesByIde: null);
+    return (continueInstall: false, selectedSkillNamesByAgent: null);
   }
 
   return (
     continueInstall: true,
-    selectedSkillNamesByIde: selectedSkillNamesByIde,
+    selectedSkillNamesByAgent: selectedSkillNamesByAgent,
   );
 }
 
@@ -836,7 +836,7 @@ typedef _SkillStatesResult = ({
 
 typedef _PromptResult = ({
   bool continueInstall,
-  Map<Agent, Set<String>>? selectedSkillNamesByIde,
+  Map<Agent, Set<String>>? selectedSkillNamesByAgent,
 });
 
 typedef _DialogOption = ({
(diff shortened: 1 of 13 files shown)


pkgs/skills/lib/src/commands/get_skills.dart
```diff
@@ -258,6 +259,14 @@ Future<bool> getSkills({
   await globalConfig.save(globalConfigFile);
   await manifest.save(manifestFile(rootPath));
 
+  await pruneSkills(
+    workspace: workspace,
+    logger: logger,
+    dialogSupport: dialogSupport,
+    targetAgents: agents,
+    quietIfNothingToPrune: true,
+  );
```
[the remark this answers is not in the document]
jakemac53: This is intended, `skills get --all` only means to install all skills, we shouldn't auto prune without dialogs.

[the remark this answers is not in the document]
jakemac53: Actually, on second thought I think presenting a dialog about pruning in this case is probably not desirable, so we should either auto-prune or skip the dialogs.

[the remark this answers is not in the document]
jakemac53: I went with the auto-prune behavior here, but leaving this as unresolved in case you think I should just skip the pruning step, @kenzieschmoll

pkgs/skills/lib/src/core/pruner.dart
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    for (final uri in localGitToPrune) {
+      manifest = manifest.withoutSourceUri(agent.cliName, uri);
+      totalGitSourcesRemoved++;
+      logger.info('  [${agent.cliName}] Removed empty local git source "$uri"');
+    }
+  }
+
+  await manifest.save(File(SkillManifest.pathIn(rootPath)));
```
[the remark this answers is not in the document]
jakemac53: This needs to stay in case they abort the global cleanup step, we want to save these changes still

pkgs/skills/lib/src/commands/get_skills.dart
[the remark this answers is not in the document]
kenzieschmoll: Should we give the user a choice whenever we are going to prune skills from their machine? Maybe they can answer to configure their preference (yes do this once, yes and do this all the time).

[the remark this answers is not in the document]
jakemac53: I don't want to get too complicated with settings, we might end up needing to create a UI for toggling settings etc then and it can scope creep badly, so I prefer to keep this simpler.

The way this works right now is it will prompt you, unless you pass `--all`, and then it will just do whatever we think is best basically (which means deleting skills from packages you no longer depend on).

pkgs/skills/lib/src/core/pruner.dart
```diff
@@ -0,0 +1,212 @@
...
+    final unreferencedPackages = sourceEntries.keys
+        .where(
+          (uri) =>
+              uri.startsWith('package:') &&
+              !referencedNames.contains(uri.substring('package:'.length)),
+        )
+        .toSet();
+
+    final emptyLocalGitSources = sourceEntries.entries
+        .where((e) => !e.key.startsWith('package:') && e.value.skills.isEmpty)
```
kenzieschmoll: nit: put 'package:' in a named const _packagePrefix so it can be reused

pkgs/skills/lib/src/core/pruner.dart
```diff
@@ -0,0 +1,212 @@
...
+      }
+      selectedItems = selectedIndices.map((i) => candidates[i]).toSet();
+    } else {
+      selectedItems = candidates.toSet();
+    }
+
+    final pkgsToPrune = selectedItems
+        .where((item) => item.startsWith('package:'))
+        .toSet();
+    final localGitToPrune = selectedItems
```
kenzieschmoll: nit: we are iterating over `selectedItems` twice instead of building these two sets in the same iteration of `selectedItems`

pkgs/skills/lib/src/core/pruner.dart
```diff
@@ -0,0 +1,212 @@
...
+        .toList();
+  }
+
+  final installer = SkillInstaller(dialogSupport);
+  var totalRemoved = 0;
+  final prunedPackages = <String>{};
+  var totalGitSourcesRemoved = 0;
+
+  for (final agent in agentsToProcess) {
+    final sourceEntries = manifest.sourceUrisForAgent(agent.cliName);
```
kenzieschmoll: nit: pull this code into a helper _pruneSkillsForAgent so that this code is more readable

pkgs/skills/lib/src/core/pruner.dart
```diff
@@ -0,0 +1,212 @@
...
+
+    final pkgsToPrune = selectedItems
+        .where((item) => item.startsWith('package:'))
+        .toSet();
+    final localGitToPrune = selectedItems
+        .where((item) => !item.startsWith('package:'))
+        .toSet();
+
+    if (pkgsToPrune.isNotEmpty) {
+      final result = await installer.removeSkillsForIde(
```
kenzieschmoll: nit: should this be `removeSkillsForAgent` instead of `removeSkillsForIde` to keep terminology consistent?

kenzieschmoll (commented):
### Review Comments

#### 1. **Redundant `manifest.save` Call Before Cleanup**
* **File:** `pkgs/skills/lib/src/core/pruner.dart`
* **Line:** L184
* **Severity:** `[NIT]` / `[CLEANUP]`
* **Details:** Line 137 already saves `manifest` after processing local package and git sources. Line 184 calls `await manifest.save(File(SkillManifest.pathIn(rootPath)));` a second time right before checking `if (manifest.isEmpty) await SkillManifest.cleanup(rootPath);`. Since `manifest` is not modified during the global git repo check phase (L140–L182), line 184 is redundant and can be safely removed.

---

#### 2. **Logging Condition Bug when only Git Sources are Pruned**
* **File:** `pkgs/skills/lib/src/core/pruner.dart`
* **Lines:** L189–L198
* **Severity:** `[MUST-FIX]` / `[BUG]`
* **Details:** If `totalRemoved == 0` (no package skills were removed) and `totalGitSourcesRemoved > 0` (one or more empty local or global git sources were removed):
  - Line 189 `if (totalRemoved == 0 && totalGitSourcesRemoved == 0)` evaluates to `false`.
  - Line 193 `else if (totalRemoved > 0)` evaluates to `false`.

  As a result, no summary message is logged when pruning only git sources.
* **Suggested Fix:**
  ```dart
  if (totalRemoved == 0 && totalGitSourcesRemoved == 0) {
    if (!quietIfNothingToPrune) {
      logger.info('No skills to prune.');
    }
  } else {
    final summaryParts = <String>[];
    if (totalRemoved > 0) {
      summaryParts.add(
        'Pruned $totalRemoved skill(s) from ${prunedPackages.length} package(s).',
      );
    }
    if (totalGitSourcesRemoved > 0) {
      summaryParts.add(
        'Removed $totalGitSourcesRemoved empty git source(s).',
      );
    }
    logger.info(summaryParts.join(' '));
  }
  ```

---

#### 3. **Missing Unit Test for Global Git Source Pruning**
* **File:** `pkgs/skills/test/commands/prune_command_test.dart`
* **Lines:** L427–L522
* **Severity:** `[SUGGESTION]`
* **Details:** The new tests added in `prune_command_test.dart` (L427–L522) cover local package pruning and local git source pruning, but there is no unit test covering the pruning of unreferenced global git sources from `GlobalConfig`. Adding a unit test for global git repo pruning will ensure full test coverage for global configuration cleanup.

pkgs/skills/lib/src/core/pruner.dart
```diff
@@ -0,0 +1,212 @@
...
+        for (final repo in reposToRemove) {
+          logger.info('Removed global git source "${repo.cloneUrl}".');
+        }
+        globalConfig = globalConfig.withoutGitRepos(reposToRemove);
+        await globalConfig.save(globalConfigFile);
+      }
+    }
+  }
+
+  await manifest.save(File(SkillManifest.pathIn(rootPath)));
```
kenzieschmoll: low

[NIT] Line 137 already saves `manifest` after processing local package and git sources (`await manifest.save(File(SkillManifest.pathIn(rootPath)));`). Since `manifest` is not modified during the global git repo check phase (L140–L182), this second `manifest.save` call on line 184 is redundant and can be safely removed.

pkgs/skills/lib/src/core/pruner.dart
```diff
@@ -0,0 +1,212 @@
...
+  if (manifest.isEmpty) {
+    await SkillManifest.cleanup(rootPath);
+  }
+
+  if (totalRemoved == 0 && totalGitSourcesRemoved == 0) {
+    if (!quietIfNothingToPrune) {
+      logger.info('No skills to prune.');
+    }
+  } else if (totalRemoved > 0) {
```
kenzieschmoll: high

[MUST-FIX] If `totalRemoved == 0` (no package skills were removed) and `totalGitSourcesRemoved > 0` (one or more empty local or global git sources were removed):
- Line 189 `if (totalRemoved == 0 && totalGitSourcesRemoved == 0)` evaluates to `false`.
- Line 193 `else if (totalRemoved > 0)` evaluates to `false`.

As a result, no summary message is logged when pruning only git sources.

```suggestion
  if (totalRemoved == 0 && totalGitSourcesRemoved == 0) {
    if (!quietIfNothingToPrune) {
      logger.info('No skills to prune.');
    }
  } else {
    final summaryParts = <String>[];
    if (totalRemoved > 0) {
      summaryParts.add(
        'Pruned $totalRemoved skill(s) from ${prunedPackages.length} package(s).',
      );
    }
    if (totalGitSourcesRemoved > 0) {
      summaryParts.add(
        'Removed $totalGitSourcesRemoved empty git source(s).',
      );
    }
    logger.info(summaryParts.join(' '));
  }
```

pkgs/skills/test/commands/prune_command_test.dart
```diff
@@ -334,5 +423,101 @@ void main() {
         containsAll(['package:pkg_a', 'package:unref_pkg']),
       );
     });
+
+    test(
```
kenzieschmoll, line 429: medium

[SUGGESTION] The new tests added in `prune_command_test.dart` (L427–L522) cover local package pruning and local git source pruning, but there is no unit test covering the pruning of unreferenced global git sources from `GlobalConfig`. Adding a unit test for global git repo pruning will ensure full test coverage for global configuration cleanup.

kenzieschmoll (approved):

jakemac53:
https://github.com/dart-lang/ai/pull/576 will resolve the test timeout, it is related to the latest flutter beta and not this PR

PR #572, opened 2026-07-30
add the streamable HTTP POST handler from protocol 2026-07-28

Yusufihsangorgel:
Next 2026-07-28 slice tracked in #162, the POST handler laid out in the [finalized-spec diff report](https://github.com/dart-lang/ai/issues/162#issuecomment-5108248966). Rebased onto main now that #571 has landed, so nothing here predates that.

Adds `package:dart_mcp/streamable_http.dart` with `handleStreamableHttpRequest`, the server side of the [Streamable HTTP transport](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http) from the finalized revision. Each POST carries one JSON-RPC request or notification which is validated against the required headers and `_meta` envelope, then dispatched to a fresh server instance via `handleRequestScopedMessage`. Responses are JSON only. Also adds `ProtocolVersion.v2026_07_28`; `ProtocolVersion.latestSupported` still points at 2025-11-25, the newest version the legacy `initialize` handshake negotiates, since a transport for the request-scoped protocol carries its own set of supported versions.

Two mapping decisions worth calling out:

- Accept and Content-Type failures are answered with 400 Bad Request rather than 406/415: the 2026-07-28 schema requires 400 on every `HeaderMismatch` body ("For HTTP, the response status code MUST be `400 Bad Request`"), and the revision names no other status for content negotiation.
- A notification POST is acknowledged with 202 before the header checks run; the spec's note on notification POSTs states "header requirements for notification POSTs are not defined by this revision".

`example/streamable_http_server.dart` serves one tool over the transport and prints a `curl` command for it.

Does not add SSE response streams, the legacy session routes, or an HTTP client; those land as separate slices in that order.

## Tests

- `dart test -p chrome,vm -c dart2wasm,dart2js,kernel,exe` passes: 1028 tests (293 in each VM configuration, 221 in each Chrome configuration; the new suite is `@TestOn('vm')` since it drives `dart:io` sockets).
- `dart analyze --fatal-infos`: no issues.
- `cd mcp_examples && dart pub get && dart analyze --fatal-infos`: no issues.

### DIFF
--- a/pkgs/dart_mcp/lib/src/api/api.dart
+++ b/pkgs/dart_mcp/lib/src/api/api.dart
@@ -30,7 +30,8 @@ enum ProtocolVersion {
   v2024_11_05('2024-11-05'),
   v2025_03_26('2025-03-26'),
   v2025_06_18('2025-06-18'),
-  v2025_11_25('2025-11-25');
+  v2025_11_25('2025-11-25'),
+  v2026_07_28('2026-07-28');
 
   const ProtocolVersion(this.versionString);
 
@@ -43,6 +44,11 @@ enum ProtocolVersion {
   static const oldestSupported = ProtocolVersion.v2024_11_05;
 
   /// The most recent version supported by the current API.
+  ///
+  /// This is the newest version the legacy `initialize` handshake negotiates.
+  /// A transport for a request-scoped protocol revision, such as
+  /// `handleStreamableHttpRequest` in `package:dart_mcp/streamable_http.dart`,
+  /// carries its own set of supported versions.
   static const latestSupported = ProtocolVersion.v2025_11_25;
 
   /// The version string used over the wire to identify this version.
--- a/pkgs/dart_mcp/lib/streamable_http.dart
+++ b/pkgs/dart_mcp/lib/streamable_http.dart
@@ -22,32 +22,6 @@ import 'server.dart';
 import 'src/utils/constants.dart';
 import 'src/utils/json_rpc_2_object.dart';
 
-// Header field names are case insensitive, so these are used both to look
-// headers up and to name them in error messages, in the casing the
-// specification writes them in.
-const _protocolVersionHeader = 'MCP-Protocol-Version';
-const _mcpMethodHeader = 'Mcp-Method';
-const _mcpNameHeader = 'Mcp-Name';
-
-/// The media type of the SSE response streams this protocol revision allows a
-/// server to answer with.
-const _eventStreamMimeType = 'text/event-stream';
-
-/// The protocol versions this handler implements.
-///
-/// The legacy handshake negotiates [ProtocolVersion.latestSupported] instead;
-/// the request-scoped protocol this transport speaks was introduced later, so
-/// the two sets are deliberately separate.
-const _supportedVersions = {ProtocolVersion.v2026_07_28};
-
-/// The methods whose `name` or `uri` parameter is mirrored in the `Mcp-Name`
-/// header, mapping each method to the parameter that carries it.
-const _mcpNameParams = {
-  CallToolRequest.methodName: Keys.name,
-  GetPromptRequest.methodName: Keys.name,
-  ReadResourceRequest.methodName: Keys.uri,
-};
-
 /// Handles one Streamable HTTP POST [request] by validating its headers and
 /// `_meta` envelope, dispatching the decoded message to a fresh server
 /// created by [serverFactory] via [handleRequestScopedMessage], and writing
@@ -85,7 +59,11 @@ const _mcpNameParams = {
 /// error a request handler throws reaches the client with whatever payload
 /// `package:json_rpc_2` attached to it, including a Dart stack trace for
 /// errors which are not an [RpcException]. Handlers which must not disclose
-/// server internals to a remote client throw [RpcException]s instead.
+/// server internals to a remote client throw [RpcException]s instead. The
+/// status such a body gets follows its error code: an internal or server
+/// error is a 500, so a handler which fails is visible to intermediaries
+/// which never read the body, and a code the specification has not mapped to
+/// a status keeps 200.
 ///
 /// If [serverFactory] or [MCPServer.initialize] throws, the request is
 /// answered with 500 and an internal-error body, and the returned future then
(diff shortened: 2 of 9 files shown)


pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -0,0 +1,523 @@
...
+  int status,
+  RpcException exception,
+  Object? origin,
+) async {
+  response
+    ..statusCode = status
+    ..headers.contentType = ContentType.json
+    ..write(jsonEncode(exception.serialize(origin)));
+  await response.close();
+}
```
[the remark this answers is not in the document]
Yusufihsangorgel: `serialize` does inject it. In json_rpc_2 4.1.0 (`RpcException.serialize` in `lib/src/exception.dart`), the request is copied into `error.data` whenever the exception's `data` is `null` or is a `Map` without a `request` key. Every `_reject` site here lands in one of those two branches: 21 of the 22 calls pass no `data`, and the one that does, the unsupported-version payload, is a map without that key. From a live server, a scalar body comes back as

```json
{"jsonrpc":"2.0","error":{"code":-32600,"message":"A message must be a JSON object","data":{"request":42}},"id":null}
```

so the suggested injection would never run. What the sentence did get wrong is the shape: the echo is the decoded message only once decoding succeeded; before that it is the raw body text (a body which was not valid JSON) or `null` (a body never read into text, as on the `Content-Type` reject path). The doc now says exactly that, and a new `rejection bodies` test group pins both `serialize` branches and all three shapes.

pkgs/dart_mcp/lib/src/api/api.dart
```diff
@@ -43,6 +44,11 @@ enum ProtocolVersion {
   static const oldestSupported = ProtocolVersion.v2024_11_05;
 
   /// The most recent version supported by the current API.
+  ///
+  /// This is the newest version the legacy `initialize` handshake negotiates.
+  /// A transport for a request-scoped protocol revision, such as
+  /// `handleStreamableHttpRequest` in `package:dart_mcp/streamable_http.dart`,
+  /// carries its own set of supported versions.
   static const latestSupported = ProtocolVersion.v2025_11_25;
```
jakemac53, line 52: Since we are making some other breaking changes, maybe we should move this elsewhere or rename it?

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -0,0 +1,549 @@
...
+import 'dart:io';
+
+import 'package:json_rpc_2/error_code.dart' as error_code;
+import 'package:json_rpc_2/json_rpc_2.dart';
+
+import 'server.dart';
+import 'src/utils/constants.dart';
+import 'src/utils/json_rpc_2_object.dart';
+
+// Header field names are case insensitive, so these are used both to look
```
jakemac53: nit: move these private members to the bottom of the file

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -0,0 +1,549 @@
...
+  if (error is! Map<String, Object?>) return HttpStatus.ok;
+  return switch (error[Keys.code]) {
+    error_code.PARSE_ERROR ||
+    error_code.INVALID_REQUEST ||
+    error_code.INVALID_PARAMS ||
+    McpErrorCodes.headerMismatch ||
+    McpErrorCodes.missingRequiredClientCapability ||
+    McpErrorCodes.unsupportedProtocolVersion => HttpStatus.badRequest,
+    error_code.METHOD_NOT_FOUND => HttpStatus.notFound,
+    _ => HttpStatus.ok,
```
jakemac53, line 466: Is OK really the best default here? It feels like it should be some sort of error response.

jakemac53 (commented):
Can you also add an example using this transport?

I still need to take a look at the tests as well, but I reviewed the actual changes.

pkgs/dart_mcp/lib/src/api/api.dart
Yusufihsangorgel: Fair, this PR is what broke the name — `v2026_07_28` is new here and I covered it with a doc comment instead.

It is ~45 references across `dart_mcp`, `dart_mcp_server` and `mcp_examples`, including the `update_mcp_spec` skill, which tells the next person to point it at the new version. And `oldestSupported` and `isSupported` describe that same handshake era, so renaming one of the three leaves the others wrong.

Rename all three to say handshake, or take the range out of `ProtocolVersion` and let the legacy handshake own what it negotiates? I lean to the second. Either way I would rather send it separately than put a 45-site rename on top of this diff.

pkgs/dart_mcp/lib/streamable_http.dart
Yusufihsangorgel: Moved them to the bottom.

pkgs/dart_mcp/lib/streamable_http.dart
Yusufihsangorgel: Changed, `INTERNAL_ERROR` and `SERVER_ERROR` answer 500 now. The inconsistency was ours: a throwing `serverFactory` already gets 500, so a crashed handler getting 200 meant the same class of failure answered differently depending on where it happened, and the 200 hid it from anything that does not read the body.

I left the fallback at 200 rather than making it an error status, and pinned it with a test now that the crash case no longer covers it. The spec reserves `-32020` to `-32099` for itself, and a revision allocating a code from there names the status along with it, so guessing one here would get those wrong.

Yusufihsangorgel:
Added `example/streamable_http_server.dart`. It serves one tool and prints a `curl` command with the headers and `_meta` envelope filled in, since there is no HTTP client here to pair it with yet. I ran it against that command to check the envelope is right.

Yusufihsangorgel:
The example has changed since I described it above, so here is where it stands. It answers on one path and 404s anything else, rejects a request that carries any `Origin` with 403, and binds port 0 like the rest of the package. `greet` reports progress when the request carries a token, and the printed command carries one, so `onNotification` runs on the documented path instead of sitting bound to nothing. Comments name the two jobs the handler leaves to a host that the example does not do, authentication and a body size limit. The README names `server/discover`, which servers are required to implement and this package does not.

jakemac53 (approved):

pkgs/dart_mcp/lib/src/api/api.dart
Yusufihsangorgel: Correcting the number I gave you here: 45 was `latestSupported` alone. With `oldestSupported` and `isSupported` it is 62 sites on `main` today, in `dart_mcp` (54), `dart_mcp_server` (2), `mcp_examples` (4), and the `update_mcp_spec` skill (2). #580 removes one of them.

PR #574, opened 2026-07-31
Include install location in skill install summary message (#552)

### LINKED ISSUE #552: Skills install messages should indicate output location
After installing skills, we list the names but not the locations. We should also indicate where they were installed to.

cc @jwren

jakemac53:
- Add relative install directory path in parentheses to the agent installation summary log message in `skills get` / `skills add`.
- Add unit test verifying install location in log messages.
- Update CHANGELOG.md.

Fixes #552

### DIFF
--- a/pkgs/skills/CHANGELOG.md
+++ b/pkgs/skills/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.0.0-wip
 
+- Include install location in log messages when installing skills.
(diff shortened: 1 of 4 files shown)


jakemac53:
Actually looks like https://github.com/dart-lang/ai/pull/573 is also tackling this, I will just review that first and only land this if that one is not sufficient for some reason.

PR #579, opened 2026-08-04
let servers set the caching hints on cacheable results

Yusufihsangorgel:
Next 2026-07-28 core-protocol slice tracked in #162, following #570: the five cacheable result factories now take the `ttlMs` and `cacheScope` caching hints. #580 is the other half, what the dispatcher fills in when a handler did not.

#570 added `CacheableResult` to read the two off a result and closed with "server-side emission of the hints (factory parameters) are follow-up slices". This is that slice, for the results the caching rules name (https://modelcontextprotocol.io/specification/2026-07-28/server/utilities/caching).

- Both are optional and are left out of the result when they are not passed. Writing a default `ttlMs` here would be indistinguishable from a real answer, since an absent one and a written `0` both read as `0`. Writing a default `cacheScope` would be worse: synthesizing `public` for a server that never declared a scope is the one guess that can leak a private response into a shared cache, which is why #570 has the getter read `null` instead.
- A negative `ttlMs` is an assertion failure. The spec requires `ttlMs >= 0`, and the reader already folds a negative value to `0`, so a server sending one would be telling clients something other than what it meant. Same shape as `Annotations`' assert on `priority`.
- `ReadResourceResult` takes the two without `nextCursor`, matching the schema: it is cacheable but not paginated.
- They are two parameters rather than one hints object because the factory is where these fields are defined, so a third hint would land here anyway, and because every other factory in the package takes its fields this way.

That page also names `server/discover`, which this package does not serve yet, so five factories change here.

## Tests

- Each factory writes both hints when given them, writes just the one it is given, and leaves both keys out when given neither. The assertions read the result map rather than the getters, since an absent `ttlMs` and a written `ttlMs: 0` both read as `0`. That `0` is covered explicitly: it is a meaningful value in the spec, not a sentinel.
- `dart test -p chrome,vm -c dart2wasm,dart2js,kernel,exe` passes (1040 tests across the four configurations).
- `dart analyze --fatal-infos` is clean here and in `mcp_examples` resolved against this branch.

One note on what is not tested: the `ttlMs` assert has no test. Asserts are compiled out under `-c exe`, so a test which expects one to throw fails there, and the package tests none of its 37 other asserts either.

Part of #162.

### DIFF
--- a/pkgs/dart_mcp/test/api/api_test.dart
+++ b/pkgs/dart_mcp/test/api/api_test.dart
@@ -179,9 +179,10 @@ void main() {
   });
 
   group('cacheable result factory', () {
-    // The results the caching rules name, minus `server/discover`, which this
-    // package does not serve yet. These assert on the map rather than the
-    // getters: an absent `ttlMs` and a written `ttlMs: 0` both read as `0`.
+    // Returns an instance of every cacheable result type with the given
+    // settings, except `server/discover`, which this package does not serve
+    // yet. The tests assert on the map rather than the getters: an absent
+    // `ttlMs` and a written `ttlMs: 0` both read as `0`.
     List<Map<String, Object?>> cacheable({
       int? ttlMs,
       CacheScope? cacheScope,
--- a/pkgs/dart_mcp/test/api/api_test.dart
+++ b/pkgs/dart_mcp/test/api/api_test.dart
@@ -177,4 +177,65 @@ void main() {
       expect(listTools.cacheScope, CacheScope.public);
     });
   });
+
+  group('cacheable result factory', () {
+    // The results the caching rules name, minus `server/discover`, which this
+    // package does not serve yet. These assert on the map rather than the
+    // getters: an absent `ttlMs` and a written `ttlMs: 0` both read as `0`.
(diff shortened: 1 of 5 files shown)


pkgs/dart_mcp/test/api/api_test.dart
```diff
@@ -177,4 +177,65 @@ void main() {
       expect(listTools.cacheScope, CacheScope.public);
     });
   });
+
+  group('cacheable result factory', () {
+    // The results the caching rules name, minus `server/discover`, which this
```
jakemac53: nit: wording on this comment is a bit awkward, maybe something like `Returns an instance of every cacheable result type  with the given settings, except server/discover which this package does not serve yet.`.

jakemac53 (commented):
Just the one tiny nit

pkgs/dart_mcp/test/api/api_test.dart
Yusufihsangorgel: Done. I kept the note about asserting on the map, since that is what explains the return type.

jakemac53 (approved):

PR #580, opened 2026-08-04
send the result fields the 2026-07-28 revision requires

Yusufihsangorgel:
Next 2026-07-28 core-protocol slice tracked in #162, following #565 and #570: the request-scoped dispatcher now sends `resultType`, `ttlMs`, and `cacheScope`. #579 is the other half, how a handler supplies the two hints itself.

Both landed read-only, and both closed with a note that server-side emission was a later slice.

- Every result on 2026-07-28 carries a `resultType`. The schema types the field as required for a server on this revision (https://modelcontextprotocol.io/specification/2026-07-28/schema#result-resulttype), and the dispatcher was leaving it off, so every result the Streamable HTTP handler returned was missing it.
- The list and read results also carry the caching hints those rules require (https://modelcontextprotocol.io/specification/2026-07-28/server/utilities/caching). `resources/read` is the one of them the schema also answers with an interim result (https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/mrtr), so it is the only one whose result type excuses the hints.
- All of it is gated on the negotiated revision, including the reserved `_meta` key from #528: it is 2026-07-28 vocabulary too. `handleRequestScopedMessage` takes the version in its `MCPServerInitialization` and is exported for transports this package does not own, so it cannot assume the revision the way `handleStreamableHttpRequest` can.
- A field the handler set is kept when the schema allows it; an assert catches one it does not, and the dartdoc on `handleRequestScopedMessage` lists what replaces it in a build with asserts disabled.

This sits in the server info stamping from #528 rather than in the result factories, because the fields belong to a protocol revision and the same server also serves 2025-11-25 over stdio.

`'complete'` moved to `ResultTypes` in `constants.dart`, so the value the reader folds to and the value the dispatcher writes are one definition.

## Tests

- A result type is recorded, and one the handler chose is kept. The hints are recorded on `tools/list` and on `resources/read`, and not on `tools/call`. Hints the handler chose are kept, one hint given and one left out is completed, and fields left `null` are replaced. An interim `resources/read` gets no hints while an interim-typed `tools/list` still gets them. An earlier revision gets none of it. All asserted on the result map, since no existing test pins the set of keys on a result. A value the schema rejects fails an assert instead.
- The dispatch tests now negotiate 2026-07-28, since the request-scoped lifecycle arrived with it; the earlier revision has its own test.
- `dart test -p chrome,vm -c dart2wasm,dart2js,kernel,exe` passes (1094 tests across the four configurations).
- `dart analyze --fatal-infos` is clean here and in `mcp_examples` resolved against this branch.

Part of #162.

### DIFF
--- a/pkgs/dart_mcp/lib/src/server/request_scoped.dart
+++ b/pkgs/dart_mcp/lib/src/server/request_scoped.dart
@@ -33,12 +33,14 @@ typedef MCPServerFactory =
 /// `io.modelcontextprotocol/serverInfo` result metadata key, carries a
 /// `resultType`, and, for the requests the caching rules name, carries `ttlMs`
 /// and `cacheScope` unless it is an interim `resources/read` result, which is
-/// not cacheable. A field the handler set is kept when the schema allows it:
-/// a `resultType` left `null` becomes `complete`, a `ttlMs` which is `null`
-/// or below zero becomes `0`, and a `cacheScope` which is `null` or outside
-/// `public` and `private` becomes `private`. The dispatcher cannot know when
-/// an answer goes stale, and `public` is the one guess which could put a
-/// private answer in a cache shared across authorization contexts. None of
+/// not cacheable. A field the handler left out is filled in: a `resultType`
+/// left `null` becomes `complete`, a `ttlMs` which is `null` becomes `0`, and
+/// a `cacheScope` which is `null` becomes `private`. The dispatcher cannot
+/// know when an answer goes stale, and `public` is the one guess which could
+/// put a private answer in a cache shared across authorization contexts.
+/// A `ttlMs` below zero or a `cacheScope` outside `public` and `private` is a
+/// bug in the server: it asserts, and in a build with asserts disabled it is
+/// replaced the same way a field left `null` is. None of
 /// this reaches a result on an earlier revision, and every error response is
 /// returned unchanged. If the server closes before responding to a request,
 /// an internal-error response is returned instead. The server may still be
--- a/pkgs/dart_mcp/lib/src/server/request_scoped.dart
+++ b/pkgs/dart_mcp/lib/src/server/request_scoped.dart
@@ -163,16 +163,11 @@ Future<Map<String, Object?>?> handleRequestScopedMessage(
               );
             }
         }
-      } catch (error, stackTrace) {
-        // The server sent a frame we could not process. Never let it wedge or
-        // escape the exchange: a request gets an internal error, anything
-        // else surfaces as an uncaught error. A failed assert is a bug in the
-        // server rather than a frame we could not read, so it surfaces even
-        // when a response still goes back; otherwise the message it carries
-        // would be lost behind the internal error.
-        if (error is AssertionError || !isRequest || response.isCompleted) {
-          Zone.current.handleUncaughtError(error, stackTrace);
-        }
+      } catch (_) {
+        // The server sent a frame we could not process. Answer the request
+        // before rethrowing so a caller waiting on it is not left hanging;
+        // the error itself still reaches the zone, which is where a bug in
+        // the server belongs.
         if (isRequest && !response.isCompleted) {
           response.complete(
             _errorResponse(
@@ -181,6 +176,7 @@ Future<Map<String, Object?>?> handleRequestScopedMessage(
             ),
           );
         }
+        rethrow;
       }
     },
     onDone: () {
--- a/pkgs/dart_mcp/lib/src/server/request_scoped.dart
+++ b/pkgs/dart_mcp/lib/src/server/request_scoped.dart
@@ -164,7 +164,13 @@ Future<Map<String, Object?>?> handleRequestScopedMessage(
       } catch (error, stackTrace) {
         // The server sent a frame we could not process. Never let it wedge or
         // escape the exchange: a request gets an internal error, anything
-        // else surfaces as an uncaught error.
+        // else surfaces as an uncaught error. A failed assert is a bug in the
+        // server rather than a frame we could not read, so it surfaces even
+        // when a response still goes back; otherwise the message it carries
+        // would be lost behind the internal error.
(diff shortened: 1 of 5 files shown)


pkgs/dart_mcp/lib/src/utils/constants.dart
```diff
@@ -2,6 +2,15 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// Constants for the values the specification names for [Keys.resultType].
+///
+/// A server may send any other string, so these are not an enum.
+extension ResultTypes on Never {
+  /// A result which is not waiting on anything, and the value the
+  /// specification tells a client to assume when the field is absent.
+  static const complete = 'complete';
```
jakemac53, line 13: We should add one here for "input_required" as well, and link to https://modelcontextprotocol.io/specification/latest/schema#resulttype

pkgs/dart_mcp/test/server/request_scoped_test.dart
```diff
@@ -59,6 +59,149 @@ void main() {
...
+      final withScope = await _dispatchShapedList(
+        (result) => {...result, Keys.cacheScope: 'public'},
+      );
+      expect(_result(withScope), containsPair(Keys.ttlMs, 0));
+      expect(_result(withScope), containsPair(Keys.cacheScope, 'public'));
+    });
+
+    test('replaces a caching hint the schema does not allow', () async {
+      final response = await _dispatchShapedList(
+        (result) => {...result, Keys.ttlMs: -1, Keys.cacheScope: 'shared'},
```
jakemac53: I would generally prefer this to be an error I think, even if other packages choose to ignore such errors, that is really just hiding a bug in the server implementation imo.

If there is a strong argument for masking these errors instead, then I would opt for adding an assert, so that in tests it will fail, even if in prod the errors are glossed over.

pkgs/dart_mcp/lib/src/utils/constants.dart
Yusufihsangorgel: Done, added `inputRequired` and the spec link in the doc comment. The tests use the constants now instead of the raw strings this PR was adding.

pkgs/dart_mcp/test/server/request_scoped_test.dart
Yusufihsangorgel: Went with the assert. The factories already carry one since #579, but a handler can hand back any map, so the dispatcher is where a result built without them gets caught; these tests build that map.

The argument for masking is already in the helper: non-map metadata is left alone instead of failing the exchange, and dropping a whole `tools/list` response over a bad `ttlMs` seemed inconsistent with that.

A bare assert did not surface anything though. It fires inside the try that guards frame handling, and the catch turned it into the generic `-32603` answer, dropping the assert message. The catch now hands `AssertionError` to the zone before answering; every other path through it is unchanged.

The tests pin both modes: the `AssertionError` with asserts on, and the old replacement in an executable build, where asserts are stripped. The exe configuration failed without that branch.

pkgs/dart_mcp/lib/src/server/request_scoped.dart
```diff
@@ -164,16 +166,20 @@ Future<Map<String, Object?>?> handleRequestScopedMessage(
       } catch (error, stackTrace) {
         // The server sent a frame we could not process. Never let it wedge or
         // escape the exchange: a request gets an internal error, anything
-        // else surfaces as an uncaught error.
+        // else surfaces as an uncaught error. A failed assert is a bug in the
+        // server rather than a frame we could not read, so it surfaces even
+        // when a response still goes back; otherwise the message it carries
+        // would be lost behind the internal error.
+        if (error is AssertionError || !isRequest || response.isCompleted) {
+          Zone.current.handleUncaughtError(error, stackTrace);
```
jakemac53: Can this just be `rethrow`?

pkgs/dart_mcp/test/server/request_scoped_test.dart
```diff
@@ -772,6 +802,16 @@ MCPServerInitialization _initialization({
 Map<String, Object?> _result(Map<String, Object?>? response) =>
     response![Keys.result] as Map<String, Object?>;
 
+/// Whether this configuration runs with asserts enabled.
+///
+/// A compiled executable has them stripped, so a test which pins an assert has
+/// to pin the behavior a release build has instead.
+bool get _assertsEnabled {
```
jakemac53: Asserts should always be enabled in tests, can you drop this and the conditions above?

pkgs/dart_mcp/lib/src/server/request_scoped.dart
Yusufihsangorgel: Almost. With `rethrow` in place of that call, the two schema rejection tests hang, nothing completes the response future. Once the error reply is out, the rest is just `rethrow` and the condition goes away. An error that used to stop at the `-32603` reply now reaches the zone too.

pkgs/dart_mcp/test/server/request_scoped_test.dart
Yusufihsangorgel: Dropped both conditions and the getter, the tests just assume asserts now. That assumption breaks on the `-c exe` run in the workflow, asserts are compiled out there, so the two affected tests carry `testOn: '!exe'` instead of the branching. Without the `testOn` that run fails. Part of my answer on the assert thread no longer holds: the replacement of a rejected value is untested on the executable build, the fill-in tests still run everywhere.

jakemac53 (approved):

pkgs/dart_mcp/test/server/request_scoped_test.dart
jakemac53: Oh interesting, that is a bug in the test package then, filed https://github.com/dart-lang/test/issues/2717.

Skipping for now on exe is fine.

PR #593, opened 2026-08-11
reject the methods the 2026-07-28 revision removed

Yusufihsangorgel:
The gate in `handleStreamableHttpRequest` only covered the two legacy lifecycle methods, but the [2026-07-28 revision](https://modelcontextprotocol.io/specification/2026-07-28/changelog) removed more than those. Every server answered `ping` with `200`, since it comes from `MCPBase`. A server picks up `logging/setLevel` and the two `resources` subscription methods from the mixins, and `notifications/roots/list_changed` from the roots capability.

`ProtocolVersion` now carries what each revision added and removed. The transport asks it, so a handler registered later gets the same `404`.

Part of #162.


pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -550,6 +552,24 @@ const _eventStreamMimeType = 'text/event-stream';
...
 const _supportedVersions = {ProtocolVersion.v2026_07_28};
 
+/// The methods this package declares which the 2026-07-28 schema no longer
+/// defines.
+///
+/// The package still declares them for the revisions the legacy handshake
+/// negotiates, while this transport implements only the request-scoped
+/// protocol, where they are unknown. The list covers every one, so a handler
+/// registered for any of them later still cannot answer it here.
+const _methodsTheRevisionRemoved = {
```
jakemac53: I think it would make sense to move the supported methods more directly next to the protocol definitions. Probably just a static method on ProtocolVersion like `static bool methodIsValid(ProtocolVersion version, String method)`, or maybe just an extension method on ProtocolVersion to make it more ergonomic: `bool methodIsValid(String method)`.

We could then attach `addedMethods` and `removedMethods` to each protocol revision enum entry, and the function would walk backwards from the given `version` in a loop. If we see `method` in removedMethods for a version, we immediately return false, if we see it in `addedMethods`, we return true, and if we make it through the loop without hitting either condition, we return false.

Yusufihsangorgel: Moved it onto `ProtocolVersion`, with `addedMethods` and `removedMethods` per revision from the schemas, and `methodIsValid` walking back as you described.

Asking it directly also turns away methods no revision defines, which took out our own `test/crash` handler, so the transport asks whether some revision defined the method and this one removed it.

pkgs/dart_mcp/test/streamable_http_test.dart
```diff
@@ -624,6 +629,70 @@ void main() {
...
+          json: body(
+            rootsListChangedNotification,
+            capabilities: {
+              'roots': {'listChanged': true},
+            },
+          ),
+        );
+        expect(status, 404);
+        expect(errorCode(text), error_code.METHOD_NOT_FOUND);
+      });
```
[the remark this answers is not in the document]
Yusufihsangorgel: I tried it. Dropping `'notifications/elicitation/complete'` from `removedMethods` leaves the new test passing, because nothing on the server registers a handler for it. `ElicitationRequestSupport` sends the notification, the handler is on the client. The same edit on `'notifications/roots/list_changed'` does fail its test.

pkgs/dart_mcp/test/streamable_http_test.dart
```diff
@@ -22,15 +22,20 @@ const version = '2026-07-28';
...
 const getPrompt = GetPromptRequest.methodName;
 const readResource = ReadResourceRequest.methodName;
 const initialize = InitializeRequest.methodName;
 const initializedNotification = InitializedNotification.methodName;
 const progressNotification = ProgressNotification.methodName;
+const ping = PingRequest.methodName;
+const setLevel = SetLevelRequest.methodName;
+const subscribe = SubscribeRequest.methodName;
+const unsubscribe = UnsubscribeRequest.methodName;
+const rootsListChangedNotification = RootsListChangedNotification.methodName;
```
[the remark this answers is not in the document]
Yusufihsangorgel: Only needed for the test in the other thread, answered there.

pkgs/dart_mcp/lib/src/api/api.dart
```diff
@@ -27,13 +27,96 @@ part 'tools.dart';
 
 /// Enum of the known protocol versions.
 enum ProtocolVersion {
-  v2024_11_05('2024-11-05'),
+  v2024_11_05(
+    '2024-11-05',
+    addedMethods: {
+      'completion/complete',
```
jakemac53: We should be able to use the static constants on the request extension types instead of strings here, just to ensure things stay in sync (they should all have a `methodName` static const field).

Yusufihsangorgel: Done for 29 of the 34. The remaining five are the `tasks` ones, which have no extension types in the package at all, so I just left them as strings.

jakemac53 (approved):

PR #595, opened 2026-08-12
send notifications/message only when a 2026-07-28 request asks

Yusufihsangorgel:

On 2026-07-28 the log level travels in each request's `_meta`, as [`io.modelcontextprotocol/logLevel`](https://modelcontextprotocol.io/specification/2026-07-28/server/utilities/logging). A server must send no `notifications/message` for a request that omits the key, and `LoggingSupport` was still sending them at its warning default.

I put the level on `MCPServerInitialization`, filled by the streamable HTTP handler from the envelope key beside the reserved ones it already reads. An unrecognized value comes back as `-32602`. That leaves the warning default just for 2025-11-25 and older, where `logging/setLevel` still sets it.

### DIFF
--- a/pkgs/dart_mcp/lib/src/server/logging_support.dart
+++ b/pkgs/dart_mcp/lib/src/server/logging_support.dart
@@ -34,6 +34,9 @@ base mixin LoggingSupport on MCPServer {
   /// Sends a [LoggingMessageNotification] to the client, if the [loggingLevel]
   /// is <= [level].
   ///
+  /// Sends nothing on 2026-07-28 unless the exchange was given a
+  /// [MCPServerInitialization.logLevel].
+  ///
   /// The [data] must either be some json serializable object, or a function
   /// which takes no arguments and returns some json serializable object.
   ///
--- a/pkgs/dart_mcp/lib/src/server/logging_support.dart
+++ b/pkgs/dart_mcp/lib/src/server/logging_support.dart
@@ -11,11 +11,22 @@ base mixin LoggingSupport on MCPServer {
   /// The current logging level, defaults to [LoggingLevel.warning].
   LoggingLevel loggingLevel = LoggingLevel.warning;
 
+  /// Whether this exchange asked for log messages, see
+  /// [MCPServerInitialization.logLevel].
+  bool _logsRequested = true;
+
   @override
   FutureOr<ServerCapabilities> initialize(
     MCPServerInitialization initialization,
   ) async {
     registerRequestHandler(SetLevelRequest.methodName, handleSetLevel);
+
+    final requested = initialization.logLevel;
+    if (requested != null) {
+      loggingLevel = requested;
+    } else if (initialization.protocolVersion >= ProtocolVersion.v2026_07_28) {
+      _logsRequested = false;
+    }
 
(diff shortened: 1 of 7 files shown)


pkgs/dart_mcp/lib/src/server/logging_support.dart
```diff
@@ -11,18 +11,32 @@ base mixin LoggingSupport on MCPServer {
   /// The current logging level, defaults to [LoggingLevel.warning].
   LoggingLevel loggingLevel = LoggingLevel.warning;
 
+  /// Whether this exchange asked for log messages, see
+  /// [MCPServerInitialization.logLevel].
+  bool _logsRequested = true;
```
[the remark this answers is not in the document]
Yusufihsangorgel: I kept the flag out of `handleSetLevel` on purpose. It only goes false on 2026-07-28, which replaced `logging/setLevel` with the per-request `_meta` key and forbids `notifications/message` for requests omitting it. Setting it there would violate that MUST NOT.

pkgs/dart_mcp/lib/src/server/logging_support.dart
```diff
@@ -11,18 +11,32 @@ base mixin LoggingSupport on MCPServer {
...
   FutureOr<ServerCapabilities> initialize(
     MCPServerInitialization initialization,
   ) async {
     registerRequestHandler(SetLevelRequest.methodName, handleSetLevel);
 
+    final requested = initialization.logLevel;
+    if (requested != null) {
+      loggingLevel = requested;
+    } else if (initialization.protocolVersion >= ProtocolVersion.v2026_07_28) {
+      _logsRequested = false;
```
jakemac53: We could consider making `loggingLevel` nullable and use that as a signal to not log instead of adding an extra boolean? We could then retain the default level of warning but only for older protocol versions?

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -326,6 +327,23 @@ Future<void> handleStreamableHttpRequest(
...
+  if (rawLogLevel != null) {
+    logLevel = LoggingLevel.values.firstWhereOrNull(
+      (level) => level.name == rawLogLevel,
+    );
+    if (logLevel == null) {
+      return _reject(
+        response,
+        HttpStatus.badRequest,
+        RpcException.invalidParams(
+          'The envelope ${Keys.logLevelMeta} must be a logging level',
```
jakemac53: Consider listing the valid logging levels and the one we received here.

pkgs/dart_mcp/lib/src/server/logging_support.dart
Yusufihsangorgel: Switched it over. The extra boolean is gone and `loggingLevel` is `LoggingLevel?` now, with the warning default only applied below 2026-07-28. 2025-11-25 clients keep what they had.

pkgs/dart_mcp/lib/streamable_http.dart
Yusufihsangorgel: It lists both now, the level we got and the valid set from `LoggingLevel.values`.

jakemac53 (approved):

PR #596, opened 2026-08-12
serve server/discover from MCPServer

Yusufihsangorgel:

The discover types landed without a dispatcher, so `server/discover` answered -32601 on a revision that requires it. I put the handler on the server, registered only from 2026-07-28 on, since answering is itself the claim to that lifecycle and the [stdio probe](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/stdio#backward-compatibility) keys on it. TypeScript and Python gate it to the same revision.

Capabilities are a public final field now and `initialize` no longer returns them. That is breaking. Mixins edit the object in place. The capabilities constructor was also skipping `completions`, just a missing line, fixed here.

### DIFF
--- a/pkgs/dart_mcp/lib/src/server/server.dart
+++ b/pkgs/dart_mcp/lib/src/server/server.dart
@@ -207,17 +207,43 @@ abstract base class MCPServer extends MCPBase {
 
   /// The capabilities [discover] advertises.
   ///
-  /// The revisions this answers for dropped `resources/subscribe`, and they
-  /// carry list changes on `subscriptions/listen` streams, which this package
-  /// does not serve yet, so the bits standing for those stay off. The
-  /// `initializeLegacy` response keeps them, since they are honest on the
-  /// revisions that handshake negotiates.
+  /// On these revisions a client only hears a list change or a resource update
+  /// over a `subscriptions/listen` stream it opened with the matching
+  /// `promptsListChanged`, `toolsListChanged`, `resourcesListChanged` or
+  /// `resourceSubscriptions` filter, and this package does not serve that
+  /// request yet. So the four bits standing for those notifications come off
+  /// here: `listChanged` on each of the three, and `subscribe` on
+  /// [ServerCapabilities.resources]. Every other key the server registered is
+  /// passed through, and `initializeLegacy` keeps all of them, since the
+  /// revisions that handshake negotiates still serve `resources/subscribe` and
+  /// send the list changes without a filter.
   ServerCapabilities get _advertisedCapabilities => ServerCapabilities.fromMap({
     ...capabilities as Map<String, Object?>,
-    if (capabilities.prompts != null) Keys.prompts: Prompts(),
-    if (capabilities.resources != null) Keys.resources: Resources(),
-    if (capabilities.tools != null) Keys.tools: Tools(),
+    if (capabilities.prompts case final prompts?)
+      Keys.prompts: Prompts.fromMap(
+        _without(prompts as Map<String, Object?>, const [Keys.listChanged]),
+      ),
+    if (capabilities.resources case final resources?)
+      Keys.resources: Resources.fromMap(
+        _without(resources as Map<String, Object?>, const [
+          Keys.listChanged,
+          Keys.subscribe,
+        ]),
+      ),
+    if (capabilities.tools case final tools?)
+      Keys.tools: Tools.fromMap(
+        _without(tools as Map<String, Object?>, const [Keys.listChanged]),
+      ),
   });
+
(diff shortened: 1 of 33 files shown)


pkgs/dart_mcp/lib/src/server/server.dart
```diff
@@ -88,6 +92,15 @@ abstract base class MCPServer extends MCPBase {
...
   Implementation? clientInfo;
 
+  /// The capabilities [initialize] registered, which [discover] advertises.
+  ///
+  /// This is the object [initialize] builds, and mixins and subclasses edit it
+  /// on their way back up rather than replacing it, so it ends up carrying
+  /// every feature they registered.
+  ///
+  /// Only assigned after [initialize] has been called.
+  late ServerCapabilities _capabilities;
```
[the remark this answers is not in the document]
Yusufihsangorgel: Capturing the return value makes this worse. A replacing override leaves the mixin handlers registered, so `discover` would advertise less than the server handles. `_capabilities` doesn't stay empty, it's the instance the mixins edited. Nothing reads it in the legacy lifecycle. I'm keeping it as is.

pkgs/dart_mcp/lib/src/server/server.dart
```diff
@@ -144,9 +157,47 @@ abstract base class MCPServer extends MCPBase {
...
     }
-    return ServerCapabilities();
+    // Registering this handler is itself a statement about the lifecycle, so
+    // only a server on a request-scoped revision does it. A client probing
+    // under the stdio backward compatibility rules would read an answer here
+    // as "this connection is modern".
+    if (protocolVersion >= _firstRequestScopedVersion) {
+      registerRequestHandler(DiscoverRequest.methodName, discover);
+    }
+    return _capabilities = ServerCapabilities();
```
jakemac53: I don't love this pattern - should we instead make capabilities a public getter and drop the return value?

That is a breaking change, but we are already doing a breaking release. Then, people cannot bash over the capabilities, they can only modify the existing object.

It is still slightly awkward, but less error prone I think.

pkgs/dart_mcp/lib/src/server/server.dart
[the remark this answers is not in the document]
Yusufihsangorgel: Dropped the return value entirely. An override cannot hand back a different instance now, and the legacy response carries the same object the mixins edited.

pkgs/dart_mcp/lib/src/server/server.dart
Yusufihsangorgel: I went with a plain final field instead of a getter. Nothing can assign over it either way, and it saves the backing field. The return value on `initialize` is dropped, so the mixins just edit it in place.

pkgs/dart_mcp/lib/src/server/server.dart
```diff
@@ -194,10 +201,24 @@ abstract base class MCPServer extends MCPBase {
...
+  /// The capabilities [discover] advertises.
+  ///
+  /// The revisions this answers for dropped `resources/subscribe`, and they
+  /// carry list changes on `subscriptions/listen` streams, which this package
+  /// does not serve yet, so the bits standing for those stay off. The
+  /// `initializeLegacy` response keeps them, since they are honest on the
+  /// revisions that handshake negotiates.
+  ServerCapabilities get _advertisedCapabilities => ServerCapabilities.fromMap({
+    ...capabilities as Map<String, Object?>,
+    if (capabilities.prompts != null) Keys.prompts: Prompts(),
```
jakemac53: I still see these capabilities listed in the latest spec https://modelcontextprotocol.io/specification/2026-07-28/server/resources ?

jakemac53: Also the way this is structured it will strip everything and not just `subscribe`, I would change it to explicitly just remove what is intended to be removed, for better future compatitibility

Yusufihsangorgel: Yeah, both bits are still there. What the revision dropped is the `resources/subscribe` method, and using that as the reason to turn the bits off was wrong.

The real reason is just delivery: those notifications are gated on a `subscriptions/listen` stream with the matching filter set, and we don't serve it yet. I rewrote the doc to say that.

Yusufihsangorgel: I had it rebuilding each capability from empty, so anything else the server set on them went out with the bits, including keys the package doesn't know about.

It copies and drops named keys now. The intended set is four, not just `subscribe`: `listChanged` on prompts, tools and resources, and `subscribe` on resources. There's a test that an unknown field survives.

jakemac53 (approved):

PR #604, opened 2026-08-15
carry the subscription id in its own metadata type

Yusufihsangorgel:

The id of the request that opened a listen stream travels with everything the server sends back on that stream, under the reserved metadata key, see https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/subscriptions.

The listen result took that id as a parameter and copied the caller's metadata into a new map to place it. The acknowledgement had no parameter for the id. I gave both messages the `MetaWithSubscriptionId` itself, with a getter that hands the object back typed, so a caller just builds it once.


pkgs/dart_mcp/lib/src/api/subscriptions.dart
```diff
@@ -121,15 +127,16 @@ extension type SubscriptionsListenResult.fromMap(Map<String, Object?> _value)
...
 extension type SubscriptionsAcknowledgedNotification.fromMap(
   Map<String, Object?> _value
-) implements Notification {
+) implements Notification, WithSubscriptionId {
   static const methodName = 'notifications/subscriptions/acknowledged';
 
   factory SubscriptionsAcknowledgedNotification({
     required SubscriptionFilter notifications,
+    required RequestId subscriptionId,
     Meta? meta,
```
jakemac53: Should we add a `MetaWithSubscriptionId` type? That way, we wouldn't have to copy the provided `Meta` map into a new one.

Yusufihsangorgel: Added it, following `MetaWithProgressToken`. I left `WithSubscriptionId` off it though. That getter reads a message's `_meta`, and the metadata object is the thing sitting under that key, not something holding one. Both messages read through the new type now.

jakemac53 (approved):

Yusufihsangorgel:
Closing this. The type already landed in https://github.com/dart-lang/ai/pull/636. Sorry for not saying so there.

PR #614, opened 2026-08-21
check the elicitation mode at both ends

Yusufihsangorgel:

The server side only asked whether the client had `elicitation`, but a client declares which modes it takes and the [spec](https://modelcontextprotocol.io/specification/2026-07-28/client/elicitation#capabilities) says a server must not send one it did not. A url-only client saw form requests and declined them as if the user had.

I split that guard into `supportsFormElicitation` and `supportsUrlElicitation`, and both ends now answer an unknown mode with invalid params. Auto-handling a `-32042` only runs when the request inside it names url. Both are breaking. The resolved getter throws on an unknown one, so the guards read a new `rawMode`.

### DIFF
--- a/pkgs/dart_mcp/lib/src/server/elicitation_request_support.dart
+++ b/pkgs/dart_mcp/lib/src/server/elicitation_request_support.dart
@@ -9,6 +9,29 @@ base mixin ElicitationRequestSupport on LoggingSupport {
   /// is based on the client capabilities.
   bool get supportsElicitation => clientCapabilities.elicitation != null;
 
+  /// Whether or not the connected client supports [ElicitationMode.form]
+  /// requests.
+  ///
+  /// Only safe to call after calling [initialize] on `super` since this
+  /// is based on the client capabilities.
+  ///
+  /// A client which declared `elicitation` but named neither mode counts as
+  /// form support, the rule the 2025-11-25 revision added alongside the mode
+  /// split.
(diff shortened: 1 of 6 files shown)


pkgs/dart_mcp/lib/src/server/elicitation_request_support.dart
```diff
@@ -9,6 +9,29 @@ base mixin ElicitationRequestSupport on LoggingSupport {
...
+  /// Only safe to call after calling [initialize] on `super` since this
+  /// is based on the client capabilities.
+  ///
+  /// A client which declared `elicitation` but named neither mode counts as
+  /// form support, the rule the 2025-11-25 revision added alongside the mode
+  /// split.
+  bool get supportsFormElicitation {
+    final elicitation = clientCapabilities.elicitation;
+    if (elicitation == null) return false;
+    return elicitation.form != null || elicitation.url == null;
```
jakemac53: If another kind of elicitation is added, this might falsely return `true`.

Should we instead change it to just check if it's an empty object?

Yusufihsangorgel: You are right, and the spec puts it the same way: 2025-11-25 says an empty capabilities object is form only. I switched it to that check and added a case for an unknown mode, which the old expression let through.

jakemac53 (approved):

PR #615, opened 2026-08-21
Add detail to instructions for creating a skill

ericwindmill:
I'm 0th customer-ing package:skills. This PR adds a bit more detail to the skill template and clarifies some documentation.

---


pkgs/skills/lib/src/commands/create_command.dart
```diff
@@ -121,3 +117,30 @@ String? _isValidSkillName(String name) {
...
+
+String _createSkillTemplate(String fullSkillName, String description) =>
+    '''---
+name: $fullSkillName
+description: ${jsonEncode(description)}
+---
+
+# $fullSkillName
+
+Brief overview of what this skill does and what it enables the agent to do.
```
jakemac53: I know many skills start this way but I don't think it is correct to put this here, it's just a waste of tokens. If you reach the body, the agent has already invoked the skill, so telling it when to invoke it or what it does is kind of pointless (it already chose to invoke it, for a specific reason presumably).

Basically, this should all be in the description, and the skill should just contain the instructions.

ericwindmill: Wow, thats a great point, and is so obvious now that I've heard it.

jakemac53 (commented):
Just a quick changelog entry please and then this is good to land 🚀

ericwindmill:
@jakemac53  - Not sure if the way I updated the changelog / pubspec is correct. LMK if it should just be a line item on the current version's change list

jakemac53 (approved):

PR #617, opened 2026-08-24
Ignore uninstalled and missing skills in skills list

### LINKED ISSUE #607: Skills list lies about installed skills
Given the following skills_config.json `skills list` gives a list of all the "installed skills" this list is wrong because the list explicitly has "isInstallled" false. Either skills list should say that it is a list of skills that can be installed or it should filter based on the skills that are explicitly installed. 

Second: when deleting the installed skills from disk `skills list` still lists them as installed. 

```
{
  "name": "web-artifacts-builder",
  "installedAt": "2026-08-18T17:36:41.186637Z",
  "isInstalled": false,
  "path": "skills/web-artifacts-builder"
}
```
[skills_config.json](https://github.com/user-attachments/files/31192130/skills_config.json)

jakemac53:
Fixes #607

- Filter out skills with `isInstalled: false` in `ListCommand`
- Filter out skills that no longer exist on disk in `ListCommand`
- Filter out empty packages and agents with no installed skills in `ListCommand`
- Filter out `isInstalled: false` in `SkillManifest.allSkills` and `allSkillsForAgent`
- Add changelog entry for package:skills
- Add regression tests in `list_command_test.dart` and `skill_manifest_test.dart`


pkgs/skills/lib/src/commands/list_command.dart
```diff
@@ -35,32 +37,66 @@ class ListCommand extends SkillsCommand {
       ..writeln('Installed skills:')
       ..writeln();
 
+    var hasInstalledSkills = false;
+
     for (final agentName in manifest.allAgents) {
       final pkgs = manifest.sourceUrisForAgent(agentName);
       if (pkgs.isEmpty) continue;
```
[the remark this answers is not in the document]
jakemac53: That isn't worth doing imo

pkgs/skills/lib/src/commands/list_command.dart
```diff
@@ -35,32 +37,66 @@ class ListCommand extends SkillsCommand {
...
-      } else {
-        header = '  $agentName:';
-      }
-      buffer.writeln(header);
+      final adapter = agentObj != null
+          ? createAgentAdapter(agentObj, rootPath, null)
+          : null;
+
+      final agentBuffer = StringBuffer();
+      var agentHasSkills = false;
```
reidbaker: What is the difference between agentHasSkillsand hasInstalledSkills?

reidbaker: FWIW you might want some object oriented type safety here because I also see pkgHasSkills below.

pkgs/skills/lib/src/commands/list_command.dart
reidbaker, line 26: This method could be easier to read if you broke out the subsections into their own methods with dart docs.

pkgs/skills/test/models/skill_manifest_test.dart
```diff
@@ -221,6 +221,32 @@ void main() {
       expect(multi.allSkills, hasLength(2));
     });
 
+    test(
```
reidbaker, line 224: You probably want to include a test where you make sure isInstalled: True is handled.

reidbaker (approved):

reidbaker (commented):
FWIW I think if you ran the cognitive_complexity package on list_command.dart your agent woudl refactor this class to be more readable.

pkgs/skills/lib/src/commands/list_command.dart
jakemac53: So there are 3 layers here - all the installed skills (across all agents) - the skills installed per agent - and then the skills per agent per package. That is what all 3 of these variables are encapsulating.

We only want to output information for agents/packages that actually have any skills installed, and similarly if zero skills are installed we have a separate message for that.

pkgs/skills/test/models/skill_manifest_test.dart
jakemac53: The base manifest does have one (and that is why the expectations below don't expect an empty list).

pkgs/skills/lib/src/commands/list_command.dart
jakemac53: That could help simplify these booleans as well so it makes more sense, I can try that out.

jakemac53:
> FWIW I think if you ran the cognitive_complexity package on list_command.dart your agent woudl refactor this class to be more readable.

cool I will try that, haven't used it before

PR #629, opened 2026-08-28
decode SSE response bodies on the client

Yusufihsangorgel:

The client side has no SSE reader yet, so I added `sseMessageStream`. A frame that fails to decode arrives as an error while the stream keeps reading, since one bad frame probably should not cost the whole response. A truncated frame never arrives and bad bytes become the replacement character. The HTML specification asks for both.

The decoder sits under `src/utils` because it needs nothing from `dart:io`. The `events` helper in the transport tests reads a whole string. It stays for now. I want this for the client transport, the next slice.


pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -517,6 +541,38 @@ const _listenStreamNotifications = {
...
+    final field = separator < 0 ? line : line.substring(0, separator);
+    var value = separator < 0 ? '' : line.substring(separator + 1);
+    if (value.startsWith(' ')) value = value.substring(1);
+    if (field == 'event') {
+      event = value;
+    } else if (field == Keys.data) {
+      data.add(value);
+    }
+  }
+}
```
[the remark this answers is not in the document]
Yusufihsangorgel: WHATWG has this the other way round. 9.2.6: "Once the end of the file is reached, any pending data must be discarded. (If the file ends in the middle of an event, before the final empty line, the incomplete event is not dispatched.)"

I checked the two reference SDKs before pushing back. The TypeScript client pipes through `EventSourceParserStream` from eventsource-parser. That transform defines no flush handler, so nothing goes out when the writable side closes. The Python client gets there by another route: it flushes the line buffer at the end, but dispatch still waits for the blank line.

Flushing would cost us something too. A dropped connection usually lands mid-payload. A truncated `data: {"jsonrpc":"2.0","id":1,"resu` would then throw a FormatException instead of just ending. The parser cannot tell that from a clean cut without parsing the data first.

The doc comment now spells the drop out and links the interpretation rules. There is a test for it as well.

pkgs/dart_mcp/test/streamable_http_test.dart
```diff
@@ -170,6 +170,211 @@ void main() {
...
+          'data: {"jsonrpc":"2.0","id":1,"result":{}}\n\n',
+        ),
+      );
+      final messages = await sseMessageStream(bytes).toList();
+
+      expect(messages, [
+        {'jsonrpc': '2.0', 'method': 'notifications/message'},
+        {'jsonrpc': '2.0', 'id': 1, 'result': <String, Object?>{}},
+      ]);
+    });
```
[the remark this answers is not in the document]
Yusufihsangorgel: Same reason as the parser thread. A test asserting the event still arrives without the trailing blank line would pin behaviour the specification tells us to drop, and both reference SDKs return nothing for that input. The test I added goes the other way and checks the truncated frame is dropped.

pkgs/dart_mcp/lib/src/utils/sse.dart
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+/// The joined `data` of each `message` event in an SSE response [bytes].
+///
+/// An event with no data has nothing to decode and never reaches the
+/// stream.
+///
+/// `data` and `message` also spell MCP payload keys. Here they name SSE
+/// framing, a separate namespace.
+Stream<String> _sseMessageData(Stream<List<int>> bytes) async* {
```
jakemac53, line 43: Probably worth linking https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream here.

pkgs/dart_mcp/lib/src/utils/sse.dart
```diff
@@ -0,0 +1,69 @@
...
+      final source = data.join('\n');
+      if (source.isNotEmpty && (event.isEmpty || event == 'message')) {
+        yield source;
+      }
+      event = '';
+      data.clear();
+      continue;
+    }
+
+    final separator = line.indexOf(':');
```
jakemac53, line 59: According to https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream, if it starts with a `:` then the line should be skipped, it might be worth more explicitly handling that, even though the current code would end up skipping it anyways.

pkgs/dart_mcp/lib/src/utils/sse.dart
```diff
@@ -0,0 +1,69 @@
...
+    final field = separator < 0 ? line : line.substring(0, separator);
+    var value = separator < 0 ? '' : line.substring(separator + 1);
+    if (value.startsWith(' ')) value = value.substring(1);
+    if (field == 'event') {
+      event = value;
+    } else if (field == 'data') {
```
jakemac53, line 65: I see also that `id` and `retry` are valid field values, do we need to handle those?

jakemac53 (approved):

pkgs/dart_mcp/lib/src/utils/sse.dart
Yusufihsangorgel: That one's on the helper now. I kept the interpretation link upstairs, since the dropped frame and the lenient decode both live in 9.2.6. Down here it's just the grammar that matters.

pkgs/dart_mcp/lib/src/utils/sse.dart
Yusufihsangorgel: Explicit now. I pulled the branch back out and all fifteen tests stayed green. It still bought one test though: a comment that spells a field, `:data: {...}`, would break a rewrite of the split.

pkgs/dart_mcp/lib/src/utils/sse.dart
Yusufihsangorgel: No. Resumable streams via `Last-Event-ID` are gone in this revision, so neither field reaches anything we hand on. The transport already ignores that header though (`streamable_http_test.dart:634` covers it), and I put the reason on the helper.

PR #632, opened 2026-08-28
answer the input requests in an input_required result

Yusufihsangorgel:

The client half of MRTR. On an `input_required` result the connection fills in
what the server asked for and sends the call again. A request it cannot build
fails the call instead of answering the rest.

The handshake stops at 2025-11-25, so only a transport setting the version
itself reaches 2026-07-28. I made `protocolVersion` nullable. Nothing fills in
`serverInfo` there either, and both sampling paths name it.

`ServerConnection` gains `sendRequestWithInputs` for a mixin adding an
extension method. `maxInputRequiredRounds` moves the cap off ten. `MCPBase`
gains `@protected` `sendRequestKeepingProgress` and `closeProgress`. A retry
holds one progress token across its rounds.

### DIFF
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -125,11 +125,12 @@ base class ServerConnection extends MCPBase {
   /// protocol version.
   ProtocolVersion? protocolVersion;
 
-  /// The [Implementation] returned from the [initialize] request.
+  /// The [Implementation] the server described itself with, or `null` until
+  /// one settles it.
   ///
-  /// Only non-null after [initialize] has successfully completed. Sampling
-  /// hands it to the handler, so set it too wherever you set
-  /// [protocolVersion].
+  /// [initialize] fills this in. Assign it directly if your transport settles
+  /// the version outside the handshake, since sampling hands it to the
+  /// handler.
   Implementation? serverInfo;
 
   /// The [ServerCapabilities] returned from the [initialize] request.
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -230,7 +230,6 @@ base class ServerConnection extends MCPBase {
        _elicitationUrlSupport = elicitationUrlSupport,
        _samplingSupport = samplingSupport,
        _rootsSupport = rootsSupport {
-    elicitationFormSupport ??= elicitationSupport;
     if (rootsSupport != null) {
       registerRequestHandler(
         ListRootsRequest.methodName,
@@ -246,7 +245,7 @@ base class ServerConnection extends MCPBase {
       );
     }
 
-    if (elicitationFormSupport != null || elicitationUrlSupport != null) {
+    if (_elicitationFormSupport != null || elicitationUrlSupport != null) {
       registerRequestHandler(ElicitRequest.methodName, (ElicitRequest request) {
         final raw = request.rawMode;
         if (raw != null && !ElicitationMode.values.any((m) => m.name == raw)) {
@@ -257,12 +256,13 @@ base class ServerConnection extends MCPBase {
         }
         switch (request.mode) {
           case ElicitationMode.form:
-            if (elicitationFormSupport == null) {
+            final formSupport = _elicitationFormSupport;
+            if (formSupport == null) {
               throw RpcException.invalidParams(
                 'This client did not declare the elicitation.form capability',
               );
             }
-            return elicitationFormSupport.handleElicitation(request, this);
+            return formSupport.handleElicitation(request, this);
           case ElicitationMode.url:
             if (elicitationUrlSupport == null) {
               throw RpcException.invalidParams(
@@ -433,7 +433,8 @@ base class ServerConnection extends MCPBase {
     WithInputResponses request,
   ) async {
     final originalRequest = request as Map<String, Object?>;
-    var result = (await sendRequest<T>(methodName, request)) as Result;
+    var result =
+        (await sendRequestKeepingProgress<T>(methodName, request)) as Result;
     for (
       var round = 0;
       result.resultType == ResultTypes.inputRequired;
@@ -456,6 +457,8 @@ base class ServerConnection extends MCPBase {
           '`${Keys.inputRequests}` or `${Keys.requestState}`.',
         );
       }
+      // TODO: Pace a leg carrying only request state, which nothing else slows
+      // down, the way the TypeScript and Python SDKs do.
       if (inputRequests != null) {
         final handlers = [
           for (final entry in inputRequests.entries)
@@ -475,7 +478,9 @@ base class ServerConnection extends MCPBase {
                 if (requestState != null) Keys.requestState: requestState,
               }
               as WithInputResponses;
-      result = (await sendRequest<T>(methodName, retryRequest)) as Result;
+      result =
+          (await sendRequestKeepingProgress<T>(methodName, retryRequest))
+              as Result;
     }
     return result as T;
   }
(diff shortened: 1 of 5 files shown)


pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -228,7 +245,7 @@ base class ServerConnection extends MCPBase {
       );
     }
 
-    if (elicitationFormSupport != null || elicitationUrlSupport != null) {
+    if (_elicitationFormSupport != null || elicitationUrlSupport != null) {
```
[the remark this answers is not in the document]
Yusufihsangorgel: They do read the same here, since the initializer copies the parameter across untouched. The form one beside it has to use the field because that one merges the deprecated `elicitationSupport`. I've matched the url side to it anyway, since the two sit in the same condition.

pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -389,7 +424,133 @@ base class ServerConnection extends MCPBase {
 
   /// Gets the requested [Prompt] from the server.
   Future<GetPromptResult> getPrompt(GetPromptRequest request) =>
-      sendRequest(GetPromptRequest.methodName, request);
+      _sendRequestWithInputs(GetPromptRequest.methodName, request);
+
+  Future<T> _sendRequestWithInputs<T extends Result>(
```
jakemac53, line 429: Should we make this public? I think for MCP extensions and such, it allows people to bring their own mixins to implement those, and use this method if those extensions allow the input required flows.

pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -134,9 +141,35 @@ base class ServerConnection extends MCPBase {
...
+  final ElicitationFormSupport? _elicitationFormSupport;
+
+  /// The [SamplingSupport] for this connection, if any.
+  final SamplingSupport? _samplingSupport;
+
+  /// The [RootsSupport] for this connection, if any.
+  final RootsSupport? _rootsSupport;
+
+  /// Maximum number of automatic retries for an `input_required` result.
+  static const _maxInputRequiredRounds = 10;
```
jakemac53, line 154: Is this a part of the specification? If not, let's make it configurable (10 sounds like a fine default though). Otherwise, lets link where in the spec this number comes from.

pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -389,7 +424,133 @@ base class ServerConnection extends MCPBase {
...
+      final retryRequest =
+          <String, Object?>{
+                for (final entry in originalRequest.entries)
+                  if (entry.key != Keys.inputResponses &&
+                      entry.key != Keys.requestState)
+                    entry.key: entry.value,
+                if (responses.isNotEmpty) Keys.inputResponses: responses,
+                if (requestState != null) Keys.requestState: requestState,
+              }
+              as WithInputResponses;
```
jakemac53, line 497: Is this cast valid even if `responses` is empty?

pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -389,7 +424,133 @@ base class ServerConnection extends MCPBase {
...
+      result =
+          (await sendRequestKeepingProgress<T>(methodName, retryRequest))
+              as Result;
+    }
+    return result as T;
+  }
+
+  Future<Result> Function() _inputRequestHandler(InputRequest inputRequest) {
```
jakemac53, line 505: I think a short doc comment would be useful on this method, briefly explaining the general input requests mechanism, and linking to the spec for more detail.

pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -421,6 +582,42 @@ base class ServerConnection extends MCPBase {
       sendRequest(CompleteRequest.methodName, request);
 }
 
+Map<String, Object?>? _inputRequestParams(
```
jakemac53, line 585: This feels like it should be a part of the extension type? Or at least the part that reads directly from the map should instead be using a nullable getter on the extension type.

pkgs/dart_mcp/test/client/input_required_test.dart
```diff
@@ -0,0 +1,944 @@
...
+
+    expect(harness.connection.protocolVersion, ProtocolVersion.v2025_11_25);
+    expect((result.content.single as TextContent).text, 'legacy result');
+    expect(harness.requests.map((request) => request['method']), [
+      InitializeRequest.methodName,
+      CallToolRequest.methodName,
+    ]);
+  });
+
+  test('does not interpret an input-required result before a version is '
```
jakemac53, line 52: Is this behavior specified? It feels to me like it should be an error to try and do anything prior the version being set, so maybe this should be hitting a more generic error?

Basically I am just curious what sort of valid interaction might involve not having an agreed upon version prior to any other interaction?

pkgs/dart_mcp/lib/src/client/client.dart
Yusufihsangorgel: Made it public as `sendRequestWithInputs`. It holds the progress token across
the rounds itself. A mixin calling it does not need the two `@protected`
helpers underneath.

pkgs/dart_mcp/lib/src/client/client.dart
Yusufihsangorgel: Not in the spec. On the MRTR page `limit` and `loop` are both zero hits, and
the four `maximum` hits are all in the page's viewport script. It is a settable
`ServerConnection.maxInputRequiredRounds` now, ten by default. I tried it as a
`connectServer` parameter first and it broke the three examples overriding that
method.

pkgs/dart_mcp/lib/src/client/client.dart
Yusufihsangorgel: Yes. `WithInputResponses` erases to `Map<String, Object?>`, so the cast never
looks at the keys, and `inputResponses` reads null when we leave it out. The
three `retries ... with request state` tests already send a round with no
responses.

pkgs/dart_mcp/lib/src/client/client.dart
Yusufihsangorgel: Added one naming the three methods a server can ask for, with the MRTR link.

pkgs/dart_mcp/lib/src/client/client.dart
Yusufihsangorgel: Moved the read onto `InputRequest.params`. It does the object check and names
the method now, and the client keeps a one-line helper for the two methods the
schema requires params on. Putting the getter back to the plain cast turns the
three `rejects non-object params` tests red with a `_TypeError`.

pkgs/dart_mcp/test/client/input_required_test.dart
Yusufihsangorgel: Not specified. 2026-07-28 has no handshake at all — the lifecycle page makes
`server/discover` optional and says an `initialize` request selects legacy
semantics.

Here it is narrower still: `ProtocolVersion.latestSupported` is `v2025_11_25`,
so `initialize` cannot produce a 2026-07-28 connection. Every one of them gets
its version from a transport assigning the field, and null is that same field
before the transport assigns it. It takes the older-revision branch rather than
its own error because both mean the connection has not told us it speaks the
revision.

I renamed the test to `treats an unsettled version like one before 2026-07-28`.

pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -150,8 +150,12 @@ base class ServerConnection extends MCPBase {
   /// The [RootsSupport] for this connection, if any.
   final RootsSupport? _rootsSupport;
 
-  /// Maximum number of automatic retries for an `input_required` result.
-  static const _maxInputRequiredRounds = 10;
+  /// How many `input_required` rounds [sendRequestWithInputs] answers before
+  /// it gives up.
+  ///
+  /// The spec sets no bound on the rounds. This one only guards against a
+  /// server that never stops asking, and a caller can move it.
```
jakemac53: Consider making it nullable, null meaning infinite rounds?

Yusufihsangorgel: It's nullable now, and I kept ten as the default. Null just skips the check, so the loop runs as long as the server keeps asking. A test clears it and finishes on the thirteenth request.

jakemac53 (approved):

PR #635, opened 2026-08-29
cache client responses from server hints

Yusufihsangorgel:

The client reuses complete results for the six 2026-07-28 operations that
carry caching hints, keyed by the parameters and the request metadata. I
wrapped `sendRequestKeepingProgress` so `sendRequestWithInputs` still hits
the cache. Entries drop on change notifications, a stale cursor, and
`shutdown`. The cache belongs to one `ServerConnection`. A `private` result
never leaves the authorization context it came from.

The multi round-trip page never mentions caching. The caching utilities page
(https://modelcontextprotocol.io/specification/2026-07-28/server/utilities/caching)
says an `input_required` result is not cacheable, and a retry carrying
`inputResponses` or `requestState` MUST NOT be cached. I kept both out even
when a `resources/read` payload also sends the hints.

### DIFF
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -376,11 +376,13 @@ base class ServerConnection extends MCPBase {
   /// response cache where the 2026-07-28 caching rules allow it.
   ///
   /// [sendRequest] and [sendRequestWithInputs] both go through here, so a
-  /// `resources/read` that later retries still consults the cache. Only a
-  /// request whose `_meta` names 2026-07-28 is cached. Cached answers drop
-  /// when the matching list-changed or resource-updated notification arrives,
-  /// when a paginated request reports its cursor is no longer valid, and on
-  /// [shutdown].
+  /// `resources/read` that later retries still consults the cache. A request
+  /// is cached when its `_meta` names 2026-07-28, or when it leaves the
+  /// version out and this connection settled on that revision. Cached answers
+  /// drop when the matching list-changed or resource-updated notification
+  /// arrives, when a paginated request reports its cursor is no longer valid,
+  /// and on [shutdown].
+  @protected
   @override
   Future<T> sendRequestKeepingProgress<T extends Result?>(
     String methodName, [
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -124,10 +124,40 @@ base class ServerConnection extends MCPBase {
   @visibleForTesting
   int get cachedResponseCount => _responseCache._entries.length;
 
+  /// The maximum number of responses cached by this connection.
+  ///
+  /// Reducing the value evicts the oldest entries first. Zero disables the
+  /// cache. The default is 512.
+  int get maxCachedResponses => _responseCache.maxEntries;
+  set maxCachedResponses(int value) {
+    RangeError.checkNotNegative(value, 'maxCachedResponses');
+    while (_responseCache._entries.length > value) {
+      _responseCache._entries.remove(_responseCache._entries.keys.first);
+    }
+    _responseCache.maxEntries = value;
+  }
+
+  /// The maximum TTL accepted for responses received by this connection.
+  ///
+  /// Zero disables the cache. The default is 24 hours.
+  Duration get maxCachedResponseTtl => _responseCache.maxTtl;
+  set maxCachedResponseTtl(Duration value) {
+    if (value.isNegative) {
+      throw ArgumentError.value(value, 'maxCachedResponseTtl');
+    }
+    _responseCache.maxTtl = value;
+  }
+
   /// Advances the cache clock by [duration].
   @visibleForTesting
   void elapseCachedResponses(Duration duration) {
     _responseCache._elapsedOffset += duration;
+  }
+
+  /// Stops the real-time part of the cache clock.
+  @visibleForTesting
+  void pauseCachedResponseClock() {
+    _responseCache._clock.stop();
   }
 
   /// The version of the protocol this connection speaks, or `null` until one
@@ -336,13 +366,7 @@ base class ServerConnection extends MCPBase {
     registerNotificationHandler<ResourceUpdatedNotification>(
       ResourceUpdatedNotification.methodName,
       (notification) {
-        // A notification that leaves `uri` out still reaches subscribers.
-        final uri = (notification as Map<String, Object?>)[Keys.uri];
-        if (uri is String) {
-          _responseCache.invalidateResource(uri);
-        } else {
-          _responseCache.invalidateAllResources();
-        }
+        _responseCache.invalidateResource(notification.uri);
         _resourceUpdatedController.sink.add(notification);
       },
     );
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -139,13 +139,16 @@ base class ServerConnection extends MCPBase {
 
   /// The maximum TTL accepted for responses received by this connection.
   ///
-  /// Zero disables the cache. The default is 24 hours.
+  /// Changing the value discards stored responses. Zero disables the cache.
+  /// The default is 24 hours.
   Duration get maxCachedResponseTtl => _responseCache.maxTtl;
   set maxCachedResponseTtl(Duration value) {
     if (value.isNegative) {
       throw ArgumentError.value(value, 'maxCachedResponseTtl');
     }
+    if (_responseCache.maxTtl == value) return;
     _responseCache.maxTtl = value;
+    _responseCache._entries.clear();
   }
 
   /// Advances the cache clock by [duration].
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -389,7 +389,13 @@ base class ServerConnection extends MCPBase {
     final meta = (request as Map<String, Object?>?)?[Keys.meta];
     final version =
         meta is Map<String, Object?> ? meta[Keys.protocolVersionMeta] : null;
-    if (version != ProtocolVersion.v2026_07_28.versionString) {
+    // Only `discover` names the revision in its own metadata. Every other
+    // request leaves it out and the negotiated version is what says whether
+    // the answer can be reused.
+    final onRevision =
+        version == ProtocolVersion.v2026_07_28.versionString ||
+        (version == null && protocolVersion == ProtocolVersion.v2026_07_28);
+    if (!onRevision) {
       return super.sendRequestKeepingProgress<T>(methodName, request);
     }
     return _responseCache.sendRequest<T>(
--- a/pkgs/dart_mcp/lib/src/client/client.dart
+++ b/pkgs/dart_mcp/lib/src/client/client.dart
@@ -6,6 +6,8 @@ import 'dart:async';
 import 'dart:collection';
 
 import 'package:async/async.dart' show StreamExtensions;
+import 'package:collection/collection.dart';
+import 'package:json_rpc_2/error_code.dart' as error_code;
 import 'package:json_rpc_2/json_rpc_2.dart';
 import 'package:meta/meta.dart';
 import 'package:stream_channel/stream_channel.dart';
@@ -16,6 +18,7 @@ import '../shared.dart';
 import '../utils/constants.dart';
 
 part 'elicitation_support.dart';
+part 'response_cache.dart';
 part 'roots_support.dart';
 part 'sampling_support.dart';
 
@@ -115,6 +118,8 @@ base class MCPClient {
 
 /// An active server connection.
 base class ServerConnection extends MCPBase {
+  final _responseCache = _ClientResponseCache();
+
   /// The version of the protocol this connection speaks, or `null` until one
   /// is settled.
   ///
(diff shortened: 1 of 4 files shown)


pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -292,24 +307,44 @@ base class ServerConnection extends MCPBase {
...
+      },
     );
 
-    registerNotificationHandler(
+    registerNotificationHandler<ResourceUpdatedNotification>(
       ResourceUpdatedNotification.methodName,
-      _resourceUpdatedController.sink.add,
+      (notification) {
+        // A notification that leaves `uri` out still reaches subscribers.
+        final uri = (notification as Map<String, Object?>)[Keys.uri];
```
jakemac53: The spec doesn't seem to mark this as optional?

https://modelcontextprotocol.io/specification/draft/schema#resourceupdatednotification

pkgs/dart_mcp/lib/src/client/response_cache.dart
```diff
@@ -0,0 +1,294 @@
...
+    parameter,
+    const DeepCollectionEquality().hash(context),
+  );
+}
+
+final class _CacheEntry {
+  const _CacheEntry(this.result, this.expiresAt);
+
+  final Map<String, Object?> result;
+  final Duration expiresAt;
```
jakemac53, line 38: I would expect a `DateTime` for a field called `expiresAt`, and probably in general I would expect this to just store an already transformed `DateTime`.

Is there a particular reason we want to instead store the duration? If there is, I would comment here and then also talk about what it is compared against and how.

pkgs/dart_mcp/lib/src/client/response_cache.dart
```diff
@@ -0,0 +1,294 @@
...
+/// the per-request context the caller sent. A result is stored only while it
+/// is complete, carries a positive `ttlMs` and carries a scope the schema
+/// allows, and the answers to an [InputRequiredResult] stay out entirely. A
+/// `ttlMs` past a day is clamped to a day. Without that a server could pin an
+/// answer for the life of the process, and a value large enough to overflow a
+/// [Duration] would write an entry that is already stale.
+///
+/// https://modelcontextprotocol.io/specification/2026-07-28/server/utilities/caching
+final class _ClientResponseCache {
+  static const _maxEntries = 512;
```
jakemac53: This should be configurable, as well as the max Ttl.

pkgs/dart_mcp/lib/src/client/response_cache.dart
```diff
@@ -0,0 +1,294 @@
...
+  static const _maxTtl = Duration(hours: 24);
+
+  final _entries = <_CacheKey, _CacheEntry>{};
+  final _pending = <_CacheKey, Object>{};
+
+  /// The resource URIs updated while a `resources/read` was in flight, under
+  /// the token of that read.
+  final _updatedWhilePending = <Object, Set<String>>{};
+  final _clock = Stopwatch()..start();
```
jakemac53, line 73: I am confused why we have a cache-wide clock here - is this only to make things more testable? I would use something like `package:clock` instead for this, so that we can inject a custom clock for testing purposes.

I also don't understand the logic for cache eviction - it seems to compare elapsed with duration, but doesn't take into account when the request was actually received.

pkgs/dart_mcp/lib/src/client/response_cache.dart
```diff
@@ -0,0 +1,294 @@
...
+/// answer for the life of the process, and a value large enough to overflow a
+/// [Duration] would write an entry that is already stale.
+///
+/// https://modelcontextprotocol.io/specification/2026-07-28/server/utilities/caching
+final class _ClientResponseCache {
+  static const _maxEntries = 512;
+  static const _maxTtl = Duration(hours: 24);
+
+  final _entries = <_CacheKey, _CacheEntry>{};
+  final _pending = <_CacheKey, Object>{};
```
jakemac53: I would expect the values here to be futures which complete with the result.

Or possibly just have `_entries = <_CacheKey, FutureOR<_CacheEntry>>{};`

pkgs/dart_mcp/lib/src/client/response_cache.dart
```diff
@@ -0,0 +1,294 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
```
jakemac53, line 1: We could also consider just updating the AsyncCache in package:async (https://pub.dev/documentation/async/latest/async/AsyncCache-class.html) to support a custom cache time per result?

pkgs/dart_mcp/lib/src/client/response_cache.dart
```diff
@@ -0,0 +1,294 @@
...
+  Future<T> _sendAndStore<T extends Result?>(
+    _CacheKey key,
+    Future<T> Function() send,
+  ) async {
+    final entry = _entries[key];
+    if (entry != null) {
+      if (entry.expiresAt.compareTo(_elapsed) > 0) {
+        return _copyMap(entry.result) as T;
+      }
+      _entries.remove(key);
```
jakemac53: We might want a more pro-active cache busting solution than this?

pkgs/dart_mcp/lib/src/client/response_cache.dart
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    final token = Object();
+    _pending[key] = token;
```
jakemac53: I would expect this to be looking for an existing pending request, waiting for that response and using it if it is cacheable. If it isn't cacheable then it would fire its own request.

pkgs/dart_mcp/lib/src/client/response_cache.dart
```diff
@@ -0,0 +1,294 @@
...
+              : Duration(milliseconds: ttlMs);
+      if (!_entries.containsKey(key) && _entries.length >= _maxEntries) {
+        _entries.remove(_entries.keys.first);
+      }
+      // An answer that arrives after one of its own contents changed is
+      // already out of date, and a read only learns which contents it carries
+      // once it gets here.
+      final updated = _updatedWhilePending[token];
+      if (updated != null && _carriesAny(resultMap, updated)) return result;
+      _entries[key] = _CacheEntry(_copyMap(resultMap), receivedAt + ttl);
```
jakemac53: Ah I see now, we are adding the offset into here. This was not clear to me elsewhere. I do think it would be simpler and more clear to just use DateTime objects instead, which would also work nicely with package:clock , but you could also just update the comments elsewhere to make it more clear how this works.

pkgs/dart_mcp/lib/src/client/response_cache.dart
```diff
@@ -0,0 +1,294 @@
...
+      if (entry.expiresAt.compareTo(_elapsed) > 0) {
+        return _copyMap(entry.result) as T;
+      }
+      _entries.remove(key);
+    }
+
+    final token = Object();
+    _pending[key] = token;
+    if (key.methodName == ReadResourceRequest.methodName) {
+      _updatedWhilePending[token] = <String>{};
```
jakemac53: It might be more clear/cheaper to just lazily add this set in, only if we actually have to add to it.

pkgs/dart_mcp/lib/src/client/response_cache.dart
jakemac53: Also if we do keep this, probably tests will want to actually have the ability to pause it.

jakemac53 (commented):
On further review I think using a stopwatch is actually the better alternative, likely faster and also not affected by wall clock adjustments that may happen. So it is overall simpler but I think just adding some comments explaining the parts that confused me would be helpful.

pkgs/dart_mcp/lib/src/client/client.dart
Yusufihsangorgel: You're right. Both the 2026-07-28 and draft schemas require `params.uri`, so the handler now reads `notification.uri` directly.

pkgs/dart_mcp/lib/src/client/response_cache.dart
Yusufihsangorgel: The field comment now says the `Duration` is monotonic stopwatch time and is compared with that stopwatch, not wall time.

pkgs/dart_mcp/lib/src/client/response_cache.dart
Yusufihsangorgel: Both limits are now configurable on `ServerConnection` through `maxCachedResponses` and `maxCachedResponseTtl`, with the existing defaults preserved.

pkgs/dart_mcp/lib/src/client/response_cache.dart
Yusufihsangorgel: I kept the monotonic stopwatch and added a test-only pause hook. The expiry comment now explains the received-time offset.

pkgs/dart_mcp/lib/src/client/response_cache.dart
Yusufihsangorgel: Pending entries now hold shared result futures. A waiter reuses the resulting cache entry when the response is cacheable.

pkgs/dart_mcp/lib/src/client/response_cache.dart
Yusufihsangorgel: AsyncCache lacks keyed entries, per-entry TTLs, targeted invalidation, and a size bound, so this cache remains local to `dart_mcp`.

pkgs/dart_mcp/lib/src/client/response_cache.dart
Yusufihsangorgel: Current invalidation covers change notifications and invalid cursors. Broader proactive busting remains a policy choice because polling adds requests without a protocol signal.

pkgs/dart_mcp/lib/src/client/response_cache.dart
Yusufihsangorgel: Concurrent callers now wait for the first response, reuse it when cacheable, and send independently after a non-cacheable result.

pkgs/dart_mcp/lib/src/client/response_cache.dart
Yusufihsangorgel: Expiry now records the stopwatch reading taken when the response arrives plus its TTL. The nearby comments spell out that comparison.

pkgs/dart_mcp/lib/src/client/response_cache.dart
Yusufihsangorgel: Updated-resource sets are now just created when an in-flight read actually needs one.

Yusufihsangorgel:
Thanks, the stopwatch stays. I added short comments for the monotonic expiry value and the test offset.

jakemac53 (approved):

pkgs/dart_mcp/lib/src/client/response_cache.dart
```diff
@@ -0,0 +1,316 @@
...
+        ttlMs <= 0 ||
+        !CacheScope.values.any((scope) => scope.name == cacheScope)) {
+      return result;
+    }
+
+    final ttl =
+        ttlMs > maxTtl.inMilliseconds ? maxTtl : Duration(milliseconds: ttlMs);
+    if (ttl <= Duration.zero || maxEntries == 0) return result;
+    if (!_entries.containsKey(key) && _entries.length >= maxEntries) {
+      _entries.remove(_entries.keys.first);
```
jakemac53, line 178: Ideally, we would actually remove the entry that will expire first or something like that? Although maybe it's best to just keep it simple.

pkgs/dart_mcp/lib/src/client/response_cache.dart
jakemac53: Yeah I was thinking something like a periodic timer that just goes through and removes stale entries. But given the cache is bounded in size, maybe it's not really worthwhile. We can easily add it later in any case.

jakemac53 (approved):

jakemac53:
Will land this tomorrow, let me know if you don't want me to before then and I won't. A few minor points open for discussion but I am happy to land this as-is, and you can choose whether or not you want to follow up 👍

pkgs/dart_mcp/lib/src/client/response_cache.dart
Yusufihsangorgel: Agreed. Stale entries get dropped on the read path anyway, so a sweep would only reclaim the ones nobody reads again.

pkgs/dart_mcp/lib/src/client/response_cache.dart
Yusufihsangorgel: Picking the one that expires first means scanning the whole map, so I kept the simpler drop.

Yusufihsangorgel:
Please go ahead. Nothing I'd change before it lands.

pkgs/dart_mcp/lib/src/client/response_cache.dart
```diff
@@ -0,0 +1,316 @@
...
+/// is complete, carries a positive `ttlMs` and carries a scope the schema
+/// allows, and the answers to an [InputRequiredResult] stay out entirely. A
+/// `ttlMs` past the configured maximum is clamped to that limit. Without a
+/// bound a server could pin an answer for the life of the process, and a value
+/// large enough to overflow a [Duration] would write an entry that is already
+/// stale.
+///
+/// https://modelcontextprotocol.io/specification/2026-07-28/server/utilities/caching
+final class _ClientResponseCache {
+  static const _maxDuration = Duration(microseconds: 9223372036854775807);
```
jakemac53: This is causing the dart2js tests to complain due to the large size. Where exactly is the number coming from anyways though? Can we just use milliseconds? I am confused why its using microseconds at all given that ttl is always expressed in milliseconds.

Yusufihsangorgel: That was the largest a `Duration` holds, saturating the expiry. But `maxTtl` already bounds the ttl, so it only mattered if you set the maximum near it. Dropped it and the two tests that needed the literal.

PR #636, opened 2026-08-29
serve the subscriptions/listen notification stream

Yusufihsangorgel:

The `subscriptions/listen` types landed without a handler. I put the acknowledgement contract on the mixin instead of the Streamable HTTP dispatcher, since the spec states it for every transport. It stamps the subscription id on the first message, holds the request until shutdown, and refuses a listen the transport did not name. Matching list and resource notifications go onto the open response, and closing it just ends the subscription without a final result.

An embedder can feed a shared notification stream so a change from one request reaches another listen. Public factories, the subscription-id shape, and the `discover` advertisement change with this wire.


pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
```diff
@@ -0,0 +1,177 @@
...
+  /// `io.modelcontextprotocol/subscriptionId`.
+  ///
+  /// Throws an [RpcException] with `-32602` if the filter is not the shape
+  /// the schema describes, and `-32600` if [nextSubscriptionId] is missing
+  /// or already names an open subscription.
+  FutureOr<SubscriptionsListenResult> handleSubscriptionsListen(
+    SubscriptionsListenRequest request,
+  ) async {
+    final fields = request as Map<String, Object?>;
+    final notifications = fields[Keys.notifications];
```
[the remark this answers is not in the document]
Yusufihsangorgel: I moved the read and the clear to the top, so a validation failure cannot leave the id behind. The later assignment is gone.

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -600,10 +715,20 @@ class _Answer {
...
+      });
+    }
   }
 
   /// Sends [result] and closes.
   Future<void> finish(Map<String, Object?> result) async {
+    cancel();
     if (_committed) {
       _response.write(_sseEvent(result));
     } else {
```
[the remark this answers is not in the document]
Yusufihsangorgel: Added the guard. There is a test now for finish running after initialize has already closed the response.

pkgs/dart_mcp/lib/src/api/subscriptions.dart
```diff
@@ -76,41 +76,67 @@ extension type SubscriptionsListenRequest.fromMap(Map<String, Object?> _value)
...
   ///
-  /// The notifications delivered on the stream carry it under the
+  /// Every message on the stream carries it under the
   /// `io.modelcontextprotocol/subscriptionId` metadata key, which is how a
-  /// client tells its streams apart.
+  /// client tells its streams apart. This reads it off the message's
+  /// [MetaWithSubscriptionId].
+  RequestId get subscriptionId =>
+      MetaWithSubscriptionId.fromMap(
+        (_value[Keys.meta] as Map<String, Object?>?) ?? const {},
```
jakemac53: The `.subscriptionId` below will fail if we have no metadata, probably better to fail earlier here and say there was no metadata at all provided instead of defaulting a const object.

pkgs/dart_mcp/lib/src/api/subscriptions.dart
```diff
@@ -76,41 +76,67 @@ extension type SubscriptionsListenRequest.fromMap(Map<String, Object?> _value)
...
+/// own: the [subscriptionId] travels in its metadata.
+///
+/// From the 2026-07-28 revision.
+extension type SubscriptionsListenResult.fromMap(Map<String, Object?> _value)
+    implements Result, WithSubscriptionId {
+  factory SubscriptionsListenResult({required MetaWithSubscriptionId meta}) =>
+      SubscriptionsListenResult.fromMap({Keys.meta: meta});
+
+  /// The metadata this message carries, including the [subscriptionId].
+  MetaWithSubscriptionId? get meta =>
```
jakemac53: Afaict, this shouldn't be nullable https://modelcontextprotocol.io/specification/draft/schema#subscriptionslistenresult (same below)

pkgs/dart_mcp/lib/src/api/subscriptions.dart
```diff
@@ -76,41 +76,67 @@ extension type SubscriptionsListenRequest.fromMap(Map<String, Object?> _value)
...
-  factory SubscriptionsListenResult({
-    required RequestId subscriptionId,
-    Meta? meta,
-  }) => SubscriptionsListenResult.fromMap({
-    Keys.meta: {...?meta?._value, Keys.subscriptionIdMeta: subscriptionId},
-  });
-
+/// This type is not intended to be constructed directly and thus has no public
+/// constructor.
+extension type WithSubscriptionId._fromMap(Map<String, Object?> _value) {
```
jakemac53, line 86: Consider adding the `MetaWithSubscriptionId get meta` getter directly to this type - then you can just call it inside `subscriptionId` (and rely on that error message when its missing), and remove the duplicate implementations in the actual types that mix this in.

pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
```diff
@@ -0,0 +1,185 @@
...
+
+  /// The first [shutdown] call, which every later one waits on.
+  Completer<void>? _shutdown;
+
+  /// The id the next `subscriptions/listen` request opens its subscription
+  /// under.
+  ///
+  /// A handler cannot read the JSON-RPC id of the request it answers, and a
+  /// subscription is named by that id. The transport serving the request sets
+  /// this before delivering it. Leaving it `null` refuses the request.
```
jakemac53, line 27: Note that I did file https://github.com/dart-lang/tools/issues/2086, and ideally we would just resolve it by fixing this underlying issue. This currently solution I think could have potential race conditions etc

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -96,14 +98,22 @@ export 'src/utils/sse.dart' show sseMessageStream;
...
+/// A listen response writes SSE comments on a timer to keep an idle stream
+/// open. Closing that response shuts down its request server and ends the
+/// subscription without a final result.
 /// [onNotification] sees every notification either way, held back or not.
+/// When [subscriptionNotifications] is provided, each listen request reads
+/// matching changes from that stream. An embedder can pass the same broadcast
+/// stream to every request and add [onNotification] values to it, allowing a
+/// change produced by one request's server to reach another request's listen
+/// stream.
 Future<void> handleStreamableHttpRequest(
```
jakemac53: This function is getting quite unwieldy - I would run this cognitive complexity skill over it https://github.com/kevmoo/analytica.dart/tree/main/skills/dart-cognitive-complexity to try and make it easier to follow.

Or you can just use the linked tool and manually try and get the complexity score down. It is getting pretty hard to follow along now.

It would be OK to do this as a follow-up to avoid crazy merge conflicts if desired.

pkgs/dart_mcp/lib/src/api/subscriptions.dart
Yusufihsangorgel: Reading it now throws on the missing field. The error names `_meta` and not the subscription key.

pkgs/dart_mcp/lib/src/api/subscriptions.dart
Yusufihsangorgel: You're right. The result requires `_meta` in the schema, and the notification meta's `subscriptionId` is a MUST on every stream notification. Neither is nullable now.

pkgs/dart_mcp/lib/src/api/subscriptions.dart
Yusufihsangorgel: Moved it there. `Result` and `Notification` both declare `meta`, so the types that mix this in still redeclare it, but only to delegate.

pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
Yusufihsangorgel: A server here handles one message and is shut down after, so nothing else observes the field. A shared server would need that fix.

pkgs/dart_mcp/lib/streamable_http.dart
Yusufihsangorgel: Fair, it has grown. I left the complexity pass out of this one.

pkgs/dart_mcp/lib/src/api/subscriptions.dart
```diff
@@ -133,8 +138,10 @@ extension type SubscriptionsListenResult.fromMap(Map<String, Object?> _value)
       SubscriptionsListenResult.fromMap({Keys.meta: meta});
 
   /// The metadata this message carries, including the [subscriptionId].
-  MetaWithSubscriptionId? get meta =>
-      _value[Keys.meta] as MetaWithSubscriptionId?;
+  ///
+  /// [Result] and [Notification] both declare a nullable `meta`, so each type
```
jakemac53, line 142: Interesting, I guess it makes sense though that you would get the LUB of any conflicting members when implementing multiple extension types. 🤷‍♂️

pkgs/dart_mcp/lib/src/server/streamable_http.dart
```diff
@@ -0,0 +1,970 @@
...
+      ..bufferOutput = false
+      ..headers.contentType = ContentType(
+        'text',
+        'event-stream',
+        charset: 'utf-8',
+      )
+      ..headers.set(HttpHeaders.cacheControlHeader, 'no-cache, no-transform')
+      ..headers.set('x-accel-buffering', 'no');
+    if (keepAliveInterval case final interval?) {
+      _firstKeepAlive = Timer(const Duration(seconds: 1), () {
```
jakemac53: Why does the first one not use `interval`?

pkgs/dart_mcp/lib/src/api/subscriptions.dart
Yusufihsangorgel: It doesn't actually take the LUB. Dropping the redeclare gives `extension_type_inherited_member_conflict`, which asks for exactly this.

pkgs/dart_mcp/lib/src/server/streamable_http.dart
Yusufihsangorgel: No good reason. It was short so the tests could see a comment inside their timeout. The interval is a parameter now and they pass 50ms.

jakemac53 (approved):

pkgs/dart_mcp/lib/src/api/subscriptions.dart
jakemac53: Oh interesting, that also makes sense, in many cases they wouldn't be compatible at all.

pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
jakemac53: Yeah, I don't love that it has to be a public field etc. I think this is fine to land right now, but it would be really nice to get a proper fix in before we publish if possible.

pkgs/dart_mcp/lib/src/server/server.dart
```diff
@@ -208,32 +209,40 @@ abstract base class MCPServer extends MCPBase {
...
+  /// from that request's server. A server with [SubscriptionsSupport] therefore
+  /// advertises the capabilities it registered. Without that mixin the four
+  /// bits standing for those notifications come off here: `listChanged` on
+  /// each of the three, and `subscribe` on [ServerCapabilities.resources].
+  /// Every other key the server registered is passed through, and
+  /// `initializeLegacy` keeps all of them, since the revisions that handshake
+  /// negotiates still serve `resources/subscribe` and send the list changes
+  /// without a filter.
+  ServerCapabilities get _advertisedCapabilities =>
+      this is SubscriptionsSupport
```
natebosch, line 223: Why is this being handled with a type check on `this` instead of method overrides?

pkgs/dart_mcp/lib/src/server/streamable_http.dart
```diff
@@ -431,9 +442,33 @@ Future<void> handleStreamableHttpRequest(
...
+  MCPServer? activeServer;
+  StreamSubscription<Map<String, Object?>>? notificationSubscription;
+  var responseClosed = false;
+  var listenFailed = false;
+  if (method == SubscriptionsListenRequest.methodName) {
+    unawaited(() async {
+      try {
+        await response.done;
+      } catch (_) {
+        // A disconnected client cannot receive another response.
```
natebosch, line 463: Can this catch a specific type of exception? `on Exception` at least?

pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
```diff
@@ -0,0 +1,185 @@
...
+  /// Ends each open subscription, under the id it was opened with.
+  final Map<RequestId, Completer<void>> _subscriptions = {};
+
+  /// The first [shutdown] call, which every later one waits on.
+  Completer<void>? _shutdown;
+
+  /// The id the next `subscriptions/listen` request opens its subscription
+  /// under.
+  ///
+  /// A handler cannot read the JSON-RPC id of the request it answers, and a
```
natebosch, line 25: [nit] capitalize ID in prose

pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
natebosch: Can you elaborate on this workaround? I don't understand what race condition you are concerned about.

pkgs/dart_mcp/lib/src/server/server.dart
Yusufihsangorgel: No reason beyond how it grew. It is a protected `advertisedCapabilities` getter now, overridden in `SubscriptionsSupport`.

pkgs/dart_mcp/lib/src/server/streamable_http.dart
Yusufihsangorgel: Narrowed to `on IOException`, the type `dart:io` raises for a socket or HTTP failure on that path.

pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
Yusufihsangorgel: Capitalized across the library prose, and in both error messages here.

pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
Yusufihsangorgel: Each listen request gets its own server. The transport calls the factory per message in `handleRequestScopedMessage`, and the slot is never shared. Two concurrent listens keep their own IDs in a test. The ID on `Parameters` (dart-lang/tools#2583) removes the slot.

PR #638, opened 2026-08-29
add a Streamable HTTP client channel

Yusufihsangorgel:
Streamable HTTP still lacked a client channel. I put `streamableHttpClientChannel` next to the server handler so the header names stay one definition.

Failures come back on the id that asked for them. A notification has none, so it only hears about a POST that already went out.

The helper requires a protocol version up front, since it never sends `initialize`. The Tools page requires Streamable HTTP clients to exclude invalid `tools/list` entries (https://modelcontextprotocol.io/specification/2026-07-28/server/tools#x-mcp-header).


pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -21,13 +22,364 @@ import 'dart:io';
...
+  } catch (error) {
+    if (!message.containsKey(Keys.id)) return;
+    yield {
+      Keys.jsonrpc: '2.0',
+      Keys.id: message[Keys.id],
+      Keys.error: {
+        Keys.code: error_code.SERVER_ERROR,
+        Keys.message: error.toString(),
+      },
+    };
```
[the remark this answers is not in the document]
Yusufihsangorgel: Right, fixed. A later result that reuses the stale id also loses tools whose `x-mcp-header` is invalid, so the test posts one of those.

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -21,13 +22,421 @@ import 'dart:io';
...
+/// is not an inbound message. Valid `x-mcp-header` annotations from
+/// `tools/list` are mirrored on later `tools/call` requests; invalid tool
+/// definitions are dropped. Does not send `initialize`.
+StreamChannel<Map<String, Object?>> streamableHttpClientChannel(
+  Uri uri, {
+  required ProtocolVersion protocolVersion,
+  required ClientCapabilities clientCapabilities,
+  Implementation? clientInfo,
+}) {
+  if (protocolVersion != ProtocolVersion.v2026_07_28) {
```
jakemac53: For future versions, I would like something that is a bit easier to update so we don't have to remember to come update this, maybe add a `supportsStreamableHttp` boolean to ProtocolVersion? That way, whenever we add a new version, it will be obvious that we need to fill that in.

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -21,13 +22,421 @@ import 'dart:io';
...
+import 'package:stream_transform/stream_transform.dart';
 
 import 'server.dart';
 import 'src/utils/constants.dart';
 import 'src/utils/json_rpc_2_object.dart';
+import 'src/utils/sse.dart' show sseMessageStream;
 
 export 'src/utils/sse.dart' show sseMessageStream;
 
+/// Creates a [StreamChannel] which POSTs each JSON-RPC message to [uri].
```
jakemac53: I think at this point it's worth separating the client and server parts of this into separate libraries, and then exporting those from here (using show).

That will make this file easier to follow and it will be obvious what the actual public API is as well.

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -21,13 +22,421 @@ import 'dart:io';
...
+  void recordOutgoing(Map<String, Object?> message, String method) {
+    if (method == ListToolsRequest.methodName && message.containsKey(Keys.id)) {
+      final params = message[Keys.params];
+      final cursor =
+          params is Map<String, Object?> ? params[Keys.cursor] : null;
+      _listToolsRequestIds[message[Keys.id]] = cursor == null;
+    }
+  }
+
+  Map<String, String> headersFor(Map<String, Object?> message) {
```
jakemac53: This method could use some doc comments explaining what the headers its generating are for and also probably some comments in the body explaining what it's doing at specific parts.

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -21,13 +22,421 @@ import 'dart:io';
...
+          'cannot be mirrored onto ${declaration.header}. The tool declares '
+          'it as ${declaration.type}.',
+        );
+      }
+      headers['$_mcpParamHeaderPrefix${declaration.header}'] = encoded;
+    }
+    return headers;
+  }
+
+  Map<String, Object?> recordIncoming(Map<String, Object?> message) {
```
jakemac53: Same here, document the high level purpose of this function

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -21,13 +22,421 @@ import 'dart:io';
...
+      Keys.error: {
+        Keys.code: error_code.SERVER_ERROR,
+        Keys.message: error.toString(),
+      },
+    });
+  }
+}
+
+/// Request state shared by the POSTs on one client channel.
+final class _StreamableHttpClientState {
```
jakemac53: This class seems to be doing a lot of special stuff around tool list methods but its not clear why that one is specialized so much, some high level documentation on the class would also be helpful.

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -21,13 +22,421 @@ import 'dart:io';
...
+Object? _valueAtPath(Map<String, Object?> arguments, List<String> path) {
+  Object? node = arguments;
+  for (final key in path) {
+    if (node is! Map<String, Object?>) return null;
+    node = node[key];
+  }
+  return node;
+}
+
+const _subschemaKeywords = {
```
jakemac53: A link to where these come from would be good, as well as maybe document the last date on which this was updated.

pkgs/dart_mcp/lib/streamable_http.dart
```diff
@@ -681,6 +1090,21 @@ String _decodeSentinel(String value) =>
         ? utf8.decode(base64.decode(value.substring(9, value.length - 2)))
         : value;
 
+/// Encodes [value] when it is not a safe plain ASCII header value.
+String _encodeSentinel(String value) {
+  final matchesSentinel = value.startsWith('=?base64?') && value.endsWith('?=');
+  final needsEncoding =
+      value.isEmpty ||
```
jakemac53: If this list of conditions is documented a link would also be good here

pkgs/dart_mcp/test/streamable_http_test.dart
```diff
@@ -174,6 +174,1328 @@ void main() {
   String jsonBody(String response) =>
       response.substring(response.indexOf('{'), response.lastIndexOf('}') + 1);
 
+  group('client channel', () {
```
jakemac53, line 177: It would be good to add a test that multiple concurrent requests can be sent, received back out of order, and still be handled correctly.

jakemac53 (commented):
🚀

pkgs/dart_mcp/lib/streamable_http.dart
Yusufihsangorgel: That makes sense. `supportsStreamableHttp` is required for every `ProtocolVersion`, so a new revision cannot compile until the support decision is explicit.

pkgs/dart_mcp/lib/streamable_http.dart
Yusufihsangorgel: The 19-line facade now exports dedicated client, server, and SSE entry points through explicit `show` lists, while implementation details live in separate libraries.

pkgs/dart_mcp/lib/streamable_http.dart
Yusufihsangorgel: Headers are now documented at the method level, with inline comments covering call-shape validation, path lookup, and typed value encoding.

pkgs/dart_mcp/lib/streamable_http.dart
Yusufihsangorgel: `recordIncoming` now documents cache refresh, pagination, invalid tool filtering, and list-change invalidation.

pkgs/dart_mcp/lib/streamable_http.dart
Yusufihsangorgel: I clarified why this state is specialized. It tracks tool-list page roles and shared header snapshots across concurrent POSTs.

pkgs/dart_mcp/lib/streamable_http.dart
Yusufihsangorgel: References now cover MCP's annotation boundary, JSON Schema applicators, and `contentSchema`, with a 2026-08-31 checked date.

pkgs/dart_mcp/lib/streamable_http.dart
Yusufihsangorgel: Value-encoding now links directly to the Streamable HTTP specification section defining the implemented sentinel conditions.

pkgs/dart_mcp/test/streamable_http_test.dart
Yusufihsangorgel: Multiple requests now run concurrently in the test, with responses intentionally returned in 2, 3, 1 order and matched to their IDs.

jakemac53 (approved):

PR #642, opened 2026-08-30
let the create tool target the project root

Yusufihsangorgel:
The create tool rejected a `directory` of `.`. The check leans on `p.isWithin`,
and that is strict enough to call `isWithin('.', '.')` false, reading it as an
escape from the project root.

I let the root through for Flutter, where `flutter create .` works. A Dart
project still needs a subdirectory, because the Dart CLI wants an empty target
unless it is forced, and forcing is not something this tool does. The error says
that now. An empty string lands in the same place as `.`, and it is rejected on
its own.


pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
```diff
@@ -154,7 +154,7 @@ base mixin DashCliSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
           details: 'Directory must be a relative path.',
         ),
       );
-    } else if (!p.isWithin('.', directory)) {
+    } else if (p.normalize(directory) != '.' && !p.isWithin('.', directory)) {
```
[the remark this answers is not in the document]
Yusufihsangorgel: Switched to `p.equals`. Both forms agree on `.`, `./`, `''`, `a/..` and `./sub`, so this one says it more directly.

pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
```diff
@@ -146,22 +146,41 @@ base mixin DashCliSupport on ToolsSupport, LoggingSupport, RootsTrackingSupport
...
-    } else if (!p.isWithin('.', directory)) {
+    } else if (!p.equals('.', directory) && !p.isWithin('.', directory)) {
       errors.add(
         ValidationError(
           ValidationErrorType.custom,
           path: [ParameterNames.directory],
           details: 'Directory must be a relative path within the project root.',
         ),
       );
+    } else if (projectType == 'dart' && p.equals('.', directory)) {
```
jakemac53: Can we rely directly on `dart create` to give a nice error here? It seems wrong that it does not support this while `flutter create` does, so I don't love hard coding this expectation.

jakemac53 (commented):
Note that this tool (and many others) are actually disabled and we are not generally actively working on them. I don't mind reviewing a small PR like this, but larger PRs probably we would not prioritize.

In general, I don't think it's worth your time to try and improve these tools that are not enabled by default.

pkgs/dart_mcp_server/lib/src/mixins/dash_cli.dart
Yusufihsangorgel: You're right. I removed the special case and now pass `.` to `dart create`, which reports the existing directory and suggests `--force`.

Yusufihsangorgel:
Thanks, that context helps. This keeps these disabled-by-default tools in the small-fix category for me, not the larger-investment category.

jakemac53 (approved):

PR #655, opened 2026-09-04
remove elicit, listRoots and createMessage from the server

Yusufihsangorgel:
The server request methods do not fit with how InputRequiredResult works, because handlers return input requirements as results. I removed elicit, listRoots and createMessage, a breaking API change.

Handlers return an InputRequiredResult instead. On revisions before 2026-07-28 a shim wraps tools/call, prompts/get and resources/read at registration, sends each InputRequest as that revision's own request and reruns the handler with the answers, up to MCPServer.maxInputRequiredRounds rounds (8 by default). WithInputResponses gained elicitResponse, sampleResponse and listRootsResponse, and a handler reads an answer as the type it asked for.

Part of dart-lang/ai#668.


pkgs/dart_mcp/lib/src/server/server.dart
```diff
@@ -310,6 +310,7 @@ abstract base class MCPServer extends MCPBase {
   /// capability, and throws an [RpcException] with
   /// [McpErrorCodes.missingRequiredClientCapability] when it has not, naming
   /// the capability the client is missing under `data.requiredCapabilities`.
+  @Deprecated('Use an `InputRequiredResult` instead.')
```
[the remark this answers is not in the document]
Yusufihsangorgel: Analysis is clean on this branch. That diagnostic doesn't report inside the package, and I checked by dropping either of the two ignore comments already in `api.dart`. Nothing changed.

jakemac53 (commented):
I think probably this message should be a bit more nuanced - because this is still actually the intended way for things to work on older protocol versions.

I think quite possibly we should be replacing these functions with new ones, which can somehow handle multiple protocol versions.

In general I think we need to think more deeply in general about how we want servers to handle multiple versions, especially until such a time as the newest MCP version is widely supported everywhere.

If somebody just switches to returning an `InputRequiredResponse` always, they need to have also somehow indicated that they do not support old versions any more. Ideally, we would statically enforce this contract in some way. Or, just handle it for people.

Yusufihsangorgel:
Reworded it so the older revisions keep this path.

On the bigger question, I just looked at the other SDKs. Go, Python and TypeScript all bridge one author handler onto both shapes at runtime. Rust hands back a typed result family instead. None of the four enforces the contract statically.

#654 already routes the legacy side for a request-scoped server, so what is left is the modern bridge.

jakemac53:
> On the bigger question, I just looked at the other SDKs. Go, Python and TypeScript all bridge one author handler onto both shapes at runtime. Rust hands back a typed result family instead. None of the four enforces the contract statically.

What do these handlers look like in those languages? What does it actually look like to use it? The paradigms are so fundamentally different here... we can't even guarantee the same server gets the retry request?

jakemac53:
Concretely, I could imagine keeping these functions, except that they would return _either_ an InputRequiredResult or the legacy result type. If it is an `InputRequiredResult`, then we can just give the server an opportunity to put some `requestState` as well (plus possibly carrying along any existing request state?), but they are expected to return that result immediately after adding any state to it.

Then, we don't deprecate the methods at all, but we are making a breaking change to return a different type, probably a new union type.

Maybe, if we can detect that the thing being requested was already answered (by reading from `inputResponses` and just return the result instead of an `InputRequiredResult`), so servers don't have to check themselves. But this might be tricky.

Yusufihsangorgel:
Correcting my own comment first. Python only bridges `@mcp.tool()` with `Annotated[T, Resolve(fn)]`, not `ctx.elicit()`. Rust carries a typed family in the `ServerHandler` signature and bridges on the client. And Go does seal the input set statically, with an unexported method on the interface.

On the retry landing somewhere else, you are right, and I think that is why all four replay instead of parking. Go wrote it down in `design/mrtr.md:263`, rejecting suspend and resume partly because "it should be possible for a retry to arrive on any server instance". Our own dispatcher builds a new server per request at `request_scoped.dart:144`, so replay is all there is here anyway.

The shapes are `legacyInputRequiredShim.ts:163`, `mrtr.go:189` and `mcpserver/resolve.py:570`. The author writes the modern handler and the library downgrades it for an older client.

Yusufihsangorgel:
I just checked what an older client does with an `InputRequiredResult`. It does not interpret the result and never retries.

The transports already split on this. Streamable HTTP serves only 2026-07-28, and stdio still takes older revisions through `initializeLegacy`.

jakemac53:
I like the general approach of forcing all servers to operate in terms of InputRequiredRequest/InputRequiredResult, because that is easier to create a backwards compatible shim for older clients using the old protocol.

If we assume we take that approach, then I think we should probably not just deprecate but actually entirely remove the current methods, as they are not really possible to shim in the other direction.

(but it is important that we do automatically put in a shim for older clients as a part of that)

Yusufihsangorgel:
Agreed. Removal looks workable. Each of them is a thin wrapper over `sendRequest`, and the shim can send the legacy request itself and re-run the handler with the responses. I will turn this PR into that.

One place it does not reach is a server asking outside a request handler. I hit it in RootsTrackingSupport, where updateRoots calls listRoots once initialize has settled. The same happens in the roots and sampling examples. Those can call `sendRequest` directly on the older revisions, which is what the wrappers already do.

jakemac53:
> Agreed. Removal looks workable. Each of them is a thin wrapper over `sendRequest`, and the shim can send the legacy request itself and re-run the handler with the responses. I will turn this PR into that.

SGTM 🚀 

> One place it does not reach is a server asking outside a request handler. I hit it in RootsTrackingSupport, where updateRoots calls listRoots once initialize has settled. The same happens in the roots and sampling examples. Those can call `sendRequest` directly on the older revisions, which is what the wrappers already do.

Yeah, basically this is just not supported any more so servers will have to figure out what that means for each use case.

So, the Dart MCP Server will have to figure out how best to handle this - we don't have to collect the roots so eagerly, the effect will just be a slower first call to analyze, although I think we anyways made that more lazy so it won't actually matter.

jakemac53:
> One place it does not reach is a server asking outside a request handler. I hit it in RootsTrackingSupport, where updateRoots calls listRoots once initialize has settled. The same happens in the roots and sampling examples. Those can call `sendRequest` directly on the older revisions, which is what the wrappers already do.

Oh I forgot this is actually part of the dart_mcp package, so I suppose we will have to figure that out here. It could take the approach of only initializing roots on the first call from the client though or something.

Yusufihsangorgel:
Deferring it looks right. On the 2026-07-28 revision only a tool, prompt or resource handler can ask the client for anything, by returning an `InputRequiredResult`. The first `updateRoots` runs from `initialize`, so deferring changes when it happens but not where the request comes from. I put the removal and the shim in this PR, and left the roots side for separate work in the same package.

pkgs/dart_mcp/example/elicitations_server.dart
```diff
@@ -106,60 +106,60 @@ base class MCPServerWithElicitation extends MCPServer
       );
     }
 
-    await _startElicitationFlow();
-    return CallToolResult(content: [Content.text(text: 'Success!')]);
-  }
+    final userInfo = request.inputResponses?['userInfo'];
+    if (userInfo is ElicitResult) {
```
jakemac53: For examples we should try and guide people to the safer approach here, and these `is` checks are not something we want people to be doing generally.

We could either expose some safe extension methods to do a similar check, possibly attached to some union type, or just manually check the type etc.

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
```diff
@@ -0,0 +1,182 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of 'server.dart';
+
+/// How many rounds a handler may answer `input_required` before the shim gives
+/// up. This guards against a handler that never settles, so it is not a knob.
+const _legacyInputRequiredMaxRounds = 8;
```
jakemac53: I do think this should be configurable and this should just be a default, see for instance https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/server/src/server/legacyInputRequiredShim.ts#L62C36-L62C50

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
```diff
@@ -0,0 +1,182 @@
...
+
+/// Replays an input-required handler through the request style of older
+/// protocol revisions.
+final class _LegacyInputRequiredShim {
+  _LegacyInputRequiredShim(this._server);
+
+  final MCPServer _server;
+
+  /// Counts the URL elicitations this server has named. The revision that takes
```
jakemac53:
```suggestion
  /// Counts the URL elicitations this server has made. The revision that takes
```

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
```diff
@@ -0,0 +1,182 @@
...
+  /// them asks for an `elicitationId` unique to the server and opaque to the
+  /// client, and a count is both.
+  var _elicitations = 0;
+
+  Future<T> fulfill<T extends Result>(
```
jakemac53: please add some API docs to this

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
```diff
@@ -0,0 +1,182 @@
...
+    if (_server.protocolVersion >= ProtocolVersion.v2026_07_28) {
+      return result as T;
+    }
+
+    for (var round = 0; result.isInputRequired; round++) {
+      if (round >= _legacyInputRequiredMaxRounds) {
+        throw RpcException(
+          error_code.INTERNAL_ERROR,
+          'The server returned input_required after '
+          '$round retries for $methodName.',
```
jakemac53:
```suggestion
          '$round retries for $methodName, which has exceeded the maximum.',
```

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
```diff
@@ -0,0 +1,182 @@
...
+        );
+      }
+
+      final inputRequired = result as InputRequiredResult;
```
jakemac53, line 72: Could this not be a `CallToolResult` etc at this point? I don't see how this is intended to exit the loop once the requests are fulfilled in general. I think adding some brief comments above the major sections explaining the high level purpose would be helpful.

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
```diff
@@ -0,0 +1,182 @@
...
+          throw _missingUrlElicitation;
+        }
+      case ElicitationMode.form:
+        if (!_server.clientCapabilities.supportsFormElicitation) {
+          throw _missingFormElicitation;
+        }
+    }
+  }
+
+  Future<Result> _sendInputRequest(InputRequest inputRequest) async {
```
jakemac53: nit: maybe call this `sendLegacyInputRequest`, and/or document that this is really translating the requests to legacy requests.

pkgs/dart_mcp/lib/src/server/resources_support.dart
```diff
@@ -210,24 +210,32 @@ base mixin ResourcesSupport on MCPServer {
...
+          error_code.INVALID_PARAMS,
+          'Resource not found',
+          data: {Keys.uri: request.uri},
+        );
+      }
+      return response;
     }
-    return response;
+
+    return _legacyInputRequiredShim.fulfill(
```
jakemac53: Can we handle this at a higher level, intercepting _all_ requests and doing the shim (more like a middleware solution) instead of having to manually handle it in each of the supported APIs?

pkgs/dart_mcp/example/elicitations_server.dart
Yusufihsangorgel: Readers are elicitResponse, sampleResponse and listRootsResponse on WithInputResponses, one per arm of the InputResponse union. A missing key comes back as null, and a payload that lacks that arm's fields throws. The example now reads elicitResponse, since those result types erase to Map and is-checks cannot tell them apart.

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
Yusufihsangorgel: Left it on MCPServer.maxInputRequiredRounds, 8 by default, RangeError below 1. I skipped the TS round timeout, because sendRequest has no timeout here.

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
Yusufihsangorgel: Sure, landed it.

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
Yusufihsangorgel: Put docs on fulfill, the class, and the registerRequestHandler wrapper.

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
Yusufihsangorgel: Kept that wording.

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
Yusufihsangorgel: Loop returns once the result is not input_required. A CallToolResult never hits the cast. I put that check at the top, above the round limit, and each of the three blocks has a short note: read the requests, send them on the old path, rerun the handler.

pkgs/dart_mcp/lib/src/server/legacy_input_required_shim.dart
Yusufihsangorgel: Called it _sendLegacyInputRequest. The doc says it maps an InputRequest onto whatever that old revision actually sent.

pkgs/dart_mcp/lib/src/server/resources_support.dart
Yusufihsangorgel: Wrapped the three methods that can answer input_required inside registerRequestHandler. The rest go through. Mixins match main again. Direct registrations are covered, and I share the result checks with the 2026-07-28 dispatcher.

pkgs/dart_mcp/example/elicitations_server.dart
```diff
@@ -106,8 +106,8 @@ base class MCPServerWithElicitation extends MCPServer
       );
     }
 
-    final userInfo = request.inputResponses?['userInfo'];
-    if (userInfo is ElicitResult) {
+    final userInfo = request.elicitResponse('userInfo');
```
jakemac53: nice, I like this 🚀

pkgs/dart_mcp/lib/src/api/input_required.dart
```diff
@@ -146,11 +146,65 @@ extension type WithInputResponses._fromMap(Map<String, Object?> _value)
...
+  /// reads each back as the type it asked for under that key, through
+  /// [elicitResponse], [sampleResponse] or [listRootsResponse]. A key it did
+  /// not ask for is one to ignore.
   Map<String, Result>? get inputResponses =>
       (_value[Keys.inputResponses] as Map?)?.cast<String, Result>();
 
+  /// The elicitation arm of the schema `InputResponse` union under [key].
+  ///
+  /// Missing [key] is null. A value without `action` throws [ArgumentError].
+  ElicitResult? elicitResponse(String key) {
```
jakemac53: nit: call this `elicitResult` just to keep the nomenclature consistent with the return type

pkgs/dart_mcp/lib/src/api/input_required.dart
```diff
@@ -146,11 +146,65 @@ extension type WithInputResponses._fromMap(Map<String, Object?> _value)
...
+
+  /// The roots arm of the schema `InputResponse` union under [key].
+  ///
+  /// Missing [key] is null. A value without `roots` throws [ArgumentError].
+  ListRootsResult? listRootsResponse(String key) {
+    final value = _typedInputResponse(key, const [Keys.roots]);
+    return value == null ? null : value as ListRootsResult;
+  }
+
+  Map<String, Object?>? _typedInputResponse(
```
jakemac53, line 184: nit: add a comment about the required key validation

pkgs/dart_mcp/lib/src/api/input_required.dart
```diff
@@ -146,11 +146,65 @@ extension type WithInputResponses._fromMap(Map<String, Object?> _value)
...
+    String key,
+    List<String> requiredKeys,
+  ) {
+    final responses = inputResponses;
+    if (responses == null || !responses.containsKey(key)) return null;
+    final value = responses[key];
+    if (value is! Map) {
```
jakemac53: This should be a `Map<String, Object?>` - I would just check that type here and then avoid wrapping the map below. Unless we really need a copy for some reason.

pkgs/dart_mcp/lib/src/api/input_required.dart
Yusufihsangorgel: All three are named after their result types, `elicitResult`, `createMessageResult` and `listRootsResult`.

pkgs/dart_mcp/lib/src/api/input_required.dart
Yusufihsangorgel: The loop has a comment saying the keys are the ones the schema requires on that arm.

pkgs/dart_mcp/lib/src/api/input_required.dart
Yusufihsangorgel: It checks for `Map<String, Object?>` and returns the value itself, no copy.

jakemac53: It calls `Map.from` below on line 197 which does a full copy

Yusufihsangorgel: That copy went away in fb6db0de. Line 191 now checks `value is! Map<String, Object?>` and returns `value` without copying. This thread still shows the older diff.

jakemac53: ah ok I must have been looking at an old version, thanks

jakemac53 (approved):

PR #669, opened 2026-09-08
add optional origin validation for Streamable HTTP

Yusufihsangorgel:
Adds optional `allowedOrigins` to `handleStreamableHttpRequest`, leaving default behavior unchanged. With it set, absent or allowed origins pass, while unlisted or repeated Origin headers return an empty 403 response.

I chose opt-in validation, as in TypeScript middleware and the Go handler. This is just the Origin half I said was ready in #162, and closing it is an answer if validation belongs outside the package.

Part of dart-lang/ai#162

### DIFF
--- a/pkgs/dart_mcp/lib/src/server/streamable_http.dart
+++ b/pkgs/dart_mcp/lib/src/server/streamable_http.dart
@@ -46,10 +46,11 @@ import 'server.dart';
 /// Notifications are acknowledged with `202 Accepted` and not dispatched,
 /// since this protocol revision defines no client-to-server notifications
 /// over HTTP. This handler reads a request body into memory, and caps it at
-/// [maxRequestBodyBytes]. It does not read the `Origin` header. The
-/// specification requires a server to validate that header and answer with
-/// 403. The check needs deployment knowledge this handler does not have, so it
-/// belongs to the embedding HTTP server, along with authentication.
+/// [maxRequestBodyBytes]. The specification requires a server to validate the
+/// `Origin` header and answer with 403. Pass [allowedOrigins] to have that
+/// check run here. Without it the check needs deployment knowledge this
+/// handler does not have, so it belongs to the embedding HTTP server, along
+/// with authentication.
 ///
 /// Responses produced by the dispatched server are written unchanged, so an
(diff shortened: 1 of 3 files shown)


pkgs/dart_mcp/lib/src/server/streamable_http.dart
```diff
@@ -126,6 +128,21 @@ Future<void> handleStreamableHttpRequest(
     return;
   }
 
+  if (allowedOrigins != null) {
+    // Read the header as a list, the way the checks below read theirs. A
+    // request that repeats it carries no one origin to check, so it is turned
+    // down with the ones this server does not allow.
+    final origins = request.headers['origin'];
+    if (origins != null &&
+        (origins.length != 1 || !allowedOrigins.contains(origins.single))) {
```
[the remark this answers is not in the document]
Yusufihsangorgel: Go's `CrossOriginProtection` keeps its trusted origins as the caller wrote them and matches the header against that set, with no lowercasing. Folding only the incoming side would stop a mixed-case entry in the list from ever matching, so I just left the comparison on the strings the caller passed.

pkgs/dart_mcp/test/streamable_http_test.dart
```diff
@@ -50,12 +50,14 @@ void main() {
...
   late StreamController<Map<String, Object?>> subscriptionNotifications;
+  Set<String>? allowedOrigins;
   final servers = <MCPServer>[];
   final notifications = <Map<String, Object?>>[];
 
   setUp(() async {
     serverFactory = _HttpTestServer.new;
     servers.clear();
+    allowedOrigins = null;
     notifications.clear();
```
jakemac53: nit: These cleanup style tasks should be handled in a `tearDown` instead of in `setUp`

jakemac53 (commented):
just the one nit

pkgs/dart_mcp/test/streamable_http_test.dart
Yusufihsangorgel: Moved them into a tearDown.

jakemac53 (approved):

Issue #668, opened 2026-09-08
Update to version 2026-07-28 of MCP


jakemac53:
cc @Yusufihsangorgel as you have been working on this, but I realized we didn't have a tracking issue.

Yusufihsangorgel:
POST, SSE answers, subscriptions/listen, server/discover, the Mcp-Param and Mcp-Name headers, caching, extensions, the client channel, removed-method 404s and InputRequired are on main. SSE close as cancellation, the Origin allow-list and the legacy input-required shim are in review. Still open are notifications/cancelled, pagination, a client listen helper and the dart_mcp_server migration once 0.6.0 ships.

Yusufihsangorgel:
Merged since the last note: the Origin allow-list, SSE close as cancellation, the removal of the three server methods, two readme changes, the #636 follow-ups and a weekly conformance run. Open as of 2026-09-16: a client example (#671), sampling tool content (#672), DEVELOPING.md (#673), a multi-version example (#676), the receiving half of cancellation (#678), a client subscription helper (#679), a timeout for the conformance client (#681) and a paging walk for the four list requests (#682).

Yusufihsangorgel:
Since the last note, #672 (sampling tool content), #681 (conformance client timeout), #673 (DEVELOPING.md), #676 (multi-version example), #682 (list paging), #684 (415 for the wrong media type) and #685 (sampling content as a list) have merged. Still open as of 2026-09-22 are the approved client example (#671), cancellation on the receiving side (#678) and the client subscription helper (#679), with #688 tracking a client-side initialization helper.

PR #671, opened 2026-09-11
add a streamable http client example

Yusufihsangorgel:
Adds a client to pair with `example/streamable_http_server.dart`. It takes the URL the server prints, discovers the server, lists its tools, calls `greet` and prints the progress notification. The example readme now describes the pair instead of the curl command.

Part of https://github.com/dart-lang/ai/issues/668

### DIFF
--- a/pkgs/dart_mcp/example/streamable_http_client.dart
+++ b/pkgs/dart_mcp/example/streamable_http_client.dart
@@ -1,0 +1,94 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// A client that connects over Streamable HTTP to
+/// `streamable_http_server.dart`.
+///
+/// Pass the printed URL to discover the server, list tools, call `greet`,
+/// and print progress.
+library;
+
+import 'dart:io';
+
+import 'package:dart_mcp/client.dart';
+import 'package:dart_mcp/streamable_http.dart';
+
+void main(List<String> args) async {
+  if (args.length != 1) {
+    stderr.writeln(
+      'Usage: dart run example/streamable_http_client.dart <url>\n'
+      'Pass the URL printed by example/streamable_http_server.dart.',
+    );
+    exitCode = 64;
+    return;
+  }
+
+  // Create a client, which is the top level object that manages all
+  // server connections.
+  final client = MCPClient(
+    Implementation(name: 'example dart client', version: '0.1.0'),
+  );
+  // The server example prints this URL after it binds.
+  final uri = Uri.parse(args.single);
+  print('connecting to server at $uri');
+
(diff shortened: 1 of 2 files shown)


pkgs/dart_mcp/example/streamable_http_client.dart
```diff
@@ -0,0 +1,94 @@
...
+        });
+        final result = await server.callTool(request);
+        if (result.isError == true) {
+          throw StateError('Tool call failed: ${result.content}');
+        } else {
+          print('Tool call succeeded: ${result.content}');
+        }
+      } else {
+        throw ArgumentError('Unexpected tool ${tool.name}');
+      }
```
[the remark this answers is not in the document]
Yusufihsangorgel: Same as tools_client.dart, on purpose. This client is paired with streamable_http_server.dart, and that server only serves greet.

pkgs/dart_mcp/example/streamable_http_client.dart
```diff
@@ -0,0 +1,94 @@
...
+      clientInfo: client.implementation,
+    ),
+  );
+  // This revision settles the version on the transport, not through
+  // initialize.
+  server.protocolVersion = protocolVersion;
+
+  try {
+    print('discovering server');
+    final discoverResult = await server.discover(
```
jakemac53, line 52: Does this not set the protocolVersion? The assignment above seems odd to me.

jakemac53: In general - I think it would be good for us to expose an optional client side helper to handle server initialization across versions - following this logic https://modelcontextprotocol.io/specification/draft/basic/versioning#backward-compatibility-with-initialization-based-versions.

This would be similar to what we do on the server side, possibly re-using the `MCPServerInitialization` type for the parameter of the new client side method. It looks like we might be able to just make the response type a `DiscoverResult` object, but this could cause problems in the future if it adds additional required fields which don't have defaults, so we could choose some other type.

Both `discover` and `initialize` should document that they will return errors on incompatible versions, and we should consider deprecating them as well.

This could be done in a separate PR, but we should file an issue to not lose track of it if so.

jakemac53 (approved):

jakemac53:
@Yusufihsangorgel lmk if you want to address the comment in this PR or separately (probably separately though?).

A simple changelog entry (add server/client streamable http client example) is also probably prudent for this as well so please add that to make the check happy.

pkgs/dart_mcp/example/streamable_http_client.dart
Yusufihsangorgel: This assignment is still needed in 65bcbe25. Here `discover()` returns its result without setting `protocolVersion`, and `initialize` answers 404 on this transport. Nothing else settles the version. I filed #688 for the helper. Once it lands this line goes away.

Yusufihsangorgel:
Added the changelog entry in 65bcbe25, and the helper goes in a separate PR under #688.

jakemac53 (approved):

PR #674, opened 2026-09-12
follow up on the review comments on #636

Yusufihsangorgel:
Answers three comments on #636 in code.

`_advertisedCapabilities` becomes a protected `advertisedCapabilities` getter and `SubscriptionsSupport` overrides it. The type check on `this` goes away. The `response.done` catch is now `on IOException`, and `id` is capitalized in the library prose.

The race question is answered on its thread, with no code change.


jakemac53 (approved):

Yusufihsangorgel:
Merged main. The close handler stays as #663 landed it. The narrowing is now on the `await response.done` catch, and nothing else moved.

PR #676, opened 2026-09-15
add a multi-version client and server

Yusufihsangorgel:
Adds paired client and server examples around a single `greet` tool. Its server uses stdio by default and serves the tool over Streamable HTTP with `--http`. Each path asks for a name before returning a greeting, and the stdio path translates that request for an older revision. The shim #655 landed does that translation. Over `--http` the result goes out as it is.

Part of #668


pkgs/dart_mcp/example/multi_version_client.dart
```diff
@@ -0,0 +1,69 @@
...
+    ),
+  );
+  print('initialized on ${initializeResult.protocolVersion?.versionString}');
+  server.notifyInitialized();
+
+  // The tool takes no arguments. It asks for the name it greets instead.
+  final result = await server.callTool(CallToolRequest(name: 'greet'));
+  for (final content in result.content) {
+    if (content.isText) print((content as TextContent).text);
+  }
```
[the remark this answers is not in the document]
Yusufihsangorgel: Fixed in 878c9f90 with the check this package documents for the union, `isInputRequired` before reading the complete result. An `is` test cannot separate the two, since both erase to the same map. The example now throws when the server asks for input it cannot answer.

pkgs/dart_mcp/example/multi_version_client.dart
```diff
@@ -0,0 +1,69 @@
...
+void main() async {
+  final client = GreetingClient(
+    Implementation(name: 'example dart client', version: '0.1.0'),
+  );
+
+  // The server serves stdio when it gets no arguments.
+  final process = await Process.start('dart', [
+    'run',
+    'example/multi_version_server.dart',
+  ]);
```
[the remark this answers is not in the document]
Yusufihsangorgel: Measured the other client examples here: six of six spawn the server the same way and none forwards its stderr. I kept this pair consistent with them. Piping a child's stderr is what the tests in this package do instead.

pkgs/dart_mcp/example/multi_version_server.dart
```diff
@@ -0,0 +1,164 @@
...
+library;
+
+import 'dart:io' as io;
+
+import 'package:dart_mcp/server.dart';
+import 'package:dart_mcp/stdio.dart';
+import 'package:dart_mcp/streamable_http.dart';
+
+void main(List<String> args) async {
+  if (args.contains('--http')) return _serveStreamableHttp();
```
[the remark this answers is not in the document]
Yusufihsangorgel: Every example here declares `void main`, fifteen of fifteen with this pair added, nine of them async. Each one relies on the event loop instead of awaiting `main`. Changing this one alone would make it the outlier. 878c9f90 leaves the signature as it is.

pkgs/dart_mcp/example/multi_version_server.dart
```diff
@@ -0,0 +1,164 @@
...
+  final greetTool = Tool(
+    name: 'greet',
+    description: 'greets a name it asks the client to collect',
+    inputSchema: Schema.object(),
+  );
+
+  /// The implementation of the `greet` tool, asking for a name on the first
+  /// call and greeting the answer the second call carries.
+  CallToolResponse _greet(CallToolRequest request) {
+    if (request.elicitResult('name') case final answer?) {
```
jakemac53, line 143: Since this is an example, I think its worth documenting what we are doing here  - checking if the request already contains an elicitResult (meaning, its a retry request with the extra input data). And maybe link to the MCP spec.

pkgs/dart_mcp/example/multi_version_client.dart
```diff
@@ -0,0 +1,75 @@
...
+  final server = client.connectServer(
+    stdioChannel(input: process.stdout, output: process.stdin),
+  );
+  // When the server connection is closed, kill the process.
+  unawaited(server.done.then((_) => process.kill()));
+
+  final initializeResult = await server.initialize(
```
jakemac53, line 39: Lets comment about what is actually happening here, depending on the protocol version?

pkgs/dart_mcp/example/multi_version_server.dart
Yusufihsangorgel: Documented in ef8ae2c2. The comment names the retry round and the key the client's answer arrives under, with a link to the revision's multi-round page.

pkgs/dart_mcp/example/multi_version_client.dart
Yusufihsangorgel: Commented at the handshake in ef8ae2c2. Below 2026-07-28 the tool's input request arrives as an `elicitation/create`, and on 2026-07-28 it comes back as the call's own result. Running the example over stdio shows the first, `--http` the second.

jakemac53: I made a similar comment on the other PR, but this example only works when _both_ the client _and_ the server support multiple versions, otherwise this initialize call would fail on a server that doesn't support it.

I think ultimately we want this client to actually demonstrate how to handle a server which doesn't support legacy, or only supports legacy modes as well.

jakemac53 (approved):
I think this is a fine starting place, but see https://github.com/dart-lang/ai/pull/671#discussion_r4063885496. I think we want that and then will want to update this some, but we can land this as is for now.

pkgs/dart_mcp/example/multi_version_client.dart
Yusufihsangorgel: Agreed. #688 tracks a client helper for servers of either era, and this example can move onto it once it lands.

Issue #37, opened 2025-04-07
Add cancellation support


jakemac53:
https://spec.modelcontextprotocol.io/specification/2024-11-05/basic/utilities/cancellation/

PR #678, opened 2026-09-15
keep a cancelled request off the wire

Yusufihsangorgel:
A `notifications/cancelled` for a request in flight here takes it off the wire. Its response, its progress and its handler's notifications all stop. Handlers run to the end, since stopping early is a SHOULD. A handler that wants to return early reads `MCPBase.captureIncomingRequestActivity`. Every valid notification reaches `MCPBase.cancellations` for a subclass to log.

A cancelled `subscriptions/listen` ends its subscription and releases the resource updates it armed. Both roles bound retained cancellations with `maxRetainedCancellations`. Passing it ends the connection instead of dropping a cancellation, and the progress tokens it also bounds forget their oldest.

Part of #37. A follow-up carries the listen half.


pkgs/dart_mcp/lib/src/shared.dart
```diff
@@ -135,19 +256,217 @@ base class MCPBase {
...
+    final params = message[Keys.params];
+    if (params is! Map<String, Object?>) return;
+    final meta = params[Keys.meta];
+    if (meta is! Map<String, Object?>) return;
+    final token = MetaWithProgressToken.fromMap(meta).progressToken;
+    if (token == null) return;
+    if (_unownedProgressTokens.length >= _maxRetainedCancellations) {
+      _unownedProgressTokens.remove(_unownedProgressTokens.first);
+    }
+    _unownedProgressTokens.add(token);
```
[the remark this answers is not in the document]
Yusufihsangorgel: Right, zero made `first` read an empty set. In 5e5ab578 both sites go through one `_disownToken`, which inserts the token and then trims while the set is over the bound, so a zero bound ends up empty again. The new test `a zero token bound retains nothing and stays up` fails without that trim.

pkgs/dart_mcp/lib/src/shared.dart
```diff
@@ -135,19 +256,217 @@ base class MCPBase {
...
+                }
+                if (_cancelledRequests.remove(id)) {
+                  if (token != null) {
+                    if (_unownedProgressTokens.length >=
+                        _maxRetainedCancellations) {
+                      _unownedProgressTokens.remove(
+                        _unownedProgressTokens.first,
+                      );
+                    }
+                    _unownedProgressTokens.add(token);
```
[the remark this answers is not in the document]
Yusufihsangorgel: This path calls the same `_disownToken` in 5e5ab578, so the trim runs after the insert here too and a zero bound holds nothing. One helper instead of two guards keeps the two sites from drifting apart later.

pkgs/dart_mcp/lib/src/server/resources_support.dart
```diff
@@ -246,31 +249,69 @@ base mixin ResourcesSupport on MCPServer {
...
 
   /// Called whenever the list of resources changes, it is the job of the client
   /// to then ask again for the list of tools.
   void _notifyResourceListChanged() => _resourceListChangedController.add(null);
 }
+
+/// A single throttled wire subscription and every protocol owner of it.
+final class _ResourceUpdateSubscription {
+  final controller = StreamController<ResourceUpdatedNotification>();
+  final modernOwners = <Object, bool Function()>{};
```
jakemac53, line 321: Lets document the key/value here.

pkgs/dart_mcp/lib/src/server/server.dart
```diff
@@ -145,11 +145,16 @@ abstract base class MCPServer extends MCPBase {
...
+  /// Serves MCP on [channel].
+  ///
+  /// [maxRetainedCancellations] caps how many client cancellations wait here
+  /// for an answer.
   MCPServer.fromStreamChannel(
     super.channel, {
     required this.implementation,
     this.instructions,
     super.protocolLogSink,
+    super.maxRetainedCancellations,
```
jakemac53: I am not convinced we should be bounding this - no requests should go unanswered, and if they are then we will anyways have other memory leaks for sure.

Or am I misunderstanding the model?

pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
```diff
@@ -187,6 +204,12 @@ base mixin SubscriptionsSupport on MCPServer {
       );
     } finally {
       _subscriptions.remove(subscriptionId);
+      if (this case final ResourcesSupport resources) {
+        await [
+          for (final uri in resourceUris)
```
jakemac53, line 209: Can we do something more efficient here, only iterating over the subscribed resources etc?

pkgs/dart_mcp/lib/src/shared.dart
```diff
@@ -37,6 +39,74 @@ base class MCPBase {
...
+  /// Every request the peer has sent that this side has not answered yet, by
+  /// JSON-RPC ID.
+  ///
+  /// An entry goes in when the request arrives and comes out when its response
+  /// leaves, so this holds exactly the requests a cancellation may still refer
+  /// to.
+  final _inFlightRequests = <Object, _IncomingRequest>{};
+
+  /// The IDs in [_inFlightRequests] the peer has cancelled.
+  final _cancelledRequests = <Object>{};
```
jakemac53: It might be safer/better to use an expando for this, on the request? Then we don't have to worry so much about cleaning it up, it will disappear with the request object.

jakemac53: nit: lets call this `_cancelledRequestIds` (its the IDs, not the requests themselves).

Or, possibly just put an expando on the incoming requests, (a boolean for if its cancelled or not), and this way it would get cleaned up automatically so less chance for leaks.

pkgs/dart_mcp/lib/src/server/resources_support.dart
Yusufihsangorgel: Documented in afbfc1ba. A key is the subscription ID that its `subscriptions/listen` request was acknowledged under, and its value answers whether that request is still uncancelled. An update goes out once any owner answers `true`.

pkgs/dart_mcp/lib/src/shared.dart
```diff
@@ -37,6 +39,74 @@ base class MCPBase {
...
+
+  /// The active request ID for each progress token.
+  final _requestsByProgressToken = <ProgressToken, Object>{};
+
+  /// Progress tokens of cancelled requests this side has already answered.
```
jakemac53: Can you "explain like I am 5" this comment, I don't really understand why we would care about unowned progress tokens

pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
```diff
@@ -43,6 +46,18 @@ base mixin SubscriptionsSupport on MCPServer {
...
+      // A cancellation for a listen request ends that subscription. Without
+      // this the handler waits until shutdown, its response never leaves, and
+      // the connection keeps the cancellation retained for a request that is
+      // never answered.
+      _cancellationSubscription = cancellations.listen((notification) {
+        final id = notification.requestId;
+        if (id == null) return;
+        final subscription = _subscriptions[id];
+        if (subscription != null && !subscription.isCompleted) {
+          subscription.complete();
```
jakemac53, line 58: Should this remove it from _subscriptions as well?

pkgs/dart_mcp/lib/src/shared.dart
```diff
@@ -135,19 +256,227 @@ base class MCPBase {
...
+  /// cancellation is reported, including one for an ID this side never saw or
+  /// has already answered, because the ID may name a request this side sent.
+  /// Only an ID that is in flight here is retained. Unknown and completed IDs
+  /// produce no error response and do not change what goes on the wire.
+  void _handleCancelled(CancelledNotification notification) {
+    // A JSON-RPC ID is a `String` or a number, so anything else cannot name a
+    // request. `RequestId` is an extension type on `Object`, so the value has
+    // to be tested instead of cast.
+    final Object? id = notification.requestId;
+    if (id == null || (id is! String && id is! num)) return;
```
jakemac53: It feels like this should at least get logged somewhere? At least the invalid IDs (but maybe not null ones).

pkgs/dart_mcp/lib/src/shared.dart
```diff
@@ -135,19 +256,227 @@ base class MCPBase {
...
+  }
+
+  /// Retains [token] as unowned, never holding more than the bound allows.
+  ///
+  /// Trimming after the insert keeps the bound exact at every value, zero
+  /// included, where the set ends up empty again and the next progress
+  /// notification for [token] reaches the peer.
+  void _disownToken(ProgressToken token) {
+    _unownedProgressTokens.add(token);
+    while (_unownedProgressTokens.length > _maxRetainedCancellations) {
```
jakemac53: Bounding this is similarly concerning to me here, what is the exact situation where we need to bound it? Partially I don't understand how it could be unowned at all, so I don't know how to judge if this is a necessary thing or not.

jakemac53 (commented):
Left a few comments to address/respond to, haven't fully finished though

pkgs/dart_mcp/lib/src/server/server.dart
Yusufihsangorgel: Removed in bcc591d3. You read it right. An entry went in only for a request already in flight and came out with its response, making that set a subset of the unanswered requests. Bounding it protected nothing the unanswered requests do not already show.

pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
Yusufihsangorgel: This iterates the accepted `resourceSubscriptions` for one subscription, not every resource the server serves, and `_stopUpdatesFor` is a map lookup per URI. A subscription with no resource filter iterates nothing at all. Unchanged in bcc591d3.

pkgs/dart_mcp/lib/src/shared.dart
Yusufihsangorgel: Gone entirely in bcc591d3. A `_IncomingRequest` already carries a cancelled flag and lives exactly as long as its request. An expando would add cleanup this already has.

pkgs/dart_mcp/lib/src/shared.dart
Yusufihsangorgel: Rewritten in bcc591d3. A token lands there by one route. A peer cancels a request, the handler answers anyway, and dropping that response takes both the request and its token out of the maps. Progress sent afterwards has nothing left to recognise it as cancelled.

pkgs/dart_mcp/lib/src/server/subscriptions_support.dart
Yusufihsangorgel: Its `finally` does. Completing here wakes the handler waiting on that completer, and the handler drops its subscription entry before the response leaves. Unchanged in bcc591d3.

pkgs/dart_mcp/lib/src/shared.dart
Yusufihsangorgel: Added in bcc591d3, on the protocol log under a `!!!` prefix alongside `<<<` and `>>>`. A cancellation with no `requestId` stays silent, as you suggested.

pkgs/dart_mcp/lib/src/shared.dart
Yusufihsangorgel: Kept, with the route documented in bcc591d3. A token is unowned once its request is gone. Dropping the response took its ID out of the in-flight map. Nothing arrives later to clean the token up. The bound covers only tokens now, under the name `maxRetainedTokens`.

pkgs/dart_mcp/lib/src/server/resources_support.dart
```diff
@@ -246,31 +249,76 @@ base mixin ResourcesSupport on MCPServer {
...
-                    ResourceUpdatedNotification.methodName,
-                    notification,
-                  );
-                }),
+      () => runOutsideRequest(() {
+        final subscription = _ResourceUpdateSubscription();
+        subscription.controller.stream
+            .throttle(resourceUpdateThrottleDelay, trailing: true)
+            .listen((notification) {
+              if (!subscription.hasLegacyOwner &&
```
jakemac53: Lets put this behind a `hasListeners` or `isActive` getter on `_ResourceUpdateSubscription`, I think its better to encapsulate this logic there

pkgs/dart_mcp/lib/src/server/resources_support.dart
```diff
@@ -246,31 +249,76 @@ base mixin ResourcesSupport on MCPServer {
...
+              sendNotification(
+                ResourceUpdatedNotification.methodName,
+                notification,
+              );
+            });
+        return subscription;
+      }),
     );
+    if (subscriptionId == null) {
+      subscription.hasLegacyOwner = true;
```
jakemac53, line 274: Probably worth a comment, that the legacy protocol has no subscription ids

pkgs/dart_mcp/lib/src/server/resources_support.dart
```diff
@@ -246,31 +249,76 @@ base mixin ResourcesSupport on MCPServer {
...
+    if (subscriptionId == null) {
+      subscription.hasLegacyOwner = false;
+    } else {
+      subscription.modernOwners.remove(subscriptionId);
+    }
+    if (subscription.hasLegacyOwner || subscription.modernOwners.isNotEmpty) {
+      return;
+    }
+    _subscribedResources.remove(uri);
+    await subscription.controller.close();
```
jakemac53, line 286: I think most of this should also move to a method on _ResourceSubscription, so that it looks something like this here:

```suggestion
    if (subscription == null) return;
    // This function would return `true` if the subscription is now dead (has no listeners)
    if (subscription.stopUpdates(subscriptionId)) {
      _subscribedResources.remove(uri);
    }
```

pkgs/dart_mcp/lib/src/shared.dart
```diff
@@ -135,19 +261,225 @@ base class MCPBase {
...
+  ///
+  /// A token this connection never tracked is not one it can call cancelled.
+  /// The request-scoped dispatcher builds a server per message, so a progress
+  /// token can reach the sink on a connection that never saw the request it
+  /// belongs to, and dropping it there would take a notification the peer is
+  /// still waiting for.
+  bool _progressWasCancelled(ProgressToken token) {
+    if (_unownedProgressTokens.contains(token)) return true;
+    final id = _requestsByProgressToken[token];
+    if (id == null) return false;
```
jakemac53: This case seems to be the whole reason for tracking `_unownedProgressTokens` - is there any reason we can't treat any unknown token as cancelled (just return true here), and drop `_unownedProgressTokens` entirely?

I still don't understand the actual flow that would cause this, and how returning false in this branch improves things.

pkgs/dart_mcp/lib/src/server/resources_support.dart
Yusufihsangorgel: f176ef6b adds `hasListeners` on `_ResourceUpdateSubscription`, and the throttled listener now checks only that.

pkgs/dart_mcp/lib/src/server/resources_support.dart
Yusufihsangorgel: In f176ef6b, `stopUpdates` returns whether no owner is still registered. It counts registered owners instead of checking `!hasListeners`, since a cancelled listen request that is not removed yet still holds the subscription open. 916d79d4 adds a test for that case, and `controller.close()` stays at the call site.

pkgs/dart_mcp/lib/src/server/resources_support.dart
Yusufihsangorgel: Added in f176ef6b, both at that line and on `stopUpdates`.

pkgs/dart_mcp/lib/src/shared.dart
Yusufihsangorgel: No good reason. 916d79d4 treats an unknown token as cancelled and removes `_unownedProgressTokens` along with its constructor parameter. My earlier explanation relied on the spec letting a peer ignore such progress, while the [progress page](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/progress) says progress notifications "MUST only reference tokens that" were provided in an active request.

jakemac53 (approved):

PR #679, opened 2026-09-15
add a client helper for subscription streams

Yusufihsangorgel:
Adds `ServerConnection.listen`, the client half of the subscription streams the server has served since #636. A handle comes back right away, reports the filter the server agreed to, and routes each notification to the subscription that opened it. Closing one subscription ends only that stream. Over stdio it sends `notifications/cancelled` for that request, over Streamable HTTP it drops the SSE response. A cancellation arriving for one subscription ends that one and leaves the connection alone.

Three helper names here are this package's own and not the schema's. Reading the outgoing request ID off the channel sink leaves `package:json_rpc_2` untouched.

Tracked in #668.


pkgs/dart_mcp/lib/src/client/subscriptions.dart
```diff
@@ -0,0 +1,140 @@
...
+    final fields = params as Map<String, Object?>?;
+    final meta = fields?[Keys.meta];
+    if (meta is! Map<String, Object?>) return;
+    final sentId = meta[Keys.subscriptionIdMeta];
+    if (sentId == null || RequestId(sentId) != id) return;
+    if (!_acknowledged.isCompleted) return;
+    if (!_notifications.isClosed) {
+      _notifications.add((method: method, params: Notification(fields!)));
+    }
+  }
```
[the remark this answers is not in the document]
Yusufihsangorgel: Applied in fa0841a4. Promotion also removes the null assertion further down, since the promoted map is the one the notification carries.

Yusufihsangorgel:
#678 puts a `notifications/cancelled` handler on `MCPBase`, and `listen` here registers one of its own. `package:json_rpc_2` throws on a second registration for one method. Whichever of these lands first, the other needs its handler moved to `MCPBase.cancellations` before it goes in.

pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -837,6 +858,101 @@ base class ServerConnection extends MCPBase {
...
+      sendNotification(
+        CancelledNotification.methodName,
+        CancelledNotification(requestId: id),
+      );
+      completeRequestLocally(this, id);
+    }
+  }
+
+  void _forwardSubscriptionNotification(String method, Object? params) {
+    for (final subscription in _subscriptions.values.toList()) {
```
jakemac53: Instead of having all the logic to check for a matching ID in `forward`, can't we just directly pull off the ID from the params here, and grab the matching subscription from `_subscriptions`? That avoids iterating all of them, having them each individually pull the IDs out of params, etc.

pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -350,6 +355,10 @@ base class ServerConnection extends MCPBase {
       PromptListChangedNotification.methodName,
       (notification) {
         _responseCache.invalidateMethod(ListPromptsRequest.methodName);
+        _forwardSubscriptionNotification(
```
jakemac53, line 361: Is this just for the `stdio` transport case? If so, lets document that.

pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -837,6 +858,101 @@ base class ServerConnection extends MCPBase {
...
+  /// The subscriptions this connection has open, keyed by [Subscription.id].
+  final _subscriptions = <RequestId, Subscription>{};
+
+  final RequestCancellation? _requestCancellation;
+
+  /// Whether [listen] has registered the server's subscription handlers.
+  ///
+  /// [_subscriptions] empties out again as subscriptions end, so it cannot
+  /// answer this.
+  bool _subscriptionHandlersRegistered = false;
```
jakemac53: It seems to me from the spec that we would get one of these for each subscription call? So I think this needs to be a map by subscription ID?

pkgs/dart_mcp/lib/src/shared.dart
```diff
@@ -129,6 +166,42 @@ base class MCPBase {
...
+      unawaited(shutdown());
+      throw StateError('Encoding "$methodName" recorded no JSON-RPC ID.');
+    }
+    final sent = Completer<void>();
+    sent.future.ignore();
+    scheduleMicrotask(() => sent.complete(Future.sync(forward)));
+    return (
+      id: id,
+      result: result.then(
+        (value) => (value as Map?)?.cast<String, Object?>() as T,
```
jakemac53, line 208: Do we need the cast here?

pkgs/dart_mcp/lib/src/shared.dart
```diff
@@ -25,6 +47,7 @@ import 'api/api.dart';
 /// - [ServerConnection] A class that represents an active server connection.
 base class MCPBase {
   late final Peer _peer;
+  final _local = StreamController<Map<String, Object?>>(sync: true);
```
jakemac53, line 59: Please document this

jakemac53 (commented):
Still trying to wrap my head around this one, left a few comments though

pkgs/dart_mcp/lib/src/client/client.dart
Yusufihsangorgel: d4f8b89e now reads the subscription ID off the raw params once and looks it up in `_subscriptions` directly.

pkgs/dart_mcp/lib/src/client/client.dart
Yusufihsangorgel: Not only stdio. On Streamable HTTP each listen response merges into this connection's incoming stream, and these handlers see those notifications there too. Only the `notifications/cancelled` fallback in `_cancelSubscription` is stdio-specific. 6025a43a documents both.

pkgs/dart_mcp/lib/src/client/client.dart
Yusufihsangorgel: That flag only guarded registering the handlers once per connection, since json_rpc_2 takes one handler per method. Acknowledgements already route by ID through `_subscriptions`. 6025a43a registers both handlers in the constructor instead and drops the flag.

pkgs/dart_mcp/lib/src/shared.dart
Yusufihsangorgel: Yes. json_rpc_2 hands back whatever map the channel decoded. Without the `.cast`, a `Map<dynamic, dynamic>` fails the `as T`, and `sendRequestKeepingProgress` a few lines down casts the same way. 6025a43a keeps it.

pkgs/dart_mcp/lib/src/shared.dart
Yusufihsangorgel: 6025a43a adds a doc comment covering who completes requests through it, why it is synchronous, and when it closes.

pkgs/dart_mcp/lib/src/client/client.dart
jakemac53: Ahhh ok that explains a lot, I did not realize we were actually merging the streams together like that. The comment clears this up though.

pkgs/dart_mcp/lib/src/shared.dart
jakemac53: Ah ok, so coming off the wire these are just a `Map<dynamic, dynamic>`, that makes sense

Issue #28, opened 2025-04-02
Support pagination for tools/resources


jakemac53:
When listing tools and resources, we currently don't support pagination.

Since we have a bespoke API for handling adding tools and resources, we need to implement this ourselves and can't just leave it up to clients (they don't have the ability today to intercept these requests).

jakemac53:
For resources in particular we might actually want to tweak the API here to handle cases with large numbers of resources better. Possibly we should have less magic and leave most of the logic up to the client.

PR #682, opened 2026-09-16
walk the pages of each list request

Yusufihsangorgel:
A client that wants every tool from a server that pages its list has to thread the cursor by hand. The fixture under `tool/` drops `nextCursor` today. Methods named `listAllTools`, `listAllResources`, `listAllResourceTemplates` and `listAllPrompts` now walk the pages and yield the items as a `Stream`. Each single-page method stays byte-identical. A default 64-page bound throws instead of walking forever, while passing `null` lifts it. Tests cover the cursor threading, the bound, the argument check and an empty page.

Part of #28.


pkgs/dart_mcp/test/client/pagination_test.dart
```diff
@@ -0,0 +1,412 @@
...
+      meta: MetaWithProgressToken(progressToken: ProgressToken('token')),
+    );
+    final progress = <num>[];
+    var streamClosed = false;
+    connection
+        .onProgress(request)
+        .listen(
+          (notification) => progress.add(notification.progress),
+          onDone: () => streamClosed = true,
+        );
```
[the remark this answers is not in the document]
Yusufihsangorgel: Applied in b48e5996. `roots_test.dart` captures and tears down a listened stream the same way in three places, so the change follows what the repository already does.

pkgs/dart_mcp/lib/src/client/client.dart
```diff
@@ -562,6 +562,126 @@ base class ServerConnection extends MCPBase {
...
+  /// no `nextCursor`.
+  ///
+  /// [request] gives the first cursor and the metadata every page carries.
+  /// [maxPageCount] bounds the walk, throwing past it; `null` lifts the bound.
+  Stream<Tool> listAllTools({
+    ListToolsRequest? request,
+    int? maxPageCount = 64,
+  }) => _listAllPages(
+    ListToolsRequest.methodName,
+    (cursor) => ListToolsRequest(cursor: cursor, meta: request?.meta),
```
jakemac53: I am a bit concerned about accidentally dropping any additional fields on these.

We could instead just copy the underlying map for the request, and only add the cursor to that.

Possibly, this could be a `copyWithCursor` extension on PaginatedRequest, so it can be shared?

Yusufihsangorgel: Added `copyWithCursor` to `PaginatedRequest` in ec2aee92. It copies the map and replaces only the cursor, and every `listAll*` hands its own request to the walk instead of building a fresh one for each page. A new test puts an entry the package never reads on a request and checks that it reaches every page.

pkgs/dart_mcp/lib/src/api/api.dart
```diff
@@ -435,6 +435,15 @@ extension type PaginatedRequest._fromMap(Map<String, Object?> _value)
   Cursor? get cursor => _value[Keys.cursor] as Cursor?;
 }
 
+extension PaginatedRequestCopy on PaginatedRequest {
+  /// A copy of this request that asks for the page at [cursor].
+  ///
+  /// Copying keeps every entry the caller put on the request, including ones
+  /// this package does not read.
+  PaginatedRequest copyWithCursor(Cursor cursor) =>
```
jakemac53, line 441: I believe this can just go directly on the `PaginatedRequest` type.

Yusufihsangorgel: Moved onto the type in 73880be0. Nothing outside the library needed the separate name.

jakemac53 (approved):

